PR #19, opened 2015-10-23
Drop package if there is a new native app version

geof90:
Making this consistent with the Cordova side. 
https://github.com/Microsoft/cordova-plugin-code-push/pull/25

@lostintangent @dlebu


lostintangent:
This fix looks good to me in terms of consistency with Cordova, but I'm really keen on getting feedback from users on whether or not they want/expect to use CodePush to send client notifications when the running app is using an older version of the binary then is available in its respective store.

PR #21, opened 2015-10-28
Use documents directory

geof90:
These prevents file permissions errors with writing to and reading from the NSHomeDirectory.


dtivel:
LGTM

ababol:
Nice!

PR #22, opened 2015-10-29
CodePush.sync() API method

lostintangent:
This provides a simplified and opinionated one-line experience to CodePush-ifying an app. The CodePush.sync() method allows automating the act of checking for an update, downloading it and then subsequently applying it, all while also supporting ignoring previously failed updates, respecting mandatory updates and displaying a standard user confirmation dialog when updates are available.

The method takes an optional options object that can be used to customize the strings used when displaying the user dialog, as well as the rollback timeout.

It returns a Promise that will be resolved with a SyncStatus code, that indicates why the sync succeeded:

0 - No update was available, so the running app is up to do
1 - The app had an optional update, but the end-user chose to ignore it
2 - The app had an optional or mandatory update and the user accepted it

If a failure occurs at any point during the sync, the returned Promise will be rejected with the error reason.

I'll update the docs in a subsequent PR, but I just wanted to get this change out since the respective Cordova change is in review.

### DIFF
--- a/CodePush.ios.js
+++ b/CodePush.ios.js
@@ -96,6 +96,15 @@ function notifyApplicationReady() {
   return NativeCodePush.notifyApplicationReady();
 }
 
+/**
+ * The sync method provides a simple, one-line experience for
+ * incorporating the check, download and application of an update.
+ * 
+ * It simply composes the existing API methods together and adds additional
+ * support for respecting mandatory updates, ignoring previously failed
+ * releases, and displaying a standard confirmation UI to the end-user
+ * when an update is available.
+ */
 function sync(options = {}) {  
   var syncOptions = {
     ignoreFailedUpdates: true,
@@ -139,8 +148,10 @@ function sync(options = {}) {
             dialogButtons[0].text = syncOptions.mandatoryContinueButtonLabel;
           } else {
             message = syncOptions.optionalUpdateMessage;
-
-            dialogButtons[0].text = syncOptions.optionalInstallButtonLabel;            
+            dialogButtons[0].text = syncOptions.optionalInstallButtonLabel;     
+            
+            // Since this is an optional update, add another button
+            // to allow the end-user to ignore it       
             dialogButtons.push({
               text: syncOptions.optionalIgnoreButtonLabel,
               onPress: () => resolve(CodePush.SyncStatus.UPDATE_IGNORED)
@@ -161,9 +172,9 @@ var CodePush = {
   setUpTestDependencies: setUpTestDependencies,
   sync: sync,
   SyncStatus: {
-    NO_UPDATE_AVAILABLE: 0,
-    UPDATE_IGNORED: 1,
-    APPLY_SUCCESS: 2    
+    NO_UPDATE_AVAILABLE: 0, // The running app is up-to-date
+    UPDATE_IGNORED: 1, // The app had an optional update and the end-user chose to ignore it
+    APPLY_SUCCESS: 2 // The app had an optional/mandatory update that was successfully downloaded and is about to be applied
   }
 };
 
--- a/CodePush.ios.js
+++ b/CodePush.ios.js
@@ -107,6 +107,9 @@ function notifyApplicationReady() {
  */
 function sync(options = {}) {  
   var syncOptions = {
+    descriptionPrefix: " Description: ",
+    appendReleaseDescription: false,
+    
     ignoreFailedUpdates: true,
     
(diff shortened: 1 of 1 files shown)


geof90:
LGTM!

CodePush.ios.js
```diff
@@ -95,12 +96,86 @@ function notifyApplicationReady() {
...
+    updateTitle: "Update available",
+    rollbackTimeout: 0,
+    
+    ...options 
+  };
+
+  return new Promise((resolve, reject) => {
+    checkForUpdate()
+      .then((remotePackage) => {
+        if (!remotePackage || (remotePackage.failedAppy && syncOptions.ignoreFailedUpdates)) {
```
dtivel: failedApply?

CodePush.ios.js
```diff
@@ -95,12 +96,86 @@ function notifyApplicationReady() {
...
+ * support for respecting mandatory updates, ignoring previously failed
+ * releases, and displaying a standard confirmation UI to the end-user
+ * when an update is available.
+ */
+function sync(options = {}) {  
+  var syncOptions = {
+    ignoreFailedUpdates: true,
+    
+    mandatoryContinueButtonLabel: "Continue",
+    mandatoryUpdateMessage: "An update is available that must be installed",
```
dtivel: Add . at end of sentence.

CodePush.ios.js
```diff
@@ -95,12 +96,86 @@ function notifyApplicationReady() {
...
   checkForUpdate: checkForUpdate,
   getCurrentPackage: getCurrentPackage,
   notifyApplicationReady: notifyApplicationReady,
-  setUpTestDependencies: setUpTestDependencies
+  setUpTestDependencies: setUpTestDependencies,
+  sync: sync,
+  SyncStatus: {
+    NO_UPDATE_AVAILABLE: 0, // The running app is up-to-date
+    UPDATE_IGNORED: 1, // The app had an optional update and the end-user chose to ignore it
+    APPLY_SUCCESS: 2 // The app had an optional/mandatory update that was successfully downloaded and is about to be applied
```
dtivel: The field name and comment describe different states.  APPLY_SUCCESS means apply succeeded; however, the comment says the "... update ... is about to be applied."  Given the implementation above the field name should be changed to DOWNLOAD_SUCCEEDED.

CodePush.ios.js
```diff
@@ -95,12 +96,86 @@ function notifyApplicationReady() {
...
+        }
+        else {
+          var message = null;
+          var dialogButtons = [
+            {
+              text: null,
+              onPress: () => { 
+                remotePackage.download()
+                  .then((localPackage) => {
+                    resolve(CodePush.SyncStatus.APPLY_SUCCESS);
```
dtivel: apply(...) starts on the next line so this status should be DOWNLOAD_SUCCEEDED.

dtivel:
LGTM

CodePush.ios.js
lostintangent: I agree this name isn't great. APPLY_SUCCESS is also being used for the Cordova plugin, so I'm cool with a name change as long as we're consistent.

CodePush.ios.js
lostintangent: Yep agreed.

CodePush.ios.js
```diff
@@ -95,12 +96,94 @@ function notifyApplicationReady() {
...
+        else {
+          var message = null;
+          var dialogButtons = [
+            {
+              text: null,
+              onPress: () => { 
+                remotePackage.download()
+                  .then((localPackage) => {
+                    resolve(CodePush.SyncStatus.UPDATE_DOWNLOADED);
+                    localPackage.apply(syncOptions.rollbackTimeout);
```
silhouettes: Hey I don't think a failure in the apply promise is being handled here - would it be better to just add below this .then():

.catch(reject)
.done();

CodePush.ios.js
```diff
@@ -95,12 +96,94 @@ function notifyApplicationReady() {
...
+          
+          // If the update has a description, and the developer
+          // explicitly chose to display it, then set that as the message
+          if (syncOptions.useReleaseDescription && remotePackage.description) {
+            message = remotePackage.description;  
+          }
+          
+          AlertIOS.alert(syncOptions.updateTitle, message, dialogButtons);
+        }
+      }, reject);
```
silhouettes: I think it's the same issue here, it might be better to do:
```
.catch(reject)
.done();
```

silhouettes:
LGTM!

PR #25, opened 2015-10-30
Bumping the version and updating the SDK dependency

lostintangent:
Bumping the version in order to release an update to NPM, and also taking a dependency on the new SDK release


dtivel:
LG

PR #26, opened 2015-11-01
Removing @flow annotation from JS module

lostintangent:
At some point, we should flow-ify the plugin, but for now, this file shouldn't be marked for type checking since it doesn't actually succeed. Additionally, we shouldn't be using the @providesModule annotation, since that's an internal feature of the RN packager that we shouldn't be trying to use.


dtivel:
LG

silhouettes:
LGTM

PR #34, opened 2015-11-04
Is first run

geof90:
Add support for isFirstRun.

@lostintangent

### DIFF
--- a/CodePush.m
+++ b/CodePush.m
@@ -9,6 +9,7 @@ RCT_EXPORT_MODULE()
 
 NSTimer *_timer;
 BOOL usingTestFolder = NO;
+BOOL didUpdate = NO;
 
 NSString * const FailedUpdatesKey = @"FAILED_UPDATES";
 NSString * const UpdateBundleFileName = @"app.jsbundle";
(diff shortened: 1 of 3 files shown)


lostintangent:
LGTM!

lostintangent:
LGTM!

CodePush.m
```diff
@@ -177,6 +179,18 @@ - (void)startRollbackTimer:(int)rollbackTimeout
...
 }
 
+RCT_EXPORT_METHOD(isFirstRun:(NSString *)packageHash
+                  resolve:(RCTPromiseResolveBlock)resolve
+                  rejecter:(RCTPromiseRejectBlock)reject)
+{
+    NSError *error;
+    BOOL isFirstRun = (nil != packageHash
+                  && [packageHash length] > 0
+                  && [packageHash isEqualToString:[CodePushPackage getCurrentPackageHash:&error]]
```
dtivel: It looks like getCurrentPackageHash and getPreviousPackageHash would both null dereference if there is an error.

CodePush.m
```diff
@@ -177,6 +179,18 @@ - (void)startRollbackTimer:(int)rollbackTimeout
...
+                  && didUpdate);
```
dtivel: Make "didUpdate" the first expression to be evaluated.  It's a simple Boolean check that could make getting the current package hash unnecessary.

Remove the surrounding parentheses.

CodePush.m
geof90: Based on the developer guide, I think this is how methods that can return errors are handled?

https://developer.apple.com/library/ios/documentation/Cocoa/Conceptual/ErrorHandlingCocoa/CreateCustomizeNSError/CreateCustomizeNSError.html#//apple_ref/doc/uid/TP40001806-CH204-BAJIIGCC

dtivel:
looks good

PR #43, opened 2015-11-10
Adding new restartImmediately flag to the apply method

lostintangent:
This adds a new parameter to the `LocalPackage.apply` method which allows the app dev to specify whether they want the app to be restarted immediately or not. The default parameter is to restart the app, which allows backwards compatibility.

This flag is useful for implementing a "silent install" experience, where you want to check for the update, download and apply it in the background, but don't forcibly restart the app. In this case, when the user naturally restarts their app (either because the OS killed it, or they closed it themselves), the app would "pick up" the downloaded update on the first run.

The bulk of this change is simply making sure that the existing `isFirstRun` property and rollback timeout support continue to work. We discussed allowing doing the restart or resume as well, but I couldn't think of a particularly compelling reason to restart on resume, from an end-user experience.


CodePush.m
```diff
@@ -57,6 +58,51 @@ - (void)cancelRollbackTimer
...
+        
+        if (pendingUpdate)
+        {
+            NSError *error;
+            NSString *pendingHash = pendingUpdate[@"hash"];
+            NSString *currentHash = [CodePushPackage getCurrentPackageHash:&error];
+            
+            // If the current hash is equivalent to the pending hash, then the app
+            // restart "picked up" the new update, but we need to kick off the
+            // rollback timer and ensure that the neccessaey state is setup.
```
dtivel: necessary

CodePush.m
```diff
@@ -112,12 +171,13 @@ - (void)startRollbackTimer:(int)rollbackTimeout
 }
 
 // JavaScript-exported module methods
+// JavaScript-exported module methods
```
dtivel: Remove

dtivel:
Looks good.

geof90:
LGTM!

README.md
```diff
@@ -248,10 +248,9 @@ Contains details about an update package that has been downloaded locally or alr
...
-If `notifyApplicationReady` is called before the time period specified by rollbackTimeout, the apply operation is considered a success.
-Otherwise, the apply operation will be marked as failed, and the application is reverted to its previous version.
+- __apply(rollbackTimeout: Number = 0, restartImmediately: Boolean = true): Promise&lt;void&gt;__: Applies this package to the application, by unzipping its contents (e.g. the JS bundle) and saving it to the location on disk that the runtime expects to find the latest version of the app. If you pass a truthy value to the `restartImmediately` parameter, then after performing the apply, the app will be immediately restarted, so that the end-user sees the changes. Otherwise, the apply will complete, but they won't take affect until the next time that the app is started (i.e. the user starts your app after they or the OS killed it).<br /><br /> If a value greater then zero is provided to the `rollbackTimeout` parameter, the application will wait for the `notifyApplicationReady` method to be called for the given number of milliseconds.
```
silhouettes: I want to just fix a few nittish things and rephrase slightly, is this okay?

- __apply(rollbackTimeout: Number = 0, restartImmediately: Boolean = true): Promise&lt;void&gt;__: Applies this package to the application by unzipping its contents (e.g. the JS bundle) and saving it to the location on disk where the runtime expects to find the latest version of the app. If the `restartImmediately` parameter is set to false, the apply will complete, but it won't take effect until the next time that the app is restarted. Otherwise, the app will be immediately restarted after performing the apply, so that the end-user sees the changes.<br /><br /> If a value greater than zero is provided to the `rollbackTimeout` parameter, the application will wait for the `notifyApplicationReady` method to be called for the given number of milliseconds.

Also, again a little nittish, but is it a bit strange that we are talking about 'restartImmediately' before we talk about 'rollbackTimeout', considering the arguments are the other way around?

silhouettes:
Looks good!

PR #49, opened 2015-11-13
Download progress

geof90:
```remotePackage.download(progressHandler?: Function)``` now accepts a progress handler, which gets periodically called with a progress status object until the download completes 
```
{ 
  receivedBytes: receivedBytes,
  totalBytes: totalBytes
}
```
In addition, downloads are now streamed to disk instead of being loaded in the device's memory as a string, hence freeing up app memory.

Added the use of this API in the example project.

Modified docs to explain this behaviour.

### CODE REVIEW

**Comment by lostintangent** [comment]
This is awesome! LGTM. After this gets merged, I'd like to release an update of the plugin.

**Comment by lostintangent** [comment]
I was just about to ask if you could update the docs as well, but it looks like you took care of that already :)

File: CodePushDownloadHandler.m [hunk 1]
```diff
@@ -0,0 +1,79 @@
...
+        self.failCallback([self.outputFileStream streamError]);
+        [self.outputFileStream close];
+        [connection cancel];
+    }
+}
+
+- (void)connection:(NSURLConnection*)connection didFailWithError:(NSError*)error
+{
+    self.failCallback(error);
+    [self.outputFileStream close];
```
dtivel: If the failCallback(...) (or progress or done) callee throws, the file stream won't be closed?

Move the closure before the callback.

CodePushDownloadHandler.m
```diff
@@ -0,0 +1,79 @@
...
+    
+    do {
+        NSUInteger bytesWritten = [self.outputFileStream write:[data bytes]
+                                                     maxLength:bytesLeft];
+        if (bytesWritten == -1) {
+            break;
+        }
+        
+        bytesLeft -= bytesWritten;
+    } while (bytesLeft>0);
```
dtivel: Fix spacing.

CodePushPackage.m
```diff
@@ -149,50 +149,51 @@ + (NSString *)getPackageFolderPath:(NSString *)packageHash
...
+                                               atomically:YES
+                                                 encoding:NSUTF8StringEncoding
+                                                    error:&error];
+                           if (error) {
+                               failCallback(error);
+                           } else {
+                               doneCallback();
+                           }
+                       }
+                       failCallback:failCallback];
```
dtivel: Fix indentation.

CodePushPackage.m
```diff
@@ -149,50 +149,51 @@ + (NSString *)getPackageFolderPath:(NSString *)packageHash
...
-    NSString *packageJsonString = [[NSString alloc] initWithData:updateSerializedData encoding:NSUTF8StringEncoding];
-    [packageJsonString writeToFile:[packageFolderPath stringByAppendingPathComponent:@"app.json"]
-                        atomically:YES
-                          encoding:NSUTF8StringEncoding
-                             error:error];
+    if (error) {
+        return failCallback(error);
+    }
+    
+    NSString *downloadFilePath = [packageFolderPath stringByAppendingPathComponent:@"app.jsbundle"];
```
dtivel: UpdateBundleFileName

package-mixins.js
```diff
@@ -1,20 +1,36 @@
 var extend = require("extend");
+var { NativeAppEventEmitter } = require('react-native');
```
dtivel: In the CodePush repo we use double quotes here and 4 spaces per indent.

Let's do the same here.

README.md
```diff
@@ -265,7 +265,7 @@ The `RemotePackage` inherits all of the same properties as the `LocalPackage`, b
 - __downloadUrl__: The URL at which the package is available for download. (String). This property is only needed for advanced usage, since the `download` method will automatically handle the acquisition of updates for you.
 
 ##### Methods
-- __download(): Promise<LocalPackage>__: Downloads the package update from the CodePush service. Returns a Promise that resolves with the `LocalPackage`.
+- __download(progressHandler?: Function): Promise<LocalPackage>__: Downloads the package update from the CodePush service. If a `progressHandler` is specified, it will be called periodically with an object (`{ totalBytes: Number, receivedBytes: Number }`) that reports the progress of the download till the download completes. Returns a Promise that resolves with the `LocalPackage`.
```
dtivel: until not till

CodePushDownloadHandler.m
```diff
@@ -0,0 +1,79 @@
...
+    [self.outputFileStream open];
+}
+
+-(void)connection:(NSURLConnection *)connection didReceiveData:(NSData *)data {
+    self.receivedContentLength = self.receivedContentLength + [data length];
+    
+    NSUInteger bytesLeft = [data length];
+    
+    do {
+        NSUInteger bytesWritten = [self.outputFileStream write:[data bytes]
```
dtivel: NSOutputStream write returns an NSInteger.

https://developer.apple.com/library/ios/documentation/Cocoa/Reference/Foundation/Classes/NSOutputStream_Class/index.html#//apple_ref/occ/instm/NSOutputStream/write:maxLength:

Update lines 47 and 50 to us NSInteger.

CodePushPackage.m
```diff
@@ -149,50 +149,51 @@ + (NSString *)getPackageFolderPath:(NSString *)packageHash
...
 
 + (void)downloadPackage:(NSDictionary *)updatePackage
-                            error:(NSError **)error
+       progressCallback:(void (^)(long, long))progressCallback
+           doneCallback:(void (^)())doneCallback
+           failCallback:(void (^)(NSError *err))failCallback
 {
     NSString *packageFolderPath = [self getPackageFolderPath:updatePackage[@"packageHash"]];
     
+    NSError *error;
```
dtivel: Is this automatically initialized?

CodePushPackage.m
```diff
@@ -149,50 +149,51 @@ + (NSString *)getPackageFolderPath:(NSString *)packageHash
...
-    if (*error) {
-        return;
-    }
-    
-    NSString *packageJsonString = [[NSString alloc] initWithData:updateSerializedData encoding:NSUTF8StringEncoding];
-    [packageJsonString writeToFile:[packageFolderPath stringByAppendingPathComponent:@"app.json"]
-                        atomically:YES
-                          encoding:NSUTF8StringEncoding
-                             error:error];
+    if (error) {
```
dtivel: error or *error?

CodePushPackage.m
```diff
@@ -149,50 +149,51 @@ + (NSString *)getPackageFolderPath:(NSString *)packageHash
...
+                                                           options:0
+                                                           error:&error];
+                           NSString *packageJsonString = [[NSString alloc]
+                                                          initWithData:updateSerializedData encoding:NSUTF8StringEncoding];
+                           
+                           [packageJsonString writeToFile:[packageFolderPath stringByAppendingPathComponent:@"app.json"]
+                                               atomically:YES
+                                                 encoding:NSUTF8StringEncoding
+                                                    error:&error];
+                           if (error) {
```
dtivel: error or *error?

CodePushPackage.m
geof90: No, the variable's value will be null.

geof90: So according to https://developer.apple.com/library/ios/releasenotes/ObjectiveC/RN-TransitioningToARC/Introduction/Introduction.html#//apple_ref/doc/uid/TP40011226-CH1-SW5, 

Since ARC is turned on for us, the value will be initialized to nil.

However, I think I will still set it to nil explicitly, so that future contributors won't have to go through this investigation. Seems like many people think it is good practice to do so anyway too. Thanks :)

package-mixins.js
```diff
@@ -1,20 +1,36 @@
...
+    download: function download(progressHandler) {
       if (!this.downloadUrl) {
         return Promise.reject(new Error("Cannot download an update without a download url"));
       }
 
+      var downloadProgressSubscription;
+      if (progressHandler) {
+        // Use event subscription to obtain download progress.   
+        downloadProgressSubscription = NativeAppEventEmitter.addListener(
+          'CodePushDownloadProgress',
```
dtivel: "

CodePushPackage.m
```diff
@@ -149,50 +149,51 @@ + (NSString *)getPackageFolderPath:(NSString *)packageHash
...
-                                                                     error:error];
-    
-    if (*error) {
-        return;
-    }
+    if (error) {
+        return failCallback(error);
+    }
+    
+    NSString *updateBundleFileName = [packageFolderPath stringByAppendingPathComponent:@"app.jsbundle"];
```
dtivel: This constant (@"app.jsbundle") is duplicated in CodePush.m.  Can you deduplicate?

dtivel:
Looks good.

CodePushDownloadHandler.m
```diff
@@ -0,0 +1,79 @@
...
+    [self.outputFileStream open];
+}
+
+-(void)connection:(NSURLConnection *)connection didReceiveData:(NSData *)data {
+    self.receivedContentLength = self.receivedContentLength + [data length];
+    
+    NSInteger bytesLeft = [data length];
+    
+    do {
+        NSInteger bytesWritten = [self.outputFileStream write:[data bytes]
```
dlebu: We should add tests for this code. There are a lot of things that can go wrong when manually writing/counting bytes. If there is no time now, we should probably add some assertions, and play with the product as much as possible before releasing.

CodePushDownloadHandler.m
```diff
@@ -0,0 +1,79 @@
...
+                                                     maxLength:bytesLeft];
+        if (bytesWritten == -1) {
+            break;
+        }
+        
+        bytesLeft -= bytesWritten;
+    } while (bytesLeft > 0);
```
dlebu: This would be a good place for an assertion. Can `bytesLeft` ever be negative?

README.md
```diff
@@ -265,7 +265,7 @@ The `RemotePackage` inherits all of the same properties as the `LocalPackage`, b
 - __downloadUrl__: The URL at which the package is available for download. (String). This property is only needed for advanced usage, since the `download` method will automatically handle the acquisition of updates for you.
 
 ##### Methods
-- __download(): Promise<LocalPackage>__: Downloads the package update from the CodePush service. Returns a Promise that resolves with the `LocalPackage`.
+- __download(progressHandler?: Function): Promise<LocalPackage>__: Downloads the package update from the CodePush service. If a `progressHandler` is specified, it will be called periodically with an object (`{ totalBytes: Number, receivedBytes: Number }`) that reports the progress of the download until the download completes. Returns a Promise that resolves with the `LocalPackage`.
```
dlebu: We should have a name for the progress interface. `DownloadProgress` maybe?

dlebu:
Some comments, looks good otherwise.

CodePushDownloadHandler.m
geof90: I have added two assertions, one for "bytesLeft should never be negative", one for "received and expected bytes should be equal in connectionDidFinishLoading"

CodePushDownloadHandler.m
geof90: I have added some tests cases in the sample app, look for Examples/CodePushDemoApp/CodePushDemoAppTests/DownloadProgressTests.

They use the react packager (for ease of running, since it starts up automatically when u run the test project) to act as a server to serve 3 pacakges, one 100kb, one 2mb, one 50mb.

PR #51, opened 2015-11-15
Restart on resume

lostintangent:
This extends/replaces the previous `restartImmediately` flag by allowing the caller to pass a `restartMode` to `LocalPackage.apply`, which adds the ability to restart the app on the next resume, as opposed to simply restarting immediately or not. The default value when a mode isn't specified is still "immediate", but this change gives devs more options, particularly if they want to employ a "silent update" mechanism, which doesn't immediately restart the app, but does on the next app resume.

Moving forward, we could also expose an API to script that allows restarting the app at a custom, app-specific time, that way you apply the update, specifying a restart mode of none, and then force the restart sooner at a point the app can determine is acceptable (e.g. the end-user navigated back to the home page).

This change only impacts the `LocalPackage.apply` method. We need to determine how to best expose the `restartMode` via `sync`, in a way that still respects the `isMandatory` option of an update and doesn't allow the app to have a weird update experience (e.g. not applying the update immediately is kind of strange when you asked the end-user for permission to install it). We'll address that in a future PR, along with exposing the new `progressCallback` feature that was added to the `RemotePackage.download` method.

### DIFF
--- a/CodePush.m
+++ b/CodePush.m
@@ -80,7 +80,7 @@ NSString * const PendingUpdateRollbackTimeoutKey = @"rollbackTimeout";
             if ([pendingHash isEqualToString:currentHash]) {
                 // We only want to initialize the rollback timer in two scenarios:
                 // 1) The app has been restarted, and therefore, the pending update is already applied
-                // 2) The app has been resumed, and the pending update indicates it supports being restarted on resume
+                // 2) The app has been resumed, and the pending update indicates it supports being restarted on resume, so we need to restart the app and the kickoff the rollback timer
                 if (isAppStart || (!isAppStart && [pendingUpdate[PendingUpdateAllowsRestartOnResumeKey] boolValue]))
                 {
                     int rollbackTimeout = [pendingUpdate[PendingUpdateRollbackTimeoutKey] intValue];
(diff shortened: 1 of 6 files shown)


CodePush.m
```diff
@@ -146,13 +190,17 @@ - (void)saveFailedUpdate:(NSString *)packageHash {
...
+     allowRestartOnResume:(BOOL)allowRestartOnResume
+{
     // Since we're not restarting, we need to store the fact that the update
     // was applied, but hasn't yet become "active".
     NSUserDefaults *preferences = [NSUserDefaults standardUserDefaults];
     NSDictionary *pendingUpdate = [[NSDictionary alloc] initWithObjectsAndKeys:
                                    packageHash,@"hash",
-                                   rollbackTimeout,@"rollbackTimeout", nil];
+                                   [NSNumber numberWithInt:rollbackTimeout],@"rollbackTimeout",
+                                   [NSNumber numberWithBool:allowRestartOnResume],@"allowRestartOnResume",
```
geof90: Any way to const-tize the keys "rollbackTimeout" and "allowRestartOnResume"?

lostintangent: Good idea. I'm const-ify those two and the "hash" key while I'm at it.

geof90:
LGTM, it is a great feature to have.

CodePush.m
```diff
@@ -56,38 +64,83 @@ - (void)cancelRollbackTimer
...
+    return @{ @"codePushRestartModeNone": @(CodePushRestartModeNone),
+              @"codePushRestartModeImmediate": @(CodePushRestartModeImmediate),
+              @"codePushRestartModeOnNextResume": @(CodePushRestartModeOnNextResume)
+              };
+};
+
+- (void)dealloc
+{
+    // Ensure the global resume handler is cleared, so that
+    // this object isn't kept alive unneccesarily
```
dtivel: unnecessarily

dtivel:
Looks good.

PR #55, opened 2015-11-18
Updated sync

geof90:
Couple of big (breaking) changes in this PR:

1. Sync now takes in two function parameters, syncStatusChangeCallback and downloadProgressCallback. These are called periodically during the sync process to report the current state the sync action is in.
2. Sync options have changed
  - It includes an option for controlling when the app should restart when the downloaded update has been installed.
  - The original options for customizing the strings that show up in the dialog is now moved to a nested object parameter. If the object parameter value passed in is falsey, the dialog will not be displayed to the user at all (it will be a silent update where the download happens automatically).
3. Renamed localPackage.**apply** to localPackage.**install**. The action of saving an updated package to the location where the app recognizes to contain the latest version of the app will now be called install instead of apply.

Updated the example app to use Sync and demonstrate its capabilities.

Updated the docs.

syncstatus


package-mixins.js
```diff
@@ -36,8 +36,8 @@ module.exports = (NativeCodePush) => {
   };
 
   var local = {
-    apply: function apply(rollbackTimeout = 0, restartMode = NativeCodePush.codePushRestartModeImmediate) {
-      return NativeCodePush.applyUpdate(this, rollbackTimeout, restartMode);
+    install: function install(rollbackTimeout = 0, installMode = NativeCodePush.codePushInstallModeImmediate) {
+      return NativeCodePush.installUpdate(this, rollbackTimeout, installMode);
```
lostintangent: Shou do we change the default install mode to on next restart here as well?

README.md
```diff
@@ -229,30 +244,30 @@ The `sync` method can be called anywhere you'd like to check for an update. That
...
 <br /><br />
 If a value greater than zero is provided to the `rollbackTimeout` parameter, the application will wait for the `notifyApplicationReady` method to be called for the given number of milliseconds.
 <br /><br />
-Note: The "rollback timer" doesn't start until the update has actually become active. If you pass a truthy value to the `restartImmediately` parameter, then the rollback timer will also start immediately. However, if you pass a falsey value, then the rollback timer will start the next time the app starts, not at the point that you called `apply`.
+Note: The "rollback timer" doesn't start until the update has actually become active. If you pass a truthy value to the `restartImmediately` parameter, then the rollback timer will also start immediately. However, if you pass a falsey value, then the rollback timer will start the next time the app starts, not at the point that you called `install`.
```
lostintangent: This should reference the install mode not the restartImmediately parameter.

README.md
```diff
@@ -229,30 +244,30 @@ The `sync` method can be called anywhere you'd like to check for an update. That
...
 
 ##### Methods
-- __apply(rollbackTimeout: Number = 0, restartImmediately: Boolean = true): Promise&lt;void&gt;__: Applies this package to the application by unzipping its contents (e.g. the JS bundle) and saving it to the location on disk where the runtime expects to find the latest version of the app. If the `restartImmediately` parameter is set to `false`, the apply will complete, but it won't take effect until the next time that the app is restarted. Otherwise, the app will be immediately restarted after performing the apply, so that the end-user sees the changes.
+- __install(rollbackTimeout: Number = 0, installMode: CodePush.InstallMode = CodePush.InstallMode.UPDATE_ON_RESTART): Promise&lt;void&gt;__: Installs this package to the application by unzipping its contents (e.g. the JS bundle) and saving it to the location on disk where the runtime expects to find the latest version of the app. If the `restartImmediately` parameter is set to `false`, the install will complete, but it won't take effect until the next time that the app is restarted. Otherwise, the app will be immediately restarted after performing the install, so that the end-user sees the changes.
```
lostintangent: Update restartImmediately mention

CodePush.ios.js
```diff
@@ -126,72 +126,106 @@ function checkForUpdate() {
  * releases, and displaying a standard confirmation UI to the end-user
  * when an update is available.
  */
-function sync(options = {}) {  
+function sync(options = {}, onSyncStatusChange, onDownloadProgress) {
```
lostintangent: Since these are callbacks not events, do we want to drop the "on" prefix and add the "Callback" suffix (e.g. syncStatusChangeCallback, downloadProgressCallback).


If you agree, we shou do also update the readme as well.

CodePush.ios.js
```diff
@@ -126,72 +126,106 @@ function checkForUpdate() {
...
     ...options 
   };
-
+  
+  onSyncStatusChange = typeof onSyncStatusChange == "function"
+    ? onSyncStatusChange
+    : function(syncStatus) {
+        switch(syncStatus) {
+          case CodePush.SyncStatus.CHECKING_FOR_UPDATE:
+            console.log("Checking for update.");
```
lostintangent: Could we add a "[CodePush]" prefix to these log messages so that users could easily filter them in or out in their dev console?

lostintangent:
This change looks great! Thanks for taking care of this big refactor :+1:

lostintangent:
LGTM!

CodePush.ios.js
```diff
@@ -126,72 +126,106 @@ function checkForUpdate() {
...
  */
-function sync(options = {}) {  
+function sync(options = {}, syncStatusChangeCallback, downloadProgressCallback) {  
   var syncOptions = {
-    descriptionPrefix: " Description: ",
-    appendReleaseDescription: false,
     
     ignoreFailedUpdates: true,
-    
-    mandatoryContinueButtonLabel: "Continue",
```
dlebu: Do we want to leave all these default messages in and use then in the case `updateDialog` is not null but the user did not specify any messages/title/labels?

CodePush.ios.js
```diff
@@ -126,72 +126,106 @@ function checkForUpdate() {
...
     ...options 
   };
-
+  
+  syncStatusChangeCallback = typeof syncStatusChangeCallback == "function"
+    ? syncStatusChangeCallback
+    : function(syncStatus) {
+        switch(syncStatus) {
+          case CodePush.SyncStatus.CHECKING_FOR_UPDATE:
+            console.log("[CodePush] Checking for update.");
```
dlebu: Having a separate method that handles logging will make it easier if we want to change the `[CodePush]` tag in the future.

CodePush.ios.js
```diff
@@ -126,72 +126,106 @@ function checkForUpdate() {
...
+          case CodePush.SyncStatus.DOWNLOADING_PACKAGE:
+            console.log("[CodePush] Downloading package.");
+            break;
+          case CodePush.SyncStatus.AWAITING_USER_ACTION:
+            console.log("[CodePush] Awaiting user action.");
+            break;
+          case CodePush.SyncStatus.INSTALLING_UPDATE:
+            console.log("[CodePush] Installing update.");
+            break;
+          case CodePush.SyncStatus.IDLE:
```
dlebu: Does this status bring any value? `UPDATE_INSTALLED` will be send immediately anyway.

dlebu: Also, I feel like it's hard to keep track of `SyncStatus` vs `SyncResult` names. Would it make sense to combine the two enums into one, and send all of them in the progress callback, but only the final ones in the result? That way, we avoid situations like this one where we have to come up with two names for the same state. (`IDLE` vs `UPDATE_INSTALLED`)

CodePush.ios.js
geof90: They are, the line `syncOptions.updateDialog = Object.assign(CodePush.DEFAULT_UPDATE_DIALOG, syncOptions.updateDialog);` handles that.

CodePush.ios.js
```diff
@@ -206,16 +240,33 @@ var CodePush = {
...
+    UPDATE_INSTALLED: 2 // The app had an optional/mandatory update that was successfully downloaded and is about to be installed.
+  },
+  SyncStatus: {
+    CHECKING_FOR_UPDATE: 0,
+    AWAITING_USER_ACTION: 1,
+    DOWNLOADING_PACKAGE: 2,
+    INSTALLING_UPDATE: 3,
+    IDLE: 4
+  },
+  DEFAULT_UPDATE_DIALOG: {
```
dlebu: Ignore my previous comment about the dialog. I see it now. :+1:

CodePush.ios.js
dlebu: Sounds good. I just saw it and added a comment. :+1:

CodePush.ios.js
```diff
@@ -206,16 +240,33 @@ var CodePush = {
...
   sync: sync,
-  RestartMode: {
-    NONE: NativeCodePush.codePushRestartModeNone, // Don't artificially restart the app. Allow the update to be "picked up" on the next app restart
-    IMMEDIATE: NativeCodePush.codePushRestartModeImmediate, // Restart the app immediately
-    ON_NEXT_RESUME: NativeCodePush.codePushRestartModeOnNextResume // Restart the app the next time it is resumed from the background
+  InstallMode: {
+    IMMEDIATE: NativeCodePush.codePushInstallModeImmediate, // Restart the app immediately
+    ON_NEXT_RESTART: NativeCodePush.codePushInstallModeOnNextRestart, // Don't artificially restart the app. Allow the update to be "picked up" on the next app restart
+    ON_NEXT_RESUME: NativeCodePush.codePushInstallModeOnNextResume // Restart the app the next time it is resumed from the background
   },
```
dlebu: Thanks for changing the names here. These are awesome!

CodePush.ios.js
dlebu: Another question regarding this. What if the dialog object is set, but only certain properties assigned? Will we get the default for the ones that are not set?

dlebu: Nevermind - I was under the impression that the `updateDialog` is a `boolean` flag, not the real dialog. In this case, the code you pointed me to earlier does just what I asked.

Examples/CodePushDemoApp/index.ios.js
```diff
@@ -20,49 +20,98 @@ var CodePush = require('react-native-code-push');
...
-    this.checkUpdate();
   },
-  checkUpdate: function() {
-    CodePush.checkForUpdate().done((update) => {
-      this.setState({ update: update });
+  sync: function() {
+    var self = this;
+    CodePush.sync(
+      { 
+        updateNotification: true,
```
dlebu: Are we still using `updateNotification`? What is the behavior here?

dlebu:
Minor comments, otherwise LGTM. The only thing I could see simplifying here is `SyncStatus` vs `SyncResult` enums.

geof90:
@dlebu I have removed SyncResult, it is just SyncStatus now. I also added an UNKNOWN_ERROR flag to SyncStatus.

CodePush.ios.js
```diff
@@ -126,75 +131,129 @@ function checkForUpdate() {
...
+          case CodePush.SyncStatus.INSTALLING_UPDATE:
+            log("Installing update.");
+            break;
+          case CodePush.SyncStatus.UP_TO_DATE:
+            log("App is up to date.");
+            break;
+          case CodePush.SyncStatus.UPDATE_IGNORED:
+            log("User cancelled the update.");
+            break;
+          case CodePush.SyncStatus.UPDATE_INSTALLED:
```
geof90: Okay, based on our offline discussion I have changed the behavior to return the UPDATE_INSTALLED status right before the restart. Took me a while because some of my tests were failing. Updated the docs as well.

dlebu: Cool, thanks for making this change. :+1:

dlebu:
LGTM!

PR #56, opened 2015-11-19
Bug fixes with the new sync refactor

lostintangent:
This PR fixes three bugs related to the `sync` refactoring:

1. The `UPDATE_IGNORED` sync status event was never fired to the `syncStatusChangeCallback`
2. The use of `Object.assign` would fail when the `updateDialog` sync options wasn't an object (e.g. you pass it true)
3. The update confirmation title wasn't being displayed due to it not using the new nested `updateDialog` object. While fixing this, I took it upon myself to rename this property to just `title` since `updateDialog.title` seems intutuive enough to not need the second "update" :)
4. Not a bug, but I replaced the use of the `extend` module with `Object.assign`

### DIFF
--- a/CodePush.ios.js
+++ b/CodePush.ios.js
@@ -210,8 +210,7 @@ function sync(options = {}, syncStatusChangeCallback, downloadProgressCallback) 
         }
         else if (syncOptions.updateDialog) {
           // updateDialog supports any truthy value (e.g. true, "goo", 12),
-          // but when we merge it's properties with the default dialog's
-          // properties, it needs to be an object
+          // but we should treat a non-object value as just the default dialog
           if (typeof syncOptions.updateDialog !== "object") {
             syncOptions.updateDialog = CodePush.DEFAULT_UPDATE_DIALOG;
           } else {
--- a/CodePush.ios.js
+++ b/CodePush.ios.js
@@ -214,7 +214,7 @@ function sync(options = {}, syncStatusChangeCallback, downloadProgressCallback) 
           if (typeof syncOptions.updateDialog !== "object") {
             syncOptions.updateDialog = CodePush.DEFAULT_UPDATE_DIALOG;
           } else {
-            syncOptions.updateDialog = Object.assign(CodePush.DEFAULT_UPDATE_DIALOG, syncOptions.updateDialog);
+            syncOptions.updateDialog = Object.assign({}, CodePush.DEFAULT_UPDATE_DIALOG, syncOptions.updateDialog);
           }
           
           var message = null;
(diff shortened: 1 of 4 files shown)


CodePush.ios.js
```diff
@@ -210,7 +209,13 @@ function sync(options = {}, syncStatusChangeCallback, downloadProgressCallback)
           resolve(CodePush.SyncStatus.UP_TO_DATE);
         }
         else if (syncOptions.updateDialog) {
-          syncOptions.updateDialog = Object.assign(CodePush.DEFAULT_UPDATE_DIALOG, syncOptions.updateDialog);
+          // updateDialog supports any truthy value (e.g. true, "goo", 12),
+          // but we should treat a non-object value as just the default dialog
+          if (typeof syncOptions.updateDialog !== "object") {
+            syncOptions.updateDialog = CodePush.DEFAULT_UPDATE_DIALOG;
+          } else {
+            syncOptions.updateDialog = Object.assign(CodePush.DEFAULT_UPDATE_DIALOG, syncOptions.updateDialog);
```
silhouettes: Will we need to re-use CodePush.DEFAULT_UPDATE_DIALOG if we call checkForUpdate() multiple times?

silhouettes:
Looks good!

geof90:
Thanks for fixing my buggy code :+1:

CodePush.ios.js
geof90: checkForUpdate() or sync()? the former doesn't use CodePush.DEFAULT_UPDATE_DIALOG right?

silhouettes: Sorry - yeah I meant sync() :P was just pointing out that DEFAULT_UPDATE_DIALOG gets modified here

geof90: Hmm. That could be a potential bug. I thought it returns a copy but apparently I am wrong.

geof90:
Sorry, are we getting rid of the use of "extend" in package-mixins.js as well?

CodePush.ios.js
lostintangent: I only added the typecheck because it doesn't work :) If you run the app in debug mode with Chrome, I'm assuming that Object.assign is actually the "real" version not a polyfill, and it throws a warning saying that it doesn't accept non-object values.

geof90: To ensure DEFAULT_UPDATE_DIALOG doesn't get modified, we should do:
```
syncOptions.updateDialog = Object.assign({}, CodePush.DEFAULT_UPDATE_DIALOG, syncOptions.updateDialog);
```

And I am pretty sure that the typecheck is not necessary. If the source (syncOptions.updateDialog) does not contain any properties, i.e it is an integer or a boolean, it will just be ignored.

geof90:
LGTM :+1:

geof90:
Probably remove extend from package.json as well since we no longer need it?

Issue #45, opened 2015-11-11
Ability to customize binary bundle location


nevir:
I was a bit surprised that [`main.jsbundle` is hard coded](https://github.com/Microsoft/react-native-code-push/blob/master/CodePush.m#L31). `[CodePush setBinaryBundleURL:]` maybe?

lostintangent:
Agreed, we should change this. We did this primarily based on the fact that the current prescription leads you down the path of calling the bundle main.jsbundle (e.g. the xcodeproj and Gradle build tasks generate that file name by default, the default templates reference that name). 

We weren't sure how common it was for folks to change this, so we went with the naïve solution first, but definitely want to be fully accommodating moving forward. Out of curiosity, why did you guys customize your bundle file name?

nevir:
Nope! Still using `main.jsbundle` ourselves, as well :P We are customizing the development location, however (in order to utilize https://github.com/mjohnston/react-native-webpack-server)

lostintangent:
We just added support for this in the referenced PR. Let me know if this doesn't work for your scenarios. Thanks!

PR #61, opened 2015-11-23
Misc. product updates/improvements

lostintangent:
This pull request introduces the following product improvements:

1. The `getBundleUrl` method of `CodePush` has been renamed to `bundleURL` to be more idiomatic with Obj-C naming conventions, but more importantly, introduces two variants of `bundleURL` that allow overwriting the bundle's file name and/or extension. This fixes issue #45

2. Expose the `CodePush.setDeploymentKey` method to script to allow app's to dynamically update the deployment key at runtime. This is being done based on customer request.

3. Exposes the `CodePush.restartApp` method to script to allow app's to install an update and then arbitrarily choose when to restart. This function already existed, and was exposed to script via `NativeModules`, but it wasn't exported as part of the `react-native-code-push` module

4. Better supports Chrome debugging by not resetting the bundle URL when restarting the app. Currently, when a CodePush update is applied, the native bridge's URL is updated to point at the update on disk, and that causes Chrome to break

### DIFF
--- a/CodePush.h
+++ b/CodePush.h
@@ -2,30 +2,26 @@
 
 @interface CodePush : NSObject<RCTBridgeModule>
 
-+ (NSURL *)getBundleUrl;
++ (NSURL *)bundleURL;
+
++ (NSURL *)bundleURLForResourceName:(NSString *)resourceName;
+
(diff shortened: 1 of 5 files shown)


geof90:
LGTM :+1:

PR #62, opened 2015-11-24
Android support

geof90:
This PR adds support for Android.


CodePush.xcodeproj/project.pbxproj
```diff
@@ -215,9 +215,8 @@
 				HEADER_SEARCH_PATHS = (
 					"$(inherited)",
 					/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/include,
-					"$(SRCROOT)/../../React/**",
-					"$(SRCROOT)/../../node_modules/react-native/React/**",
 					"$(SRCROOT)/node_modules/react-native/React/**",
+					"$(SRCROOT)/Examples/CodePushDemoApp/node_modules/react-native/React/**",
```
lostintangent: Why did you make this change?

CodePushNativePlatformAdapter.js
```diff
@@ -0,0 +1,67 @@
...
+    "downloadUpdate",
+    "getConfiguration",
+    "getCurrentPackage",
+    "isFailedUpdate",
+    "isFirstRun",
+    "notifyApplicationReady",
+    "setDeploymentKey"
+  ];
+  
+  methodsToPromisify.forEach((methodName) => {
```
lostintangent: Do we need to do this now that https://github.com/facebook/react-native/commit/b86a6e3b44a63e92cf3a7976d2fa26c4bf412df1#commitcomment-14069903 went in to master?

README.md
```diff
@@ -82,6 +102,60 @@ To let the CodePush runtime know which deployment it should query for updates ag
...
+    
+    // 2. Optional: extend FragmentActivity if you intend to show a dialog prompting users about updates.
+    public class MainActivity extends FragmentActivity implements DefaultHardwareBackBtnHandler {
+        ...
+        
+        @Override
+        protected void onCreate(Bundle savedInstanceState) {
+            ...
+            // 3. Initialize CodePush with your deployment key and an instance of your MainActivity.
+            // You can also set the deployment key in code by assigning the key to the `[CodePushConfig current].deploymentKey` property.*
```
lostintangent: This comment line doesn't make sense :)

android/app/src/main/java/com/microsoft/reactnativecodepush/CodePush.java
```diff
@@ -0,0 +1,412 @@
+package com.microsoft.reactnativecodepush;
```
lostintangent: Should this just be "com.microsoft.codepush"?

CodePush.xcodeproj/project.pbxproj
geof90: The deleted lines were wrong I think. $(SRCROOT) in the CodePush project is relative to itself. Having "$(SRCROOT)/node_modules/react-native/React/**" is sufficient. The added line supports loading from the Example project's node_modules in the case where the example project is the source root.

CodePushNativePlatformAdapter.js
geof90: Maybe not. I'll try to see if it works.

android/app/src/main/java/com/microsoft/reactnativecodepush/CodePush.java
```diff
@@ -0,0 +1,412 @@
...
+    public String getDocumentsDirectory() {
+        return codePushPackage.getDocumentsDirectory();
+    }
+
+    public String getBundleUrl(String assetsBundleFileName) {
+        this.assetsBundleFileName = assetsBundleFileName;
+        String binaryJsBundleUrl = ASSETS_BUNDLE_PREFIX + assetsBundleFileName;
+        try {
+            String packageFile = codePushPackage.getCurrentPackageBundlePath();
+            if (packageFile == null) {
```
lostintangent: Do we also need to check whether the binary bundle is newer?

android/app/src/main/java/com/microsoft/reactnativecodepush/CodePush.java
```diff
@@ -0,0 +1,412 @@
...
+        if (pendingUpdateString != null) {
+            try {
+                JSONObject pendingUpdateJSON = new JSONObject(pendingUpdateString);
+                String pendingHash = pendingUpdateJSON.getString(PENDING_UPDATE_HASH_KEY);
+                String currentHash = codePushPackage.getCurrentPackageHash();
+                if (pendingHash.equals(currentHash)) {
+                    int rollbackTimeout = pendingUpdateJSON.getInt(PENDING_UPDATE_ROLLBACK_TIMEOUT_KEY);
+                    initializeUpdateWithRollbackTimeout(rollbackTimeout, needsRestart);
+
+                    settings.edit().remove(PENDING_UPDATE_KEY).commit();
```
lostintangent: We should always remove the pending update, and add an assert that the current and pending hash matches, instead of using the if statement above.

android/app/src/main/java/com/microsoft/reactnativecodepush/CodePush.java
```diff
@@ -0,0 +1,412 @@
...
+                        pendingUpdateString + " stored in SharedPreferences", e);
+            } catch (IOException e) {
+                // There is no current package hash.
+                throw new CodePushUnknownException("Should not register a pending update without a saving a current package", e);
+            }
+        }
+    }
+
+    private void checkForPendingUpdateDuringResume() {
+        if (resumablePendingUpdateAvailable) {
```
lostintangent: Does Java use underscores as a convention for naming instance variables?

android/app/src/main/java/com/microsoft/reactnativecodepush/CodePush.java
```diff
@@ -0,0 +1,412 @@
...
+        }, timeout);
+    }
+
+    private class CodePushNativeModule extends ReactContextBaseJavaModule {
+
+        private void loadBundle() {
+            String assetsBundleFileUrl = CodePush.this.getBundleUrl(CodePush.this.assetsBundleFileName);
+            Intent intent = mainActivity.getIntent();
+            mainActivity.finish();
+            mainActivity.startActivity(intent);
```
lostintangent: Does the RN bridge for Android not have a way to reload itself? Also, I don't see where you're using the `assetsBundleFileUrl` variable.

android/app/src/main/java/com/microsoft/reactnativecodepush/CodePush.java
```diff
@@ -0,0 +1,412 @@
...
+        }
+    }
+
+    private void checkForPendingUpdateDuringResume() {
+        if (resumablePendingUpdateAvailable) {
+            checkForPendingUpdate(/*needsRestart*/ true);
+        }
+    }
+
+    private WritableMap constantsToExport() {
```
lostintangent: What is this method for? Doesn't the override in the native module below handle this?

android/app/src/main/java/com/microsoft/reactnativecodepush/CodePush.java
```diff
@@ -0,0 +1,412 @@
...
+                @Override
+                public void onHostDestroy() {
+                }
+            });
+
+            return nativeModules;
+        }
+
+        @Override
+        public List<Class<? extends JavaScriptModule>> createJSModules() {
```
lostintangent: Do we need to override this method since we aren't actually using it?

android/app/src/main/java/com/microsoft/reactnativecodepush/CodePushConfig.java
```diff
@@ -0,0 +1,69 @@
...
+
+    public void setServerUrl(String serverUrl) {
+        this.serverUrl = serverUrl;
+    }
+
+    public String getServerUrl() {
+        return serverUrl;
+    }
+
+    public void setAppVersion(String appVersion) {
```
lostintangent: This shouldn't be writable, along with build version as well, since it doesn't make any sense to mutate these values.

android/app/src/main/java/com/microsoft/reactnativecodepush/CodePushConfig.java
```diff
@@ -0,0 +1,69 @@
+package com.microsoft.reactnativecodepush;
+
+import android.content.Context;
+import android.content.pm.PackageInfo;
+import android.content.pm.PackageManager;
+
+import com.facebook.react.bridge.ReadableMap;
+import com.facebook.react.bridge.WritableNativeMap;
+
+public class CodePushConfig {
```
lostintangent: I would be in favor of completely removing this class. I don't think it adds enough value to exist separately from the `Codepush` class. I eventually want to refactor it out of the Obj-c implementation.

android/app/src/main/java/com/microsoft/reactnativecodepush/CodePush.java
geof90: In case we have a native code push SDK, maybe its better to keep the react native prefix there.

package-mixins.js
```diff
@@ -1,4 +1,13 @@
-var { NativeAppEventEmitter } = require("react-native");
+var Platform = require("Platform");
+var EventEmitter;
+
+if (Platform.OS === "android") {
+  var { DeviceEventEmitter } = require("react-native");
+  EventEmitter = DeviceEventEmitter;
+} else if (Platform.OS === "ios") {   
+  var { NativeAppEventEmitter } = require("react-native");
+  EventEmitter = NativeAppEventEmitter;
```
lostintangent: I thought iOS supports `DeviceEventEmitter` as well. Could we use that for both?

android/app/src/main/java/com/microsoft/reactnativecodepush/CodePush.java
geof90: I wanted to but I think we can't construct a File object for a file in the assets folder, its a packaged apk file, so I can't find any way to retrieve the date.

package.json
```diff
@@ -19,6 +19,6 @@
     "code-push": "^1.1.1-beta"
   },
   "devDependencies": {
-    "react-native": "0.11.4"
+    "react-native": "0.14.2"
```
lostintangent: Do we need to update this to 0.15.0? Or does this version support everything we need?

package-mixins.js
geof90: This must have changed in a recent release. I'll try that out

android/app/src/main/java/com/microsoft/reactnativecodepush/CodePush.java
geof90: We are implementing an interface, so if you do not override the method, the compiler will throw an error.

android/app/src/main/java/com/microsoft/reactnativecodepush/CodePushInstallMode.java
```diff
@@ -0,0 +1,15 @@
+package com.microsoft.reactnativecodepush;
+
+public enum CodePushInstallMode {
+    IMMEDIATE(0),
```
lostintangent: Do we need to explicitly set the numeric values? Won't they be 0, 1, and 2 naturally?

android/app/src/main/java/com/microsoft/reactnativecodepush/CodePush.java
geof90: Not that I know of. Let me know if you found a way to do it. The assetsBundleFileUrl was leftover when I tried to find a way to do just that - invoke the bridge to navigate to the script. I will remove it.

android/app/src/main/java/com/microsoft/reactnativecodepush/CodePush.java
geof90: Its not a hard rule. Consistency is the main concern. I will be happy to add underscores if you think it is important.

android/app/src/main/java/com/microsoft/reactnativecodepush/CodePush.java
geof90: Will throw exception if it is different instead. http://stackoverflow.com/questions/23113497/what-are-buildconfig-debug-conditional-checks

android/app/src/main/java/com/microsoft/reactnativecodepush/CodePush.java
geof90: This should be removed.

android/app/src/main/java/com/microsoft/reactnativecodepush/CodePush.java
lostintangent: I think we need to figure out a way to do this before merging this in, since giving precedence to the binary version is important for the app store update scenario.

package.json
geof90: actually, maybe lets revert it to 0.11.4, since this only affects iOS.

android/app/src/main/java/com/microsoft/reactnativecodepush/CodePush.java
geof90: Cordova side uses the date of the entire resources folder for comparison. I just fixed it to follow what we did there.

package-mixins.js
geof90: Got DeviceEventEmitter to work on iOS :)

README.md
```diff
@@ -44,7 +44,27 @@ Add a new value, `$(SRCROOT)/../node_modules/react-native-code-push` and select
...
+
+To integrate CodePush into your Android project, do the following steps:
+
+1. In your `android/settings.gradle` file, make the following additions:
+    
+    ```gradle
+    include ':app', ':react-native-code-push'
+    project(':react-native-code-push').projectDir = new File(rootProject.projectDir, '../node_modules/react-native-code-push/android/app')
+    ```
+2. In your `android/app/build.gradle` file, add CodePush as one of the dependencies:
```
lostintangent: How about...'add the `:react-native-code-push` project as a compile-time dependency:'

README.md
```diff
@@ -75,6 +95,58 @@ To let the CodePush runtime know which deployment it should query for updates ag
 1. Open your app's `Info.plist` and add a new `CodePushDeploymentKey` entry, whose value is the key of the deployment you want to configure this app against (e.g. the Staging deployment for FooBar app). You can retreive this value by running `code-push deployment ls <appName>` in the CodePush CLI, and copying the value of the `Deployment Key` column which corresponds to the deployment you want to use. 
 2. In your app's `Info.plist` make sure your `CFBundleShortVersionString` value is a valid [semver](http://semver.org/) version (e.g. 1.0.0 not 1.0)
 
+## Plugin Configuration - Android
+
+After installing the plugin and sync-ing your Android Studio project with Gradle, you need to configure your app to consult CodePush for the location of your JS bundle, since it will "take control" of managing the current and all future versions. To do this, perform the following steps:
```
lostintangent: syncing?

README.md
```diff
@@ -75,6 +95,58 @@ To let the CodePush runtime know which deployment it should query for updates ag
...
+                    .setJSBundleFile(codePush.getBundleUrl("index.android.bundle"))
+                    
+                    // 6. Expose the CodePush module to JavaScript.
+                    .addPackage(codePush.getReactPackage())
+                    ...
+        }
+    }
+    ```
+
+2. Let the CodePush runtime know which deployment it should query for updates against. Be sure to set the `versionName` in your `android/app/build.gradle`:
```
lostintangent: The deployment key was configured above. Should we change this to simply. "Ensure that the `versionName` property in your `android/app/build.gradle` file is set to a semver compliant value."

README.md
```diff
@@ -75,6 +95,58 @@ To let the CodePush runtime know which deployment it should query for updates ag
 1. Open your app's `Info.plist` and add a new `CodePushDeploymentKey` entry, whose value is the key of the deployment you want to configure this app against (e.g. the Staging deployment for FooBar app). You can retreive this value by running `code-push deployment ls <appName>` in the CodePush CLI, and copying the value of the `Deployment Key` column which corresponds to the deployment you want to use. 
 2. In your app's `Info.plist` make sure your `CFBundleShortVersionString` value is a valid [semver](http://semver.org/) version (e.g. 1.0.0 not 1.0)
 
+## Plugin Configuration - Android
+
+After installing the plugin and sync-ing your Android Studio project with Gradle, you need to configure your app to consult CodePush for the location of your JS bundle, since it will "take control" of managing the current and all future versions. To do this, perform the following steps:
+
+1. Initialize the module in MainActivity.java:
```
lostintangent: Maybe change this to "Update the `MainActivity.java` file to use CodePush"?

lostintangent:
LGTM!

PR #63, opened 2015-11-24
Removing 'Name' suffix

lostintangent:
This PR simply removes the `Name` suffix from the new `bundleURLForResource` methods, which aligns more closely with the `URLForResource` method of the `NSBundle` class, which these are effectively mirroring.


geof90:
LGTM :+1:

PR #64, opened 2015-11-24
Updating restartApp and setDeploymentKey

lostintangent:
This PR makes two changes:

1. Removes the previously added `setDeploymentKey` method, and replaces it with a new `deploymentKey` parameter on the `sync` and `checkForUpdate` methods

2. Creates a new `restartApp` method that is more appropriate for public consumption, as opposed to the previous version which was meant exclusively for internal usage by the immediate installation experience

### DIFF
--- a/CodePush.ios.js
+++ b/CodePush.ios.js
@@ -1,20 +1,10 @@
 'use strict';
 
+var { AlertIOS } = require("react-native");
 var NativeCodePush = require("react-native").NativeModules.CodePush;
+var packageMixins = require("./package-mixins")(NativeCodePush);
 var requestFetchAdapter = require("./request-fetch-adapter.js");
 var Sdk = require("code-push/script/acquisition-sdk").AcquisitionManager;
-var packageMixins = require("./package-mixins")(NativeCodePush);
-
-var { AlertIOS } = require("react-native");
-
-// This function is only used for tests. Replaces the default SDK, configuration and native bridge
-function setUpTestDependencies(providedTestSdk, providedTestConfig, testNativeBridge){
-  if (providedTestSdk) testSdk = providedTestSdk;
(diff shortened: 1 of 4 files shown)


geof90:
So restartApp will not actually restart the app if there's no pending update?

lostintangent:
Correct. Please suggest a better name :) The reason I did it this way is because the primary scenario it is trying to help with is letting you "restart early" for a previously installed update. It's not meant as a general purpose way to restart the app, since that wouldn't really be that useful.

geof90:
How about "loadInstalledBundle()"?

lostintangent:
What do you think of the latest change? I removed the global `restartApp` method and instead, moved it to the `LocalPackage` object instance after a non-immediate update is installed. This keeps the global API smaller. Currently, this solution would only work for `checkForUpdate` workflows, which I think is fine for now, since this is a pretty advanced scenario.

geof90:
Having it part of the global API feels more natural to me. "Booting up a pending bundle" sounds like an action that is performed by the global CodePush system rather than something performed on the localpackage.

lostintangent:
How about `restartPendingUpdate`? I like the term "PendingUpdate" better than "InstalledBundle". "Installed" doesn't necessarily mean pending, and I think the "pending" part is important. We don't currently use the term "Bundle" on the JS-side of the API, and now with Android support, I'd rather not introduce it, since this is more of an iOS name. I also prefer "restart" over "load" since it seems a little more clear that the app will restart, whereas "load" is a little ambiguous.

geof90:
restartPendingUpdate sounds good to me!

lostintangent:
Is that a sign-off? :)

geof90:
yup LGTM

PR #68, opened 2015-11-24
Fixing caching issue with specifying a deployment key

lostintangent:
The `getSDK` function was caching its results, which would result in there being issues with trying to override the deployment key in one call to `sync`/`checkForUpdate`, and then wanting to rely on the "default" key on a subsequent call. This change removes the caching behavior completely from `getSDK` since it isn't really needed, and its updates the `checkForUpdate` method to make a copy on the `config` object when it modifies it in place, as opposed to updating the instance it gets back from the `getConfiguration` promise.

### DIFF
--- a/CodePush.ios.js
+++ b/CodePush.ios.js
@@ -7,24 +7,20 @@ var requestFetchAdapter = require("./request-fetch-adapter.js");
 var Sdk = require("code-push/script/acquisition-sdk").AcquisitionManager;
 
 function checkForUpdate(deploymentKey = null) {
-  var config;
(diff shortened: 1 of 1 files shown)


silhouettes:
Would it make more sense to pass the deploymentKey to queryUpdateWithCurrentPackage() in the SDK, rather than modifying the config and re-constructing it each time?

silhouettes:
Or yeah, I guess this is fine, I'm aware that we're trying to make small incremental changes without turning things into a big refactor!

PR #71, opened 2015-11-26
Fix rollback

geof90:


CodePush.m
```diff
@@ -390,10 +354,18 @@ - (void)startRollbackTimer:(int)rollbackTimeout
...
         } else {
-            if (installMode != CodePushInstallModeImmediate) {
-                _resumablePendingUpdateAvailable = (installMode == CodePushInstallModeOnNextResume);
-                [self savePendingUpdate:updatePackage[@"packageHash"]
-                        rollbackTimeout:rollbackTimeout];
+            [self savePendingUpdate:updatePackage[@"packageHash"]
+                    rollbackTimeout:rollbackTimeout];
+            
+            if (installMode == CodePushInstallModeImmediate) {
+                [self restartPendingUpdate];
```
lostintangent: We shouldn't need to check for the pending update before restarting here.

CodePush.m
```diff
@@ -390,10 +354,18 @@ - (void)startRollbackTimer:(int)rollbackTimeout
...
+            } else if (installMode == CodePushInstallModeOnNextResume) {
+                // Register for app resume notifications so that we
+                // can check for pending updates which support "restart on resume"
+                [[NSNotificationCenter defaultCenter] addObserver:self
+                                                         selector:@selector(restartPendingUpdate)
```
lostintangent: Same as above, shouldn't the resume handler simply restart?

CodePush.m
```diff
@@ -165,19 +114,34 @@ - (instancetype)init
...
-                                                     name:UIApplicationWillEnterForegroundNotification
-                                                   object:[UIApplication sharedApplication]];
+        [self handleInitIfPendingUpdate];
     }
     
     return self;
 }
 
+
+- (void)handleInitIfPendingUpdate
```
lostintangent: What about "initializeUpdateAfterRestart" or something like that?

CodePush.m
```diff
@@ -7,7 +7,7 @@
 #import "CodePush.h"
 
 @implementation CodePush {
-    BOOL _resumablePendingUpdateAvailable;
+    BOOL hasRestartListener;
```
lostintangent: Should this be `hasResumeListener`? Also, can we add the underscore prefix? :)

lostintangent:
LGTM!

PR #73, opened 2015-11-26
Delete old packages

geof90:
This PR

1. Clears older packages beyond the "previousPackage"
2. Fixes the header search path on iOS for NPM v3
3. Fixes a "require" issue with the "Platform" module which is a submodule of React
4. Fixes a bug with "install immediate" on android

### DIFF
--- a/AlertAdapter.js
+++ b/AlertAdapter.js
@@ -1,6 +1,6 @@
 'use strict';
 
-var Platform = require("react-native").Platform;
+var { Platform } = require("react-native");
 var Alert;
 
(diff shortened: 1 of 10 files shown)


AlertAdapter.js
```diff
@@ -1,6 +1,6 @@
 'use strict';
 
-var Platform = require("Platform");
+var Platform = require("react-native").Platform;
```
lostintangent: var { Platform } ?

lostintangent:
LGTM!

PR #74, opened 2015-11-26
Fixing isFirstRun on iOS after calling restartApp

lostintangent:
The `didUpdate` global variable was never being reset to `NO` and therefore, if you called `restartApp` after an `IMMEDIATE` or `ON_NEXT_RESUME` update was applied, the `LocalPackage.isFirstRun` property would incorrectly be `true`. This change simply makes the `didUpdate` flag an ivar so that every time the app is restarted, it will be reset to `NO` and only set to `YES` when there was actually an update applied before the restart occurred. I also took it upon myself to rename the variable to better reflect its semantics.

This issue doesn't impact the Android implementation since when a reload happens, the `CodePush` class is collected along with the `MainActivity`, and therefore, doesn't rely on any global state like the iOS version was.

### DIFF
--- a/CodePush.m
+++ b/CodePush.m
@@ -8,11 +8,11 @@
 
 @implementation CodePush {
     BOOL _hasResumeListener;
+    BOOL _isFirstRunAfterUpdate;
 }
 
 RCT_EXPORT_MODULE()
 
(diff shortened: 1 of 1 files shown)


geof90:
LGTM!

PR #75, opened 2015-11-28
fix mandatoryContinueButtonLabel

oney:
Just remind you. You can fix it by yourself.


lostintangent:
Thanks for the catch @oney! This fix looks good.

PR #76, opened 2015-11-28
Download file in background on Android

oney:
On Android, downloading file will block UI interactions and animations now. We should download file in background.

### DIFF
--- a/android/app/src/main/java/com/microsoft/codepush/react/CodePush.java
+++ b/android/app/src/main/java/com/microsoft/codepush/react/CodePush.java
@@ -23,6 +23,7 @@ import android.content.SharedPreferences;
 import android.content.pm.ApplicationInfo;
 import android.content.pm.PackageInfo;
 import android.content.pm.PackageManager;
+import android.os.AsyncTask;
 
 import org.json.JSONException;
 import org.json.JSONObject;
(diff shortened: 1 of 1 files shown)


lostintangent:
This looks good to me! CCing @geof90 to sign off and merge. Thanks @oney for two contributions in one day :+1:

geof90:
Hi @oney,

Are you absolutely certain that the code in the native call is executed on the main thread? To check if this is true, can you insert this piece of code in the method and see what appears in your console: 

```
                        if (Looper.myLooper() == Looper.getMainLooper()) {
                            System.out.println("ON MAIN THREAD");
                        } else {
                            System.out.println("NOT ON MAIN THREAD");
                        }
```

"NOT ON MAIN THREAD" was what appeared for me.

oney:
It showed `NOT ON MAIN THREAD`.
But the weird thing is that UI interactions and animations were blocked if downloading was not in AsyncTask.

geof90:
I see. If it is indeed blocking animations, then this fix is needed. I assumed that if it is not on the main thread, it would not block the JS thread or the main thread, which is why I left out all the async dispatches when I was porting the iOS code over to android. Perhaps there is some thread signalling going on under the hood. I will go ahead and merge your change. Thanks!

PR #77, opened 2015-11-28
Asyncify methods

geof90:
Asyncifies other methods that perform file IO to prevent them from blocking animations. These methods are also async on the iOS side.

Bumped the version to 1.3.1.


lostintangent:
LGTM!

PR #78, opened 2015-11-29
Prevent error in deleting old packages from failing the install operation.

geof90:
Change is only to iOS, Android doesnt seem to face this issue.


lostintangent:
LGTM!

PR #81, opened 2015-12-02
Avoid dismissing alert dialog when the user touches the screen

oney:
When update is mandatory, it doesn't make sense that the users can touch the screen to dismiss alert.


lostintangent:
This change makes a lot of sense and looks good to me!

dtivel:
Looks good.

Issue #84, opened 2015-12-03
Application crashes after syncing new version


ajwhite:
I've added the following to my `componentDidMount`

```js
CodePush.sync({
  updateDialog: true,
  installMode: CodePush.InstallMode.ON_NEXT_RESTART
});
```

The following behavior is experienced:

1. When the application loads, the update prompt is presented
2. The update prompt is confirmed
3. Manually restart application
4. Application loads with new version
5. Update prompt activates *again*
6. App continuously reloads in the background (without confirming the repeated dialog)
7. Manually restart application
8. Application crashes and cannot open with the following log message:

```
Dec  3 12:34:56 Atticuss-MacBook-Pro RobinRooms[21558]: Deriving mainBundle URL...
Dec  3 12:34:56 Atticuss-MacBook-Pro RobinRooms[21558]: file:///Users/ajwhite/Library/Developer/CoreSimulator/Devices/B1FAC6C9-136F-4027-ADC8-3F1EFDC6283A/data/Containers/Data/Application/F0E5D633-B97A-4677-A4E6-BBF8C9584FB1/Documents/CodePush/3b32e6b0b45734580c78111f704a611c41d0edb1b1211bf75970c73154982aa1/app.jsbundle
Dec  3 12:34:56 Atticuss-MacBook-Pro assertiond[4384]: assertion failed: 15B42 13B137: assertiond + 13207 [2F0DDEBE-B278-35C0-8523-3236DA94649E]: 0x1
Dec  3 12:34:56 --- last message repeated 2 times ---
Dec  3 12:34:56 Atticuss-MacBook-Pro RobinRooms[21558]: SyntaxError: Unexpected end of script
Dec  3 12:34:56 Atticuss-MacBook-Pro RobinRooms[21558]: *** Terminating app due to uncaught exception 'RCTFatalException', reason: 'Message: SyntaxError: Unexpected end of script'
	*** First throw call stack:
	(
		0   CoreFoundation                      0x01464a84 __exceptionPreprocess + 180
		1   libobjc.A.dylib                     0x006b0e02 objc_exception_throw + 50
		2   CoreFoundation                      0x014649ad +[NSException raise:format:] + 141
		3   RobinRooms                          0x0008a65b RCTFatal + 358
		4   RobinRooms                          0x000a5db2 -[RCTBatchedBridge stopLoadingWithError:] + 305
[...]

PR #85, opened 2015-12-04
Fixing semver check issue in checkForUpdate

lostintangent:
This PR makes two changes to the `checkForUpdate` method:

1. Fixes the version check to use semver equivalence instead of doing a naive string comparison. This solves an issue where releasing a CodePush update the targets `v1.0.0` would break the comparison, since the "v" prefix wouldn't match the app's current version of `1.0.0`. Using semver comparison "normalizes" the "v" out correctly. This issue was reported via #84.

2. Adds a safety check which resolves the returned `Promise` to `null` if the update it received from the server has the same hash as the currently running update. There is no known scenario where this could happen, but I'd rather err on the side of caution here to make sure that end-users never get unexpected updates.

I also added some descriptive comments to hopefully make the `checkForUpdate` function a little more explainatory, since it's starting to get a little complex.

### DIFF
--- a/CodePush.js
+++ b/CodePush.js
@@ -12,7 +12,7 @@ function checkForUpdate(deploymentKey = null) {
   
   /*
    * Before we ask the server if an update exists, we
-   * need to retreive three pieces of information from the 
+   * need to retrieve three pieces of information from the 
    * native side: deployment key, app version (e.g. 1.0.1)
    * and the hash of the currently running update (if there is one).
    * This allows the client to only receive updates which are targetted
--- a/CodePush.js
+++ b/CodePush.js
@@ -1,19 +1,33 @@
 'use strict';
 
-var requestFetchAdapter = require("./request-fetch-adapter.js");
-var Sdk = require("code-push/script/acquisition-sdk").AcquisitionManager;
+var { Alert } = require("./AlertAdapter");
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -16,9 +16,10 @@
     "url": "https://github.com/Microsoft/react-native-code-push"
   },
   "dependencies": {
-    "code-push": "^1.1.1-beta"
+    "code-push": "^1.1.1-beta",
+    "semver": "^5.1.0"
```
ajwhite: Yeah good call on this dependency, v-flexible to different formats that conform to semver

silhouettes:
Looks good!

CodePush.js
```diff
@@ -1,19 +1,33 @@
...
+var requestFetchAdapter = require("./request-fetch-adapter.js");
+var Sdk = require("code-push/script/acquisition-sdk").AcquisitionManager;
+var semver = require("semver");
 
 function checkForUpdate(deploymentKey = null) {
   var config, sdk;
   
+  /*
+   * Before we ask the server if an update exists, we
+   * need to retreive three pieces of information from the
```
dtivel: retrieve

dtivel:
Looks good.

PR #89, opened 2015-12-04
Delete bundle file if it already exists

geof90:
This bug only occurs when switching between different deployment keys dynamically using the API. Because the packages are stored in folders named using the package hash, it is possible to switch to a different deployment that yields an update with the same package hash as one that is previously installed under a different deployment. In this case, `moveItemAtPath` will fail because the destination file already exists. This fix takes care of that.


lostintangent:
LGTM!

PR #95, opened 2015-12-07
Rollback if the app crashed

oney:
Currently, we can use `rollbackTimeout` parameter in `sync` method to rollback update if it doesn't call `notifyApplicationReady` method.
```javascript
componentDidMount() {
  CodePush.notifyApplicationReady();
  CodePush.sync({
    installMode: CodePush.InstallMode.IMMEDIATE,
    rollbackTimeout: 5000,
  });
},
```
However, if the app crashes with restarting, the rollback timer won't execute. Then, the app will use the broken update in next launch.
The goal of this PR is solving this issue. If the app crashes with restarting, it will know this thing and rollback in next launch. Maybe the codes needs some refactoring, but the concept is there.

### DIFF
--- a/CodePush.m
+++ b/CodePush.m
@@ -15,6 +15,7 @@ RCT_EXPORT_MODULE()
 
 static NSTimer *_timer;
 static BOOL usingTestFolder = NO;
+static BOOL launchingApp = YES;
 
 static NSString *const FailedUpdatesKey = @"CODE_PUSH_FAILED_UPDATES";
 static NSString *const PendingUpdateKey = @"CODE_PUSH_PENDING_UPDATE";
@@ -42,6 +43,10 @@ static NSString *const PendingUpdateRollbackTimeoutKey = @"rollbackTimeout";
                   withExtension:(NSString *)resourceExtension
 {
     NSError *error;
+    if (launchingApp) {
+        launchingApp = NO;
+        [self rollbackIfNotConfirmPackageSuccess];
+    }
     NSString *packageFile = [CodePushPackage getCurrentPackageBundlePath:&error];
     NSURL *binaryJsBundleUrl = [[NSBundle mainBundle] URLForResource:resourceName withExtension:resourceExtension];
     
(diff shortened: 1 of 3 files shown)


lostintangent:
Hey @oney! Thanks for submitting this PR. I've actually been thinking about this experience a lot lately and I agree that we need to improve this. What I've been thinking about is whether we can do away with the `rollbackTimeout`/`notifyApplicationReady` entirely, and instead, have the plugin detect that an error occurred during the app load after a restart, and do the rollback transparently. 

For example, on the iOS-side, could we subscribe to [this](https://github.com/facebook/react-native/blob/master/React/Base/RCTBatchedBridge.m#L447) notification, mark the fact that the current update failed, and then after the app restarts, look for that mark and roll the user back. That would conceptually be similar to what your PR is doing, but it would potentially have the benefit of removing the rollback timeout, and therefore, requiring less work with the same experience.

I haven't tested out this theory or looked into the analogue on the Android side, but I'm really keen on making some progress here. What do you think?

CodePush.m
```diff
@@ -42,6 +43,10 @@ + (NSURL *)bundleURLForResource:(NSString *)resourceName
                   withExtension:(NSString *)resourceExtension
 {
     NSError *error;
+    if (launchingApp) {
```
geof90: Will this work if the install mode is ON_NEXT_RESTART? Because my conjecture is that case this flag will be true on start, hence updates will ALWAYS rollback.

oney: Yes, you are right. It will always rollback in ON_NEXT_RESTART mode. It should record the install mode to handle it in other way.

oney:
Of course. if the rollback are automatic, it will be awesome! I don't think you must merge this PR, just want to give some help.
I will use these code in my app now(I only use IMMEDIATE install mode) and I will look forward to automatic rollback.

geof90:
Hi @oney,

I just merged in a PR https://github.com/Microsoft/react-native-code-push/pull/104 (which is basically very much inspired by your code). It should address the issue that you are facing, plus also generalize it to work for the ON_NEXT_RESTART install mode. Do leave some comments there or file an issue otherwise. Thanks once again for your ideas and contributions!

Will close this PR.

oney:
You guys do a great job! Thanks! :metal:

PR #99, opened 2015-12-08
Adding logging to UNKNOWN_ERROR sync status

lostintangent:
If a developer calls `done()` on the `Promise` that is returned from `sync`, any errors that occur as part of the process will result in a red-box during development. Additionally, if they add a `catch` block to the `Promise`, they can be notified and inspect the actual error that occurred.

However, if a developer simply calls `sync()`, then any exceptions (e.g. non-semver compliant app version in `Info.plist`, missing deployment key) are swallowed, and the only diagnostics information they have is the "Unknown error" message that is logged by `sync`. 

This change simply logs the error message to the console, along with rejecting the `Promise` and notifying the `syncStatusCallback` about the error. This way, the simple consumption scenario is a bit easier to diagnose when an error occurs.

Additionally, any errors that were occurring as part of the `download` or `install` portion of the `sync` method were also being swallowed, so this change does the same `Promise` rejection in those situations as was already being done for the `checkForUpdate` portion.

### DIFF
--- a/CodePush.js
+++ b/CodePush.js
@@ -294,6 +294,7 @@ function sync(options = {}, syncStatusChangeCallback, downloadProgressCallback) 
       })
       .catch((error) => {
         syncStatusChangeCallback(CodePush.SyncStatus.UNKNOWN_ERROR);
+        log(error.message);
         reject(error);
       })
       .done();
--- a/CodePush.js
+++ b/CodePush.js
@@ -219,6 +219,12 @@ function sync(options = {}, syncStatusChangeCallback, downloadProgressCallback) 
       };
   
   return new Promise((resolve, reject) => {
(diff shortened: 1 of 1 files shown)


CodePush.js
```diff
@@ -294,6 +294,7 @@ function sync(options = {}, syncStatusChangeCallback, downloadProgressCallback)
       })
       .catch((error) => {
         syncStatusChangeCallback(CodePush.SyncStatus.UNKNOWN_ERROR);
+        log(error.message);
```
shishirx34: nit: don't know if it matters, but may be log error before calling syncStatusChangeCallback, if for some reason callback itself crashes we won't have this logged?

shishirx34:
LGTM!

geof90:
LGTM!

PR #100, opened 2015-12-09
Use Application Support Directory

geof90:
This PR:

1. Uses the Libraries/Application Support directory instead of the Documents directory to store files, so as to better conform to [Apple's guidelines](https://developer.apple.com/library/ios/documentation/FileManagement/Conceptual/FileSystemProgrammingGuide/FileSystemOverview/FileSystemOverview.html#//apple_ref/doc/uid/TP40010672-CH2-SW1)

2. Bumps the modified date in the file metadata of downloaded bundles to the current date at which it was installed. After supporting asset updates, the files in the zip would keep their file metadata, which is not what we want - we want to use the installed date in comparison with the binary's date to determine which version is newer.

### DIFF
--- a/CodePush.h
+++ b/CodePush.h
@@ -20,7 +20,7 @@
 + (NSURL *)bundleURLForResource:(NSString *)resourceName
                   withExtension:(NSString *)resourceExtension;
 
-+ (NSString *)getDocumentsDirectory;
++ (NSString *)getApplicationSupportDirectory;
 
 @end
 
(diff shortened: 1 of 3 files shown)


lostintangent:
LGTM! Is it safe to assume that this change from the documents to library directory shouldn't break any existing users? I suppose it's possible you could get an update that you already have since the logic would now look in a different location for what you're current package is.

geof90:
@lostintangent 

This change will just mean that all of the updates in the documents directory will now be ignored (treated as if they were never installed), since it no longer goes to the Documents directory to find the status file.

lostintangent:
Makes sense. We should probably mention in the release notes that you should uninstall the app on the simulator or devices so as to clean out the old documents directory.

PR #101, opened 2015-12-09
Version bump

lostintangent:
Simply bumping the version so that we can get out a couple key bug fixes to NPM.


qingfeng:
:beers:

Issue #87, opened 2015-12-04
failedApply -> failedInstall


dtivel:
https://github.com/Microsoft/react-native-code-push/blob/f0b334fb7b7999cbe230021439f3c9d39611a5a7/CodePush.m#L359

lostintangent:
Fixed with #102

PR #102, opened 2015-12-09
Fixing comment to reflect API name change

lostintangent:
We renamed `failedApply` to `failedInstall` but forgot to update this comment to reflect that change. This fix addresses bug #87.


dtivel:
Looks good.

PR #104, opened 2015-12-10
Remove rollback timeout

geof90:
Breaking Changes:
1. The notion of a rollback timeout is now removed. In the event that an app crashes after an update, CodePush will automatically rollback and load the previous version when the user starts the app again.
2. If you are using the advanced API, you MUST call notifyApplicationReady() somewhere in your bundle. This is to notify CodePush that the update is okay. Otherwise, CodePush will treat the update as a faulty one and rollback when the app next restarts. If you are using the Sync API, notifyApplicationReady() is called inside sync(), so you do not need to change anything.

Bug fix:
1. On Android, the updates were not being loaded when the app is in developer mode, even when the dev server is not running. This was caused by React Native caching the bundle downloaded from the dev server, and since dev server bundles always take precedence over any other locally stored bundles, CodePush updates will never be run even after they are installed. This fixes that issue by simply deleting the cached file when an update is applied.

### DIFF
--- a/CodePush.m
+++ b/CodePush.m
@@ -119,7 +119,7 @@ static NSString *const PendingUpdateWasInitializedKey = @"wasInitialized";
         _isFirstRunAfterUpdate = YES;
         BOOL wasInitialized = [pendingUpdate[PendingUpdateWasInitializedKey] boolValue];
         if (wasInitialized) {
-            // Pending update was initialized, but notifiyApplicationReady was not called.
+            // Pending update was initialized, but notifyApplicationReady was not called.
             // Therefore, deduce that it is a broken update and rollback.
             [self rollbackPackage];
         } else {
--- a/CodePush.m
+++ b/CodePush.m
@@ -106,8 +106,9 @@ static NSString *const PendingUpdateWasInitializedKey = @"wasInitialized";
 }
 
 /*
- * This method starts the rollback protection timer
- * and is used when a new update is initialized.
+ * This method is used when the app is started to either
+ * initialize a pending update or rollback a faulty update
+ * to the previous version.
  */
 - (void)initializeUpdateAfterRestart
 {
@@ -122,7 +123,7 @@ static NSString *const PendingUpdateWasInitializedKey = @"wasInitialized";
             // Therefore, deduce that it is a broken update and rollback.
             [self rollbackPackage];
         } else {
-            // Mark that we tried to initiazlie the new update, so that if it crashes,
+            // Mark that we tried to initialize the new update, so that if it crashes,
             // we will know that we need to rollback when the app next starts.
             [self savePendingUpdate:pendingUpdate[PendingUpdateHashKey]
                      wasInitialized:YES];
@@ -174,10 +175,9 @@ static NSString *const PendingUpdateWasInitializedKey = @"wasInitialized";
     // Write the current package's hash to the "failed list"
     [self saveFailedUpdate:packageHash];
     
-    // Do the actual rollback and then
-    // refresh the app with the previous package
+    // Rollback to the previous version and de-register the new update
+    [CodePushPackage rollbackPackage];
     [self removePendingUpdate];
-    [self loadBundle];
 }
 
 /*
--- a/CodePush.m
+++ b/CodePush.m
@@ -22,7 +22,7 @@ static NSString *const PendingUpdateKey = @"CODE_PUSH_PENDING_UPDATE";
 // These keys are already "namespaced" by the PendingUpdateKey, so
 // their values don't need to be obfuscated to prevent collision with app data
 static NSString *const PendingUpdateHashKey = @"hash";
-static NSString *const PendingUpdateWasInitializedKey = @"wasInitialized";
+static NSString *const PendingUpdateIsLoadingKey = @"isLoading";
 
 @synthesize bridge = _bridge;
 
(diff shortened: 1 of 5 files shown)


android/app/src/main/java/com/microsoft/codepush/react/CodePush.java
```diff
@@ -204,52 +187,60 @@ private void saveFailedUpdate(String packageHash) {
...
+        if (pendingUpdateString == null) {
+            return null;
+        }
+
+        try {
+            JSONObject pendingUpdate = new JSONObject(pendingUpdateString);
+            return pendingUpdate;
+        } catch (JSONException e) {
+            // Should not happen.
+            throw new CodePushUnknownException("Unable to parse pending update metadata " +
```
dlebu: How about handling the exception silently and not re-throwing here?

dlebu:
Do we have any automated tests for this change?

geof90:
@dlebu Not yet. We are behind on implementing tests for a few features, e.g the sync implementation currently has no automated test coverage. There is a task in the backlog assigned to me for writing all automated tests. We just want to quickly get this out there because right now a number of customers have given us feedback that the present way that rollbacks work is not working for them (app crashes before notifyApplicationReady is called that terminates the app preventing the rollback timer from ever kicking in).

CodePush.m
```diff
@@ -130,16 +117,17 @@ - (void)initializeUpdateAfterRestart
...
     if (pendingUpdate) {
         _isFirstRunAfterUpdate = YES;
-        int rollbackTimeout = [pendingUpdate[PendingUpdateRollbackTimeoutKey] intValue];
-        if (0 != rollbackTimeout) {
-            dispatch_async(dispatch_get_main_queue(), ^{
-                [self startRollbackTimer:rollbackTimeout];
-            });
+        BOOL wasInitialized = [pendingUpdate[PendingUpdateWasInitializedKey] boolValue];
+        if (wasInitialized) {
+            // Pending update was initialized, but notifiyApplicationReady was not called.
```
dtivel: notifyApplicationReady

CodePush.m
```diff
@@ -216,38 +203,35 @@ - (void)saveFailedUpdate:(NSString *)packageHash
 }
 
 /*
+ * This method is called in notifyApplicationReady to register the fact that
+ * the pending update succeeded and therefore can be removed.
```
dtivel: I don't think you should include caller information in comments.  Just state what this method does.

CodePush.m
```diff
@@ -130,16 +117,17 @@ - (void)initializeUpdateAfterRestart
     
     if (pendingUpdate) {
         _isFirstRunAfterUpdate = YES;
-        int rollbackTimeout = [pendingUpdate[PendingUpdateRollbackTimeoutKey] intValue];
-        if (0 != rollbackTimeout) {
-            dispatch_async(dispatch_get_main_queue(), ^{
-                [self startRollbackTimer:rollbackTimeout];
-            });
+        BOOL wasInitialized = [pendingUpdate[PendingUpdateWasInitializedKey] boolValue];
+        if (wasInitialized) {
```
dtivel: Not crazy about this name.  I think a better name would be "isUpdateLoading."  From the point that the bridge reloads to now, that is what's happening.  If the value is true here, then the update loaded successfully.

README.md
```diff
@@ -248,9 +248,9 @@ When you require `react-native-code-push`, the module object provides the follow
 
 * [getCurrentPackage](#codepushgetcurrentpackage): Retrieves the metadata about the currently installed update (e.g. description, installation time, size).
 
-* [notifyApplicationReady](#codepushnotifyapplicationready): Notifies the CodePush runtime that an installed update is considered successful. This is an optional API, but is useful when you want to expicitly enable "rollback protection" in the event that an exception occurs in code that you've deployed to production. 
+* [notifyApplicationReady](#codepushnotifyapplicationready): Notifies the CodePush runtime that an installed update is considered successful. If you are manually checking for and install updates (i.e. not using the [sync](#codepushsync) method to handle it all for you), then this method **MUST** be called, otherwise CodePush will regard the update as a failed one and rollback to the previous version when the app next restarts.
```
dtivel: "...and installing updates..."

"...be called; otherwise, CodePush will treat the update as failed and..."

README.md
```diff
@@ -248,9 +248,9 @@ When you require `react-native-code-push`, the module object provides the follow
...
 
-* [restartApp](#codepushrestartapp): Immediately restarts the app. If there is an update pending, it will be immediately displayed to the end-user, and the rollback timer (if specified when installing the update) will begin. Otherwise, calling this method simply has the same behavior as the end-user killing and restarting the process.
+* [restartApp](#codepushrestartapp): Immediately restarts the app. If there is an update pending, it will be immediately displayed to the end-user. Otherwise, calling this method simply has the same behavior as the end-user killing and restarting the process.
```
dtivel: "end user" (no hyphen)

README.md
```diff
@@ -310,9 +310,7 @@ codePush.getCurrentPackage()
...
-
-If the `rollbackTimeout` parameter was not specified, the CodePush runtime will not enforce any automatic rollback behavior, and therefore, calling this function is not required and will result in a no-op.
+Notifies the CodePush runtime that a freshly installed update should be considered successful, and therefore, an automatic client-side rollback isn't necessary. It is mandatory to call this function somewhere in the code of the updated bundle. Otherwise, when the app next restarts, the CodePush runtime will assume that the installed update has failed and roll back to the previous version. This behavior exists to help ensure that your end-users aren't blocked by a broken update.
```
dtivel: "end users" (not hyphenated)

README.md
```diff
@@ -459,11 +455,7 @@ Contains details about an update that has been downloaded locally or already ins
...
-<br /><br />
-*Note: The "rollback timer" doesn't start until the update has actually become active. If the `installMode` is `IMMEDIATE`, then the rollback timer will also start immediately. However, if the `installMode` is `UPDATE_ON_RESTART` or `UPDATE_ON_RESUME`, then the rollback timer will start the next time the app starts or resumes, not at the point that you called `install`.*
+- __install(installMode: CodePush.InstallMode = CodePush.InstallMode.ON_NEXT_RESTART): Promise&lt;void&gt;__: Installs the update by saving it to the location on disk where the runtime expects to find the latest version of the app. The `installMode` parameter controls when the changes are actually presented to the end-user. The default value is to wait until the next app restart to display the changes, but you can refer to the [`InstallMode`](#installmode) enum reference for a description of the available options and what they do.
```
dtivel: "end user" (no hyphen)

dtivel:
Looks good.

PR #105, opened 2015-12-11
Clear dev bundle cache

geof90:
The dev bundle cache should be cleared during init, not inside loadBundle(), otherwise, InstallMode.ON_NEXT_RESTART would not work.

### DIFF
--- a/android/app/src/main/java/com/microsoft/codepush/react/CodePush.java
+++ b/android/app/src/main/java/com/microsoft/codepush/react/CodePush.java
@@ -53,6 +53,7 @@ public class CodePush {
     private final String CODE_PUSH_TAG = "CodePush";
     private final String DOWNLOAD_PROGRESS_EVENT_NAME = "CodePushDownloadProgress";
     private final String RESOURCES_BUNDLE = "resources.arsc";
+    private final String REACT_DEV_BUNDLE_CACHE_FILE_NAME = "ReactNativeDevBundle.js";
 
     private CodePushPackage codePushPackage;
     private CodePushReactPackage codePushReactPackage;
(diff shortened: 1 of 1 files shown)


android/app/src/main/java/com/microsoft/codepush/react/CodePush.java
```diff
@@ -53,6 +53,7 @@
     private final String CODE_PUSH_TAG = "CodePush";
     private final String DOWNLOAD_PROGRESS_EVENT_NAME = "CodePushDownloadProgress";
     private final String RESOURCES_BUNDLE = "resources.arsc";
+    private final String REACT_DEV_BUNDLE_CACHE_FILE_NAME = "ReactNativeDevBundle.js";
```
lostintangent: We should add a comment here that mentions that this file needs to be kept in sync with the equivalent constant in the React Native repo, and even point to the file/line that its located at.

shishirx34:
LGTM!

dtivel:
Looks good.

android/app/src/main/java/com/microsoft/codepush/react/CodePush.java
```diff
@@ -236,6 +238,8 @@ private void initializeUpdateAfterRestart() {
                     // Therefore, deduce that it is a broken update and rollback.
                     rollbackPackage();
                 } else {
+                    // Clear the React dev bundle cache so that new updates can be loaded.
+                    clearReactDevBundleCache();
```
lostintangent: It might not be entirely necessary, but I'd still love to know whether we can do this only during debug builds. I'd like to be hyper-sensitive with any work we do as part of the app load process, to keep CodePush as impactless as possible.

Could you quickly try seeing whether `BuildConfig.DEBUG` can be used here? I'm assuming that it would be true/false depending on what the "parent" project configuration is, and therefore, when the dev support is enabled, we would clear the bundle, but when it's not, we wouldn't.

geof90: @lostintangent BuildConfig is a generated class that is packaged to the parent project, e.g. com.mycompany.myapp.BuildConfig. We have no idea what their namespace is from our plugin package. One possible way is to get developers to pass in that value to CodePush via the constructor so that we can use it. That will involve a small but nonetheless breaking change to the current setup instructions. What do you think?

geof90: offline discussion - can't do this due to bug with android https://code.google.com/p/android/issues/detail?id=52962

PR #108, opened 2015-12-11
Fix date comparison bug

geof90:
This fixes an issue with the native bundle somehow taking precedence over a newly CodePush installed bundle, caused by http://stackoverflow.com/questions/27675454/zipentry-gettime-unpredictable-results. It turns out that on certain devices and time zones, the binary date returned could be newer than a CodePush installed bundle. The fix writes the the binary's resources modified time in each installed pacakge metadata, and use that for comparison instead.

iOS resources are not packaged as a zip, so this change does not impact iOS.

Also, added logging when the binary version is used over a CodePush installed package, and whenever a rollback is applied.

### DIFF
--- a/android/app/src/main/java/com/microsoft/codepush/react/CodePush.java
+++ b/android/app/src/main/java/com/microsoft/codepush/react/CodePush.java
@@ -1,6 +1,6 @@
 package com.microsoft.codepush.react;
 
-import com.facebook.react.*;
+import com.facebook.react.ReactPackage;
 import com.facebook.react.bridge.JavaScriptModule;
 import com.facebook.react.bridge.LifecycleEventListener;
 import com.facebook.react.bridge.NativeModule;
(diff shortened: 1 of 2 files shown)


android/app/src/main/java/com/microsoft/codepush/react/CodePush.java
```diff
@@ -119,15 +128,22 @@ public String getBundleUrl(String assetsBundleFileName) {
...
-            File packageFile = new File(packageFilePath);
-            if (packageFile.lastModified() < binaryResourcesModifiedTime) {
+            ReadableMap packageMetadata = codePushPackage.getCurrentPackage();
+            // May throw NumberFormatException.
+            Long binaryModifiedDateDuringPackageInstall = Long.parseLong(CodePushUtils.tryGetString(packageMetadata, BINARY_MODIFIED_TIME_KEY));
+            if (binaryModifiedDateDuringPackageInstall == binaryResourcesModifiedTime) {
+                return packageFilePath;
+            } else {
                 // The binary version is newer.
+                CodePushUtils.log("Found a package installed via CodePush that was installed " +
```
lostintangent: What if we simply said "Loading JS bundle from binary", and then add another log to the CodePush path that said "Loading JS bundle from latest CodePush update"?

lostintangent:
LGTM!

Issue #110, opened 2015-12-15
Calling sync twice in a row


JAStanton:
if I call:
```js
CodePush.sync({ installMode: installMode: CodePush.InstallMode.ON_NEXT_RESUME })
```
and it finds an update and downloads it but before I restart or resume the app the app I call:
```js
CodePush.sync({ installMode: installMode: CodePush.InstallMode.IMMEDIATE })
```

Would you expect it to immediately restart or no-op?
Because right now nothing happens, and I expected it to restart immediately.

lostintangent:
The expected behavior would be to no-op, since the "pending" update from the first call to sync is already downloaded, and therefore, the second call to sync would resolve its promise as no update available. Sync doesn't currently check for pending updates, which is why this behavior exists. That said, we've discussed modifying this behavior, so just because it's the currently expected behavior, doesn't mean it's the right behavior :)

If you're interested in "artificially" restarting the app in order to apply a previously downloaded update, you could call the restartApp module-level method as a workaround. If however, you want to restart the app only if there is a pending update, there isn't currently a way to do that.

We've discussed doing three different things that might help your scenario, and I'd be interested if any of them sound sufficient:

1. Adding an isPending property to the LocalPackage object that is returned when calling getCurrentPackage. This could be used to query the current package and if the latest update is still pending, restart the app. This might make sense to do in general, but it's probably a roundabout way of solving this scenario specifically.

2. Add a parameter to restartApp that specifies you only want to restart if there is a pending update. Semantically, I feel like this solution is pretty decent if you want to call sync at one point, and then "force" a pending update in another. 

3. Modify sync to look for pending updates, and effectively overwrite the previously specified install mode. The only reason I'm not in love with this one is that the only thing I can imagine people wanting to use this solution for is to apply a pending update sooner, and I'm not sure if this is the most intuitive way to solve it. Maybe it is, I'm not sure.

What are your thoughts? Doing the first or second items would be super straight forward and not breaking changes, so we could address them pretty quickly.

JAStanton:
Yeah the first two solutions would work best actually. I think `isPending` on the LocalPackage would probably make the most sense. Thats the first place I looked :)

lostintangent:
OK cool. I'll send a PR out for this tonight. Stay tuned and thanks for the feedback!

JAStanton:
Thanks!

nevir:
[...]

Issue #107, opened 2015-12-11
[Overrides] Varargs doesn't agree for overridden method


ben-manes:
[Google ErrorProne](http://errorprone.info/) static analysis fails due to the [Override pattern](http://errorprone.info/bugpattern/Overrides) (via the [gradle-plugin](https://github.com/tbroyer/gradle-errorprone-plugin)).

```
/Users/ben/projects/loaddocs/loaddocs-react-native/node_modules/react-native-code-push/android/app/src/main/java/com/microsoft/codepush/react/CodePush.java:271: error: [Overrides] Varargs doesn't agree for overridden method
                protected Void doInBackground(Object[] params) {
                               ^
    (see http://errorprone.info/bugpattern/Overrides)
  Did you mean 'protected Void doInBackground(Object... params) {'?
/Users/ben/projects/loaddocs/loaddocs-react-native/node_modules/react-native-code-push/android/app/src/main/java/com/microsoft/codepush/react/CodePush.java:322: error: [Overrides] Varargs doesn't agree for overridden method
                protected Void doInBackground(Object[] params) {
                               ^
    (see http://errorprone.info/bugpattern/Overrides)
  Did you mean 'protected Void doInBackground(Object... params) {'?
/Users/ben/projects/loaddocs/loaddocs-react-native/node_modules/react-native-code-push/android/app/src/main/java/com/microsoft/codepush/react/CodePush.java:361: error: [Overrides] Varargs doesn't agree for overridden method
                protected Void doInBackground(Object[] params) {
                               ^
    (see http://errorprone.info/bugpattern/Overrides)
  Did you mean 'protected Void doInBackground(Object... params) {'?
Note: /Users/ben/projects/loaddocs/loaddocs-react-native/node_modules/react-native-code-push/android/app/src/main/java/com/microsoft/codepush/react/CodePush.java uses unchecked or unsafe operations.
Note: Recompile with -Xlint:unchecked for details.
3 errors
:react-native-code-push:compileReleaseJavaWithJavac FAILED
```

ben-manes:
fyi, the workaround for now is to disable the bug pattern.

```gradle
allprojects {
  gradle.projectsEvaluated {
    tasks.withType(JavaCompile) {
      options.compilerArgs << "-Xep:Overrides:OFF"
    }
  }
}
```

lostintangent:
Great catch @ben-manes! This should be a pretty straight forward and we'll try to get to it early next week. Thanks again for reporting it.

[...]

PR #113, opened 2015-12-17
isPending property on package objects

lostintangent:
This PR primarily introduces a new `isPending` property onto the package objects, to allow apps to determine whether the current package (obtained by calling `getCurrentPackage`) represents an installed but not yet applied update. This is a feature we knew we needed to add, but bug #110 help re-prioritize it higher. Moving forward, I would like to improve the way we handle the `isFirstRun` and `failedInstall` properties, so that they don't require extra roundtrips to the native side. I implemented `isPending` this way, so as to not create another roundtrip, but I didn't want to modify more code right now than was necessary.

Additionally, while addressing this item, I made a few additional "cleanup" changes:

1. The `_timer` ivar in the iOS implementation wasn't necessary anymore now that we removed the `rollbackTimeout` parameter from `sync` and `LocalPackage.install`.

2. I re-ordered the methods declared in the `CodePush` class in the Android implementation for better readability. This is the cause of the majority of the changes in `CodePush.java`.

3. I updated the `package-mixins.js` file to use a few ES6 features (e.g. property name shorthands, arrow functions, ES6 module imports, const vs. var), for better readability/semantic value.

4. I addressed bug #107, since it was super simple.

### DIFF
--- a/CodePush.m
+++ b/CodePush.m
@@ -15,7 +15,7 @@ RCT_EXPORT_MODULE()
 
 static BOOL usingTestFolder = NO;
 
-// These keys represent the names we use to store data in NSUserDdefaults
+// These keys represent the names we use to store data in NSUserDefaults
 static NSString *const FailedUpdatesKey = @"CODE_PUSH_FAILED_UPDATES";
 static NSString *const PendingUpdateKey = @"CODE_PUSH_PENDING_UPDATE";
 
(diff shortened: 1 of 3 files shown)


geof90:
LGTM! There are a few other files where we could add the ES6 features, i had changed them locally and will put them out in my next PR.

lostintangent:
Yeah, I decided to keep my ES6 changes localized to just the package mixins file for now, since I was already working in there. I didn't realize you had made those ES6 changes already, so I apologize for duplicating efforts!

PR #116, opened 2015-12-18
Tests + Bugfix

geof90:
This PR adds more automated tests using XCTest and RCTTestModule/RCTTestRunner.
To run the tests, simply go to the XCode testing tab and click the gray triangle. Unfortunately, RCTTestModule/RCTTestRunner is not yet released on Android, however the tests can still be run by pointing the file to the desired test case JavaScript file in MainActivity, granted its a more tedious process.

tests

Also found and fixed a bug when installing under InstallMode.IMMEDIATE, where the update would not be applied if notifyApplicationReady() was not called before the component was mounted, due to the fact that `[CodePush loadBundle]` was called twice, once from pacakage-mixins.js, and once from `CodePush.installUpdate`, resulting in the rollback process being erroneously triggered.

Also added logging on iOS to display the selected bundle locations when the app starts and logging before rollbacks are applied.


qingfeng:
COOL! :beers:

CodePushPackage.m
```diff
@@ -481,4 +486,31 @@ + (void)rollbackPackage
     [self updateCurrentPackageInfo:info error:&error];
 }
 
++ (void)downloadAndReplaceCurrentBundle:(NSString *)remoteBundleUrl
+{
+    NSURL *urlRequest = [NSURL URLWithString:remoteBundleUrl];
+    NSError *error = nil;
+    NSString *downloadedBundle = [NSString stringWithContentsOfURL:urlRequest
+                                                encoding:NSUTF8StringEncoding
+                                                   error:&error];
```
shishirx34: align at ':' ?

Examples/CodePushDemoApp/CodePushDemoAppTests/CheckForUpdateTests/testcases/NewUpdateTest.js
```diff
@@ -0,0 +1,58 @@
...
+  downloadURL: "http://www.windowsazure.com/blobs/awperoiuqpweru",
+  description: "Angry flappy birds",
+  appVersion: "1.5.0",
+  label: "2.4.0",
+  isMandatory: false,
+  isAvailable: true,
+  updateAppVersion: false,
+  packageHash: "hash123",
+  packageSize: 1024
+};
```
shishirx34: not super important, but may be consider moving the above two(and other related redundant resources in these tests) packages to a resources.json(or something like that) file and read it once before running tests and assign them in here, so that in future if you have a change to the structure of acquisition package you don't have to make the change at multiple places. 

Maintenance of tests becomes a pain after a certain point, its best to keep them as low maintenance as possible.

Again, not super important, if you think it is worth it.

shishirx34:
LGTM! Great work Geoff! :+1: :shipit:

CodePush.js
```diff
@@ -130,11 +130,8 @@ function getCurrentPackage() {
 }
 
 function getSDK(config) {
```
lostintangent: With this simplification, is this function needed anymore? Or could we inline it within `checkForUpdate`?

Examples/CodePushDemoApp/CodePushDemoAppTests/CheckForUpdateTests/testcases/NewUpdateTest.js
lostintangent: I was thinking the same thing :)

Examples/CodePushDemoApp/CodePushDemoAppTests/InstallUpdateTests/testcases/IsPendingTest.js
```diff
@@ -0,0 +1,46 @@
+"use strict";
```
lostintangent: Should we also add a test for checking `isPending` after a call to `getCurrentPackage` since that is actually populated differently and is the most common case?

CodePush.m
```diff
@@ -351,9 +386,7 @@ - (void)savePendingUpdate:(NSString *)packageHash
             [self savePendingUpdate:updatePackage[PackageHashKey]
                           isLoading:NO];
             
-            if (installMode == CodePushInstallModeImmediate) {
-                [self loadBundle];
-            } else if (installMode == CodePushInstallModeOnNextResume) {
+            if (installMode == CodePushInstallModeOnNextResume) {
```
lostintangent: Was the fix for the immediate-applied update big simply removing this call to `loadBundle` and allowing the script-side to trigger the restart? Would this bug affect the Android plugin?

lostintangent:
LGTM!

CodePush.m
geof90: Yes it would, I am currently working on android changes now, will incorporate all of yours and shishir's feedback soon.

Examples/CodePushDemoApp/CodePushDemoAppTests/InstallUpdateTests/testcases/IsPendingTest.js
geof90: Added!

PR #117, opened 2015-12-19
Remove custom dialog

geof90:
This PR upgrades to RN 0.17.0, and in so doing, removes our custom dialog module implementation since the [OOTB RN Alert module is now cross platform](http://facebook.github.io/react-native/docs/alert.html#content).

Fixed a minor bug with tests, the remotePackage for InstallUpdateTests was exported under the "default" property and not directly as the exported object itself.


Examples/CodePushDemoApp/android/app/proguard-rules.pro
```diff
@@ -15,3 +15,46 @@
 #-keepclassmembers class fqcn.of.javascript.interface.for.webview {
 #   public *;
 #}
+
```
geof90: Just a note that this file change is auto-generated from running the `react-native upgrade` command.

shishirx34:
LGTM!

lostintangent:
LGTM, but I'd like to address the few bugs we have logged before making a change in RN version dependency.

lostintangent:
Since this is simply a code-cleanup PR, which doesn't add any new features, but does represent a breaking change, I think we should wait to merge this along with adding assets support for Android, since that will require a new version of React Native as well. If your [PR](https://github.com/facebook/react-native/pull/4527) gets merged in time for 0.18.0, then I think it would be great to release a new minor release of the plugin that depends on that RN version, and adds in Android asset support and removes our dialog implementation at the same time.

geof90:
Closed because of issues with the RN Alert being unable to support our scenario, such as it being dismissable when the user touches the outside of the dialog.

PR #122, opened 2015-12-21
Adding release detail

lostintangent:
We weren't recommending specifying `--dev false` when calling `react-native bundle`, and the assets docs weren't clear enough that the "release" directory needs to be created manually before calling `react-native bundle`.


geof90:
LGTM!

Issue #114, opened 2015-12-17
Old pendingUpdate with new updated binary


JAStanton:
This might be an edge case but It's worth bringing up. What would happen if I download an app from the app store lets call build A. then I open build A and sometime later I have an update from code-push. the user syncs w/ an install mode of next time the app is opened. I close the app. some time passes before I re-open the app, a new build from the app store is automatically synced from the app store to my phone from apple, this is build B. if I open that app again w/ build b installed. Will that pending update for build A apply to my shiny new build B?

The reason why I ask is because I ran into this while trying to test code-push w/ Fabric / Crashlytics.
I pushed Build A, it opened, downloaded a pending update. I closed the app, without deleting the app I pushed a new build to Fabric, Build B, opened the build and the pending update overwrote my build A.

lostintangent:
I would expect Build B's JS bundle to take effect since the [CodePush bundleURL] method would give it precedence over any previously downloaded CodePush updates. That said, as you mentioned, we aren't clearing out the previous pending update currently, and so there may be some unexpected behavior in this scenario specifically.

This is a good bug report, and it would be pretty straight forward to fix, since we already have logic to detect an App Store update having occurred. Thanks for reporting this!

geof90:
I agree that we should definitely clear older updates and I am working towards a fix for that, but can I confirm that either of you were seeing a bundle downloaded in an older store version being loaded in a new store version? Because that should definitely not happen. Not clearing pendingUpdate information should not lead to this happening either.

lostintangent:
Nope, I wasn't able to repro this issue. I just agreed that we should clean this old state out :)

geof90:
Should we do a check in [CodePush bundleURL] on the saved package appVersion instead of just the date, just to further iron clad ourselves from this issue?

lostintangent:
That sounds good to me. That way the date check helps in scenarios where the version doesn't have to be bumped (e.g. development, Fabric), and the appVersion would be primarily for production scenarios

Issue #123, opened 2015-12-21
Fatal exception - Error reading binary modified date from package metadata


ajwhite:
I've upgraded CodePush from `1.3.2-beta` to `1.5.2-beta` along with ReactNative from `0.15.0` to `0.17.0` and seem to hit a problem when starting up the app on Android:

In my `onCreate` I have the following calls on the builder:

```java
.setJSMainModuleName("index.android")
.setJSBundleFile(codePush.getBundleUrl("index.android.bundle"))
.addPackage(codePush.getReactPackage())
```

```
E/AndroidRuntime(10892): FATAL EXCEPTION: main
E/AndroidRuntime(10892): Process: com.robinpowered.rooms, PID: 10892
E/AndroidRuntime(10892): java.lang.RuntimeException: Unable to start activity ComponentInfo{com.robinpowered.rooms/com.robinrooms.MainActivity}: com.microsoft.codepush.react.CodePushUnknownException: Error in reading binary modified date from package metadata
E/AndroidRuntime(10892): 	at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:2325)
E/AndroidRuntime(10892): 	at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:2387)
E/AndroidRuntime(10892): 	at android.app.ActivityThread.access$800(ActivityThread.java:151)
E/AndroidRuntime(10892): 	at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1303)
E/AndroidRuntime(10892): 	at android.os.Handler.dispatchMessage(Handler.java:102)
E/AndroidRuntime(10892): 	at android.os.Looper.loop(Looper.java:135)
E/AndroidRuntime(10892): 	at android.app.ActivityThread.main(ActivityThread.java:5254)
E/AndroidRuntime(10892): 	at java.lang.reflect.Method.invoke(Native Method)
E/AndroidRuntime(10892): 	at java.lang.reflect.Method.invoke(Method.java:372)
E/AndroidRuntime(10892): 	at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:903)
E/AndroidRuntime(10892): 	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:698)
E/AndroidRuntime(10892): Caused by: com.microsoft.codepush.react.CodePushUnknownException: Error in reading binary modified date from package metadata
[...]

PR #125, opened 2015-12-22
Clear updates on new binary

geof90:
Fixes https://github.com/Microsoft/react-native-code-push/issues/114 and https://github.com/Microsoft/react-native-code-push/issues/123 (they are slightly different)

### DIFF
--- a/CodePush.m
+++ b/CodePush.m
@@ -109,6 +109,7 @@ static NSString *const PackageIsPendingKey = @"isPending";
 {
     [CodePushPackage clearUpdates];
     [self removePendingUpdate];
+    [self removeFailedUpdates];
 }
 
 
@@ -267,6 +268,17 @@ static NSString *const PackageIsPendingKey = @"isPending";
     
     [failedUpdates addObject:packageHash];
     [preferences setObject:failedUpdates forKey:FailedUpdatesKey];
+    [preferences synchronize];
+}
+
+/*
+ * This method is used to clear away failed updates in the event that
+ * a new app store binary is installed.
+ */
++ (void)removeFailedUpdates
+{
+    NSUserDefaults *preferences = [NSUserDefaults standardUserDefaults];
+    [preferences removeObjectForKey:FailedUpdatesKey];
     [preferences synchronize];
 }
 
--- a/CodePush.m
+++ b/CodePush.m
@@ -62,7 +62,13 @@ static NSString *const PackageIsPendingKey = @"isPending";
     NSDate *binaryDate = [binaryFileAttributes objectForKey:NSFileModificationDate];
     NSDate *packageDate = [appFileAttribs objectForKey:NSFileModificationDate];
     NSString *binaryAppVersion = [[CodePushConfig current] appVersion];
-    NSString *packageAppVersion = [appFileAttribs objectForKey:@"appVersion"];
+    NSDictionary *currentPackageMetadata = [CodePushPackage getCurrentPackage:&error];
+    if (error || !currentPackageMetadata) {
+        NSLog(logMessageFormat, binaryJsBundleUrl);
+        return binaryJsBundleUrl;
+    }
+    
+    NSString *packageAppVersion = [currentPackageMetadata objectForKey:@"appVersion"];
     
     if ([binaryDate compare:packageDate] == NSOrderedAscending && [binaryAppVersion isEqualToString:packageAppVersion]) {
         // Return package file because it is newer than the app store binary's JS bundle
--- a/CodePush.m
+++ b/CodePush.m
@@ -61,13 +61,16 @@ static NSString *const PackageIsPendingKey = @"isPending";
     NSDictionary *appFileAttribs = [[NSFileManager defaultManager] attributesOfItemAtPath:packageFile error:nil];
     NSDate *binaryDate = [binaryFileAttributes objectForKey:NSFileModificationDate];
     NSDate *packageDate = [appFileAttribs objectForKey:NSFileModificationDate];
+    NSString *binaryAppVersion = [[CodePushConfig current] appVersion];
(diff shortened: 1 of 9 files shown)


shishirx34:
LGTM

CodePush.m
```diff
@@ -100,16 +103,12 @@ + (void)setUsingTestConfiguration:(BOOL)shouldUseTestConfiguration
 }
 
 /*
- * This is used to clean up all test updates. It can only be used
- * when the testConfigurationFlag is set to YES, otherwise it will
- * simply no-op.
+ * WARNING: This cleans up all downloaded and pending updates.
  */
-+ (void)clearTestUpdates
++ (void)clearUpdates
```
lostintangent: Should we also clear out failed updates? The Cordova plugin does this and I think it makes sense since those updates may work not that there is a new binary.

android/app/src/main/java/com/microsoft/codepush/react/CodePush.java
```diff
@@ -130,13 +130,21 @@ public String getBundleUrl(String assetsBundleFileName) {
...
+
+            String pacakgeAppVersion = CodePushUtils.tryGetString(packageMetadata, "appVersion");
+            if (binaryModifiedDateDuringPackageInstall != null &&
+                    binaryModifiedDateDuringPackageInstall == binaryResourcesModifiedTime &&
+                    this.appVersion.equals(pacakgeAppVersion)) {
                 CodePushUtils.logBundleUrl(packageFilePath);
                 return packageFilePath;
             } else {
                 // The binary version is newer.
+
```
lostintangent: Shouldn't you be calling clearUpdates here?

geof90: Yep sorry, was going to add that in. Am multitasking between 3 PRs

CodePush.m
geof90: Yep, that makes sense.

geof90: Done

CodePush.m
```diff
@@ -100,16 +109,13 @@ + (void)setUsingTestConfiguration:(BOOL)shouldUseTestConfiguration
...
 {
-    if ([CodePush isUsingTestConfiguration]) {
-        [CodePushPackage clearTestUpdates];
-        [self removePendingUpdate];
-    }
+    [CodePushPackage clearUpdates];
```
lostintangent: Should we not do this part in debug mode? Otherwise, you could continuously get the same CodePush update over and over while deploying new binaries to the simulator and device right?

geof90: Hmm, I thought [[CodePush bundleURL] and in turn [CodePush clearUpdates] never gets called in debug mode](https://github.com/Microsoft/react-native-code-push/blob/clear-updates-on-new-binary/CodePush.m#L229)?

lostintangent: That's would prevent bundleURL from getting called after an update is installed, but the initial app load would still trigger bundeURL of the dev uses it for the JS bundle location. While debugging with Chrome and the packager, they wouldn't call bundleURL, so you're right, this would never get hit. Thanks for the clarification!

android/app/src/main/java/com/microsoft/codepush/react/CodePush.java
```diff
@@ -299,11 +312,10 @@ public static void setUsingTestConfiguration(boolean shouldUseTestConfiguration)
         testConfigurationFlag = shouldUseTestConfiguration;
     }
 
-    public void clearTestUpdates() {
-        if (isUsingTestConfiguration()) {
-            codePushPackage.clearTestUpdates();
-            removePendingUpdate();
-        }
+    public void clearUpdates() {
+        codePushPackage.clearUpdates();
```
lostintangent: Since getBundleUrl is always called, even in debug mode, wouldn't this get called and cause some potential oddities in the dev workflow?

geof90: In Android, yes getBundleUrl will get called, but it would resolve to the package file if there is an update installed, so clearUpdates() wouldn't get called. The only time it would get called is if a new binary is applied, which I think makes sense even in the dev scenario. What do you think?

lostintangent: I just worry about getting an update available over and over again in dev mode when you're not bumping the version but you're redeploying new binaries. The update check wouldn't send a hash (since it was cleared) so the server would keep giving it the update. In iOS, that wouldn't happen due to the behavior we discussed, but on Android it seems like this could happen.

geof90: The trouble is knowing whether we are in dev mode from the CodePush class, since it is set in the parent project's BuildConfig. The only way I can think of right now is to have them pass in that value to us in the constructor or the call to getBundleUrl. Do you have a better suggestion?

lostintangent: As I think about it, it doesn't seem too bad to add a new parameter to the constructor called "debugMode" or something. We could use it for this and we could also only clear the local cache file of that value is true.

geof90: Okay, I have made the change to according to your feedback.

android/app/src/main/java/com/microsoft/codepush/react/CodePush.java
```diff
@@ -185,7 +203,9 @@ private void initializeUpdateAfterRestart() {
                     rollbackPackage();
                 } else {
                     // Clear the React dev bundle cache so that new updates can be loaded.
-                    clearReactDevBundleCache();
+                    if (!this.isDebugMode) {
```
lostintangent: Shouldn't this only happen in debug mode?

lostintangent:
LGTM!

PR #129, opened 2015-12-26
Adding conditional check to restartApp method

lostintangent:
This PR allows the `CodePush.restartApp` method to be called and only result in an app restart if there is actually a pending update to apply. It defaults to `false` to maintain back-compat as well as prevent confusion, since it's potentially unintuitive for the function to no-op unless you expect it to. 

This change makes it easier for an app to call `sync` with a non-immediate `InstallMode`, and then call `restartApp` in a well-known spot (e.g. navigating to the root component). This allows the app to discover/download/install updates silently, but apply them sooner if possible, without forcing superfluous app restarts, and without needing to rely on `getCurrentPackage`/`LocalPackage.isPending` to achieve this experience.

Aside from introducing `LocalPackage.isPending` (which makes a lot of sense semantically), this PR allows the bug issued in #110 to be achieved more simply.

### DIFF
--- a/README.md
+++ b/README.md
@@ -339,7 +339,7 @@ If you are using the `sync` function, and doing your update check on app start, 
 codePush.restartApp(onlyIfUpdateIsPending: Boolean = false): void;		
 ```		
 		
-Immediately restarts the app. If there is an update pending, it will be presented to the end user and the "rollback protection" feature will ensure it succeeds. Otherwise, calling this method simply has the same behavior as the end user killing and restarting the process. If a truthy value is passed to the `onlyIfUpdateIsPending` parameter, then the app will only be restarted if there is actually a pending update waiting to be applied. Otherwise, this method call will no-op.
+Immediately restarts the app. If a truthy value is provided to the `onlyIfUpdateIsPending` parameter, then the app will only restart if there is actually a pending update waiting to be applied.
 
 This method is for advanced scenarios, and is primarily useful when the following conditions are true:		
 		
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -336,12 +336,15 @@ If you are using the `sync` function, and doing your update check on app start,
 #### codePush.restartApp		
 		
 ```javascript		
-codePush.restartApp(): void;		
+codePush.restartApp(onlyIfUpdateIsPending: Boolean = false): void;		
 ```		
 		
-Immediately restarts the app. If there is an update pending, it will be presented to the end user and the rollback timer (if specified when installing the update) will begin. Otherwise, calling this method simply has the same behavior as the end user killing and restarting the process. This method is for advanced scenarios, and is primarily useful when the following conditions are true:		
+Immediately restarts the app. If there is an update pending, it will be presented to the end user and the rollback timer (if specified when installing the update) will begin. Otherwise, calling this method simply has the same behavior as the end user killing and restarting the process. If a truthy value is passed to the `onlyIfUpdateIsPending` parameter, then the app will only be restarted if there is actually a pending update waiting to be applied. Otherwise, this method call will no-op.
```
geof90: nit: Think the expression could be improved slightly, the way it is written now sounds like it will no-op if a falsey value was passed to the `onlyIfUpdateIsPending` parameter.

Maybe something like "If a truthy value is passed to the `onlyIfUpdateIsPending` parameter, then the method call will restart the app if there is actually a pending update waiting to be applied, or no-op otherwise."

geof90:
LGTM!

PR #130, opened 2015-12-26
Revert to module.exports

geof90:
This PR is needed because of https://phabricator.babeljs.io/T2212, so that babel@6 users can continue to use `require("react-native-code-push")` instead of having to do `require("react-native-code-push").default`.


lostintangent:
LGTM!

PR #131, opened 2015-12-28
Fixing immediate restart

lostintangent:
Immediately installed updates use the native bridge directly to restart the app, and not the JS-facade, and therefore, need to specify a value for the newly introduced `onlyIfUpdateIsPending` parameter. 

I also removed the `"use strict"` directives from the three JS files since Babel adds this automatically when using ES6 modules.


shishirx34:
LGTM

PR #135, opened 2015-12-30
MOD: Note on React Native 0.14 change the API of react-native bundle,…

shawnXiao:
… --out has been renamed for --bundle-output

### DIFF
--- a/README.md
+++ b/README.md
@@ -207,14 +207,14 @@ If an update is available, it will be silently downloaded, and installed the nex
 
 Once your app has been configured and distributed to your users, and you've made some JS changes, it's time to release it to them instantly! If you're not using React Native to bundle images (via `require('./foo.png')`), then perform the following steps. Otherwise, skip to the next section instead:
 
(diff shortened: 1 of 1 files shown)


lostintangent:
Hey @shawnXiao, `--out` was the pre-RN 0.14.0 name for this parameter. As of RN 0.14.0 (which is the minimum version the CodePush supports), this parameter is named `--bundle-output`. 

Which version of React Native are you using? Are you seeing the parameter show up as `--out`?

shawnXiao:
https://facebook.github.io/react-native/docs/running-on-device-ios.html

> Note that on 0.14 we'll change the API of react-native bundle. The major changes are:

>API is now entry-file <path> based instead of url based.
>Need to specify which platform you're bundling for --platform <ios|android>.
>Option --out has been renamed for --bundle-output.
>Source maps are no longer automatically generated. Need to specify --sourcemap-output <path>

lostintangent:
Those release notes are saying that the `--out` option was renamed to `--bundle-output`. The "for" should have been a "to" to be clearer.

lostintangent:
I'm going to close this PR since as mentioned, the current docs reflect the current parameter names for the RN bundle command. 

If you're using RN 0.14.0 and actually having an issue following our docs, then please re-submit this PR.

PR #136, opened 2015-12-30
Doc changes

lostintangent:
This PR makes the following changes:

1. Includes detail about the automatic rollback protection feature to the "How does it work?" section.

1. Updates our plugin consumption example to use the ES6 import module syntax instead of the CommonJS require syntax.

1. Documents the new constructor for the Java `CodePush` class, including details about what "debug mode" actually does.

1. Explains when and why the `checkForUpdate` method would return a `null` value.

1. Adds a few hyperlinks for better navigability throughout the document.

1. Adds a example to `sync` showing how to display a release's description in the update dialog.

### DIFF
--- a/README.md
+++ b/README.md
@@ -283,8 +283,13 @@ Queries the CodePush service to see whether the configured app deployment has an
 
 This method returns a `Promise` which resolves to one of two possible values:
 
-* `null` if there is no update available.
-* A `RemotePackage` instance which represents an available update that can be inspected and/or subsequently downloaded.
+1. `null` if there is no update available. This occurs in the following scenarios:
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -559,8 +581,14 @@ Constructs the CodePush client runtime and includes methods for integrating Code
 
 - __CodePush(String deploymentKey, Activity mainActivity)__ - Creates a new instance of the CodePush runtime, that will be used to query the service for updates via the provided deployment key. The `mainActivity` parameter should always be set to `this` when configuring your `ReactInstanceManager` inside the `MainActivity` class.
```
geof90: Probably state the default value of isDebugMode?

geof90:
LGTM!

PR #139, opened 2016-01-06
Fixing overriding _bundleURL on RN 0.18

skevy:
In RN 0.18 - _bundleURL became a private property. Import the "Private" category here to allow writing to it.

### DIFF
--- a/CodePush.m
+++ b/CodePush.m
@@ -1,4 +1,5 @@
 #import "RCTBridgeModule.h"
+#import "RCTBridge+Private.h"
 #import "RCTConvert.h"
 #import "RCTEventDispatcher.h"
 #import "RCTRootView.h"


lostintangent:
Thanks @skevy! I was planning to make this change this exact change this afternoon as a response to the bug you referenced. The change looks great to me :+1:

lostintangent:
Hey @skevy, what do you think about the #141 that @joenoon submitted? Your solution is the cleanest approach, but I'd like to maintain backwards compat for now, and the KVC solution allows for that.

skevy:
@lostintangent I think that's fine. Probably better. When we get a few versions down the line, we can use `RCTBridge+Private`.

lostintangent:
Awesome. Thanks @skevy!

PR #140, opened 2016-01-08
RN 0.18.0-rc fix bundleURL Assignment to readonly property

joenoon:
RN 0.18.0-rc made bundleURL readonly.  This adds a class extension to expose the bundleURL as readwrite to allow `loadBundle` to work.

### DIFF
--- a/CodePush.m
+++ b/CodePush.m
@@ -4,6 +4,10 @@
 #import "RCTRootView.h"
 #import "RCTUtils.h"
 
+@interface RCTBridge ()
+@property (nonatomic, strong, readwrite) NSURL *bundleURL;
+@end
+
 #import "CodePush.h"
 
 @implementation CodePush {


lostintangent:
Thanks @joenoon! PR #139 actually fixed this issue already, but I like this approach since it may be backwards compatible. Presumably this solution would support 0.18 as well as older RN versions? If so, that is super ideal.

joenoon:
@lostintangent Oh nice, I missed #139 !  It looks like the RCTBridge+Private.h file has only been around since Dec 15, 2015, so yes I guess this PR would be better for backward compatibility.  Its too bad because I like #139 more!

lostintangent:
Yep agreed. I hate having to duplicate code, but I'm also pretty sensitive about introducing breaking changes in our plugin, so it's a balancing act :) I'm going to run our tests using your fix and see if it is in fact backwards compatible, and if so, I think it will be a better solution for us right now, until we decide to actually take a new dependency on a different RN version.

lostintangent:
Unfortunately, this doesn't appear to be backwards compatible :( Apparently you can only redeclare a property as `readwrite` in a class extension if its primary interface is already declared as `readonly`. In pre-0.18, the `bundleURL` property is already `readwrite`, and therefore, you can't redeclare it as `readwrite` again.

Unless you know of a way to workaround that limitation, I think it makes sense to merge #139, if we can't find a backwards compatible fix to this breaking change.

@joenoon Thoughts?

lostintangent:
I'm going to close this PR in favor of us discussing the other two alternatives as better approaches. Thanks for the contribution and let me know if you think that is reasonable!

PR #141, opened 2016-01-09
alternative way to set property

joenoon:
This is the only other way I can think of.  re #139 and #140


lostintangent:
This is a great suggestion! I just verified that this does in fact work for 0.18 and earlier, so it's perfectly backwards compat :) I'm comfortable merging this in, but I'll wait for @geof90 to sign off as well.

geof90:
LGTM!

PR #147, opened 2016-01-14
Add missing slash

cmcewen:
Readme was missing a slash


lostintangent:
Good catch, thanks @cmcewen!

PR #150, opened 2016-01-15
Report Acquisition Status

geof90:
Goal: To allow client devices to report the status of their update acquisition (whether an update is successfully installed or rolled back, which update is the client running). The eventual goal is to allow developers to see this data from the CLI so that they can better manage their updates.

This PR: Configures the plugin to send a status report update everytime CodePush.notifiyApplicationReady is called. Keep track of the history of all status reports sent so that the same report will not be sent twice (even though the server knows how to handle repeat reports).

Note that this takes a dependency on code-push 1.5.0-beta, which has not been published to NPM yet. Therefore a release will be submitted for this change after the code-push SDK changes have also been release.

### DIFF
--- a/CodePush.js
+++ b/CodePush.js
@@ -103,9 +103,9 @@ function getPromisifiedSdk(requestFetchAdapter, config) {
     });
   };
 
-  sdk.reportStatusDeploy = (package, status) => {
+  sdk.reportStatusDeploy = (deployedPackage, status) => {
     return new Promise((resolve, reject) => {
-      module.exports.AcquisitionSdk.prototype.reportStatusDeploy.call(sdk, package, status, (err) => {
+      module.exports.AcquisitionSdk.prototype.reportStatusDeploy.call(sdk, deployedPackage, status, (err) => {
         if (err) {
           reject(err);
         } else {
@@ -115,9 +115,9 @@ function getPromisifiedSdk(requestFetchAdapter, config) {
     });
   };
 
-  sdk.reportStatusDownload = (package, status) => {
+  sdk.reportStatusDownload = (downloadedPackage, status) => {
     return new Promise((resolve, reject) => {
-      module.exports.AcquisitionSdk.prototype.reportStatusDownload.call(sdk, package, (err) => {
+      module.exports.AcquisitionSdk.prototype.reportStatusDownload.call(sdk, downloadedPackage, (err) => {
         if (err) {
           reject(err);
         } else {
--- a/CodePush.js
+++ b/CodePush.js
@@ -62,6 +62,7 @@ async function checkForUpdate(deploymentKey = null) {
   } else {     
     const remotePackage = { ...update, ...PackageMixins.remote };
     remotePackage.failedInstall = await NativeCodePush.isFailedUpdate(remotePackage.packageHash);
+    remotePackage.deploymentKey = deploymentKey || nativeConfig.deploymentKey;
     return remotePackage;
   }
 }
@@ -102,12 +103,41 @@ function getPromisifiedSdk(requestFetchAdapter, config) {
     });
   };
   
+  sdk.reportStatus = (package, status) => {
+    return new Promise((resolve, reject) => {
+      module.exports.AcquisitionSdk.prototype.reportStatusDeploy.call(sdk, package, status, (err) => {
+        if (err) {
+          reject(err);
+        } else {
+          resolve();
+        }
+      }); 
+    });
+  };
+  
   return sdk;
 }
 
 /* Logs messages to console with the [CodePush] prefix */
 function log(message) {
   console.log(`[CodePush] ${message}`)
+}
+
+async function notifyApplicationReady() {
+  await NativeCodePush.notifyApplicationReady();
+  const statusReport = await NativeCodePush.getNewStatusReport();
+  if (statusReport) {
+    const config = await getConfiguration();
+    if (statusReport.appVersion) {
+      config.appVersion = statusReport.appVersion;
+      const sdk = getPromisifiedSdk(requestFetchAdapter, config);
+      sdk.reportStatus();
+    } else {
+      config.deploymentKey = statusReport.package.deploymentKey;
+      const sdk = getPromisifiedSdk(requestFetchAdapter, config);
+      sdk.reportStatus(statusReport.package, statusReport.status);
+    }
+  }
 }
 
 function restartApp(onlyIfUpdateIsPending = false) {
@@ -269,7 +299,7 @@ const CodePush = {
   getConfiguration,
   getCurrentPackage,
   log,
-  notifyApplicationReady: NativeCodePush.notifyApplicationReady,
+  notifyApplicationReady,
   restartApp,
   setUpTestDependencies,
   sync,
--- a/CodePush.js
+++ b/CodePush.js
@@ -60,7 +60,7 @@ async function checkForUpdate(deploymentKey = null) {
   if (!update || update.updateAppVersion || (update.packageHash === localPackage.packageHash)) {
     return null;
   } else {     
-    const remotePackage = { ...update, ...PackageMixins.remote };
+    const remotePackage = { ...update, ...PackageMixins.remote(sdk.reportStatusDownload) };
     remotePackage.failedInstall = await NativeCodePush.isFailedUpdate(remotePackage.packageHash);
     remotePackage.deploymentKey = deploymentKey || nativeConfig.deploymentKey;
     return remotePackage;
@@ -102,8 +102,8 @@ function getPromisifiedSdk(requestFetchAdapter, config) {
       }); 
     });
   };
-  
-  sdk.reportStatus = (package, status) => {
+
+  sdk.reportStatusDeploy = (package, status) => {
     return new Promise((resolve, reject) => {
       module.exports.AcquisitionSdk.prototype.reportStatusDeploy.call(sdk, package, status, (err) => {
         if (err) {
@@ -114,7 +114,19 @@ function getPromisifiedSdk(requestFetchAdapter, config) {
       }); 
     });
   };
-  
+
+  sdk.reportStatusDownload = (package, status) => {
+    return new Promise((resolve, reject) => {
+      module.exports.AcquisitionSdk.prototype.reportStatusDownload.call(sdk, package, (err) => {
+        if (err) {
+          reject(err);
+        } else {
+          resolve();
+        }
+      }); 
+    });
+  };
+
   return sdk;
 }
 
@@ -129,13 +141,12 @@ async function notifyApplicationReady() {
   if (statusReport) {
     const config = await getConfiguration();
     if (statusReport.appVersion) {
-      config.appVersion = statusReport.appVersion;
       const sdk = getPromisifiedSdk(requestFetchAdapter, config);
-      sdk.reportStatus();
+      sdk.reportStatusDeploy();
     } else {
       config.deploymentKey = statusReport.package.deploymentKey;
       const sdk = getPromisifiedSdk(requestFetchAdapter, config);
-      sdk.reportStatus(statusReport.package, statusReport.status);
+      sdk.reportStatusDeploy(statusReport.package, statusReport.status);
     }
   }
 }
--- a/CodePush.js
+++ b/CodePush.js
@@ -115,7 +115,7 @@ function getPromisifiedSdk(requestFetchAdapter, config) {
     });
   };
 
-  sdk.reportStatusDownload = (downloadedPackage, status) => {
+  sdk.reportStatusDownload = (downloadedPackage) => {
     return new Promise((resolve, reject) => {
       module.exports.AcquisitionSdk.prototype.reportStatusDownload.call(sdk, downloadedPackage, (err) => {
         if (err) {
--- a/CodePush.m
+++ b/CodePush.m
@@ -14,6 +14,7 @@
 RCT_EXPORT_MODULE()
 
 static BOOL didRollback = NO;
+static BOOL isRunningBinaryVersion = NO;
 static BOOL testConfigurationFlag = NO;
 
 // These constants represent valid deployment statuses
@@ -62,6 +63,7 @@ static NSString *const PackageIsPendingKey = @"isPending";
     
     if (error || !packageFile) {
         NSLog(logMessageFormat, binaryJsBundleUrl);
+        isRunningBinaryVersion = YES;
         return binaryJsBundleUrl;
     }
     
@@ -73,6 +75,7 @@ static NSString *const PackageIsPendingKey = @"isPending";
     NSDictionary *currentPackageMetadata = [CodePushPackage getCurrentPackage:&error];
     if (error || !currentPackageMetadata) {
         NSLog(logMessageFormat, binaryJsBundleUrl);
+        isRunningBinaryVersion = YES;
         return binaryJsBundleUrl;
     }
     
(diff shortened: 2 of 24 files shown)


CodePush.m
```diff
@@ -468,6 +529,55 @@ - (void)savePendingUpdate:(NSString *)packageHash
...
+                resolve(@{ @"package": lastFailedPackage, @"status": DeploymentFailed });
+                return;
+            }
+        }
+    }
+    
+    // Check if the current CodePush package has been reported
+    NSError *error;
+    NSDictionary* currentPackage = [CodePushPackage getCurrentPackage:&error];
+    if (currentPackage) {
```
lostintangent: Couldn't this logic be driven off of "isFirstRun"? Presumably we should know whether we need to report a successful install without needing to check whether it has been reported already. Maybe I'm missing something, but I just want to make sure that we do as little as possible as part of the app start.

CodePush.m
```diff
@@ -468,6 +529,55 @@ - (void)savePendingUpdate:(NSString *)packageHash
...
+    if ([self isDeploymentStatusNotYetReported:appVersion]) {
+        [self recordDeploymentStatusReported:appVersion
+                                      status:DeploymentSucceeded];
+        resolve(@{ @"appVersion": appVersion });
+        return;
+    }
+    
+    // Check if there was a rollback that was not yet reported
+    NSUserDefaults *preferences = [NSUserDefaults standardUserDefaults];
+    NSMutableArray *failedUpdates = [preferences objectForKey:FailedUpdatesKey];
```
lostintangent: Could we just set a flag before doing a rollback that we then check here? I'd love to keep this function as simple as possible.

CodePush.js
```diff
@@ -110,6 +123,23 @@ function log(message) {
   console.log(`[CodePush] ${message}`)
 }
 
+async function notifyApplicationReady() {
+  await NativeCodePush.notifyApplicationReady();
+  const statusReport = await NativeCodePush.getNewStatusReport();
+  if (statusReport) {
+    const config = await getConfiguration();
+    if (statusReport.appVersion) {
+      config.appVersion = statusReport.appVersion;
```
lostintangent: Is this necessary? Wouldn't the status report already include the same app version that would be part of the config?

CodePushConfig.m
```diff
@@ -40,6 +50,7 @@ - (instancetype)init
                             buildVersion,BuildVdersionConfigKey,
                             serverURL,ServerURLConfigKey,
                             deploymentKey,DeploymentKeyConfigKey,
+                            clientUniqueId,ClientUniqueIDConfigKey,
```
lostintangent: This needs to be above the deployment key entry since it can be nil (e.g. the dev didn't add a key to the info.plist file), and nil will signal the end of the key pairs being added to the dictionary.

CodePushConfig.m
```diff
@@ -31,6 +33,14 @@ - (instancetype)init
     NSString *deploymentKey = [infoDictionary objectForKey:@"CodePushDeploymentKey"];
     NSString *serverURL = [infoDictionary objectForKey:@"CodePushServerURL"];
     
+    NSUserDefaults* userDefaults = [NSUserDefaults standardUserDefaults];
+    NSString* clientUniqueId = [userDefaults stringForKey:ClientUniqueIDConfigKey];
```
lostintangent: In what scenario would this value change that would make it beneficial for us to persist it?

CodePushConfig.m
```diff
@@ -10,6 +11,7 @@ @implementation CodePushConfig {
 static NSString * const BuildVdersionConfigKey = @"buildVersion";
 static NSString * const DeploymentKeyConfigKey = @"deploymentKey";
 static NSString * const ServerURLConfigKey = @"serverUrl";
+static NSString * const ClientUniqueIDConfigKey = @"clientUniqueId";
```
lostintangent: Alpha sort

CodePush.m
geof90: Could notifyApplicationReady be called multiple times? Suppose the developer puts notifyApplicationReady in componentDidMount() to one of the components in the app, perhaps the main page. If the user then switches to a different page/component and comes back later, isFirstRun would still evaluate to true.

CodePush.m
geof90: We could. Although if the flag evaluates to true, we still need to fetch the latest failed update's metadata. The flag would serve as an optimization and prevent a disk hit if the app did not rollback.

CodePushConfig.m
geof90: It is an iOS bug. identifierForVendor sometimes changes when an app is updated. http://stackoverflow.com/a/33130588/2670678

I have seen other plugins such as cordova-plugin-device that also persist the value to workaround this problem. Seems like its safer to just keep this.

CodePush.m
lostintangent: The scenario where there isn't anything to report is exactly what I'd like to optimize. Ideally, calling notifyApplicationReady should have an extremely low impact on an app that hasn't even released an update yet. Maybe what you have is fine, i just wanted to mention the desire to keep app start as lean as possible, since this will run every time, but be unnecessary most of the time.

CodePush.m
geof90: I'll do an optimization to check isFirstRun first, then check whether it's been reported.

request-fetch-adapter.js
```diff
@@ -28,4 +28,18 @@ module.exports = {
       callback(err);
     }
   }
-};
\ No newline at end of file
+};
+
+function getHttpMethodName(verb) {
```
dtivel: Should add a comment indicating that this should stay in sync with the CodePush SDK.  It's not obvious here that the array below should be synchronized with the SDK's Verb enum (https://github.com/Microsoft/code-push/blob/master/sdk/script/acquisition-sdk.ts#L6).

CodePush.m
```diff
@@ -468,6 +532,59 @@ - (void)savePendingUpdate:(NSString *)packageHash
...
+ */
+RCT_EXPORT_METHOD(getNewStatusReport:(RCTPromiseResolveBlock)resolve
+                            rejecter:(RCTPromiseRejectBlock)reject)
+{
+    if (didRollback) {
+        // Check if there was a rollback that was not yet reported
+        NSUserDefaults *preferences = [NSUserDefaults standardUserDefaults];
+        NSMutableArray *failedUpdates = [preferences objectForKey:FailedUpdatesKey];
+        if (failedUpdates) {
+            NSDictionary* lastFailedPackage = [failedUpdates lastObject];
```
dtivel: The convention is to put * with the variable name not with the type name.  Also applies below.

CodePush.m
```diff
@@ -237,6 +285,22 @@ - (void)loadBundle
...
+{
+    NSUserDefaults *preferences = [NSUserDefaults standardUserDefaults];
+    NSMutableDictionary *sentStatusReports = [preferences objectForKey:StatusReportsKey];
+    if (sentStatusReports == nil) {
+        sentStatusReports = [NSMutableDictionary dictionary];
+    } else {
+        sentStatusReports = [sentStatusReports mutableCopy];
+    }
+    
+    [sentStatusReports setValue:status forKey:appVersionOrPackageIdentifier];
```
dtivel: This dictionary will be ever growing, never shrinking?

Can't we clear it at some point, such as before adding a successful deployment status?

dtivel:
Looks good.

request-fetch-adapter.js
geof90: Added!

lostintangent:
LGTM!

CodePush.js
```diff
@@ -101,7 +102,31 @@ function getPromisifiedSdk(requestFetchAdapter, config) {
       }); 
     });
   };
-  
+
+  sdk.reportStatusDeploy = (package, status) => {
```
lostintangent: We can't use the identifier `package` since Babel will reject it.

CodePush.js
```diff
@@ -101,7 +102,31 @@ function getPromisifiedSdk(requestFetchAdapter, config) {
...
+        if (err) {
+          reject(err);
+        } else {
+          resolve();
+        }
+      }); 
+    });
+  };
+
+  sdk.reportStatusDownload = (package, status) => {
```
lostintangent: We can't use the identifier `package` here.

package.json
```diff
@@ -16,7 +16,7 @@
     "url": "https://github.com/Microsoft/react-native-code-push"
   },
   "dependencies": {
-    "code-push": "^1.1.1-beta",
+    "code-push": "1.5.0-beta",
```
lostintangent: We'll need to update this to `1.5.1-beta` to accommodate the `package` fix I have a PR out for.

CodePushConfig.m
```diff
@@ -31,6 +33,9 @@ - (instancetype)init
     NSString *deploymentKey = [infoDictionary objectForKey:@"CodePushDeploymentKey"];
     NSString *serverURL = [infoDictionary objectForKey:@"CodePushServerURL"];
     
+    NSUserDefaults* userDefaults = [NSUserDefaults standardUserDefaults];
```
lostintangent: Xcode complains about this variable not being used. Should we cache the UUID in defaults after all? I noticed that is what the Cordova device plugin does, and I wonder if that helps support edge-case scenarios where a new UUID would be generated when upgrading an existing binary installation.

geof90: I think we should.

lostintangent:
BTW, I notice that the "Loading JS bundle from.." message is logged in Xcode's console every time my app calls `notifyApplicationReady`. I didn't debug it yet, but this seems like an unexpected bug. Could you take a look at this and make sure something isn't being unnecessarily triggered as part of the new metrics feature?

CodePush.m
```diff
@@ -468,6 +528,60 @@ - (void)savePendingUpdate:(NSString *)packageHash
...
+                    [self recordDeploymentStatusReported:lastFailedPackageIdentifier];
+                    resolve(@{ @"package": lastFailedPackage, @"status": DeploymentFailed });
+                    return;
+                }
+            }
+        }
+    } else if (_isFirstRunAfterUpdate) {
+        // Check if the current CodePush package has been reported
+        NSError *error;
+        NSDictionary* currentPackage = [CodePushPackage getCurrentPackage:&error];
```
lostintangent: `NSDictionary *currentPackage`. Also, since you're not doing anything with the error, would it make sense to just pass `nil` as the error parameter?

CodePush.m
```diff
@@ -468,6 +528,60 @@ - (void)savePendingUpdate:(NSString *)packageHash
...
+            }
+        }
+    } else {
+        if (isRunningBinaryVersion) {
+            // Check if the current appVersion has been reported. Use date as the binary identifier to
+            // handle binary releases that do not modify the appVersion.
+            NSURL *binaryJsBundleUrl = [CodePush bundleURL];
+            NSDictionary *binaryFileAttributes = [[NSFileManager defaultManager] attributesOfItemAtPath:[binaryJsBundleUrl path] error:nil];
+            NSTimeInterval binaryDate = [[binaryFileAttributes objectForKey:NSFileModificationDate] timeIntervalSince1970];
+            NSString* binaryIdentifier = [NSString stringWithFormat:@"%f", binaryDate];
```
lostintangent: `NSString *binaryIdentifier`

CodePushConfig.m
lostintangent: `NSString *clientUniqueId`

CodePush.m
```diff
@@ -468,6 +528,60 @@ - (void)savePendingUpdate:(NSString *)packageHash
...
+                [self recordDeploymentStatusReported:currentPackageIdentifier];
+                resolve(@{ @"package": currentPackage, @"status": DeploymentSucceeded });
+                return;
+            }
+        }
+    } else {
+        if (isRunningBinaryVersion) {
+            // Check if the current appVersion has been reported. Use date as the binary identifier to
+            // handle binary releases that do not modify the appVersion.
+            NSURL *binaryJsBundleUrl = [CodePush bundleURL];
```
lostintangent: I'm not sure if it makes sense to call the `bundleURL` method here. Our current expectations for debugging are that this method isn't called while debugging with Chrome, so I'm not sure what negative side-effects might occur if this is now called.

CodePush.m
```diff
@@ -13,11 +13,18 @@ @implementation CodePush {
...
 static BOOL testConfigurationFlag = NO;
 
+// These constants represent valid deployment statuses
+static NSString *const DeploymentSucceeded = @"DeploymentSucceeded";
+static NSString *const DeploymentFailed = @"DeploymentFailed";
+
 // These keys represent the names we use to store data in NSUserDefaults
 static NSString *const FailedUpdatesKey = @"CODE_PUSH_FAILED_UPDATES";
 static NSString *const PendingUpdateKey = @"CODE_PUSH_PENDING_UPDATE";
+static NSString *const LastDeploymentReportKey = @"CODE_PUSH_LAST_DEPLOYMENT_REPORT";
```
lostintangent: Alpha sort

CodePush.m
```diff
@@ -13,11 +13,18 @@ @implementation CodePush {
 
 RCT_EXPORT_MODULE()
 
+static BOOL didRollback = NO;
+static BOOL isRunningBinaryVersion = NO;
 static BOOL testConfigurationFlag = NO;
 
+// These constants represent valid deployment statuses
+static NSString *const DeploymentSucceeded = @"DeploymentSucceeded";
+static NSString *const DeploymentFailed = @"DeploymentFailed";
```
lostintangent: Sort these two

CodePush.m
```diff
@@ -468,6 +528,60 @@ - (void)savePendingUpdate:(NSString *)packageHash
...
+        NSDictionary* currentPackage = [CodePushPackage getCurrentPackage:&error];
+        if (currentPackage) {
+            NSString* currentPackageIdentifier = [self getPackageStatusReportIdentifier:currentPackage];
+            if (currentPackageIdentifier && [self isDeploymentStatusNotYetReported:currentPackageIdentifier]) {
+                [self recordDeploymentStatusReported:currentPackageIdentifier];
+                resolve(@{ @"package": currentPackage, @"status": DeploymentSucceeded });
+                return;
+            }
+        }
+    } else {
```
lostintangent: Should this be an `else if`?

CodePush.m
```diff
@@ -468,6 +528,60 @@ - (void)savePendingUpdate:(NSString *)packageHash
...
+        if (isRunningBinaryVersion) {
```
lostintangent: This condition seems to run every time I call `notifyApplicationReady`, so I'm not sure whether the report is being saved or not.

CodePush.m
```diff
@@ -468,6 +528,53 @@ - (void)savePendingUpdate:(NSString *)packageHash
...
+            NSString *currentPackageIdentifier = [self getPackageStatusReportIdentifier:currentPackage];
+            if (currentPackageIdentifier && [self isDeploymentStatusNotYetReported:currentPackageIdentifier]) {
+                [self recordDeploymentStatusReported:currentPackageIdentifier];
+                resolve(@{ @"package": currentPackage, @"status": DeploymentSucceeded });
+                return;
+            }
+        }
+    } else if (isRunningBinaryVersion || [_bridge.bundleURL.scheme hasPrefix:@"http"]) {
+        // Check if the current appVersion has been reported.
+        NSString *appVersion = [[CodePushConfig current] appVersion];
```
lostintangent: During scenarios were the appVersion doesn't need to change, this wouldn't re-run (e.g. development, TestFlight) correct?

PR #156, opened 2016-01-22
Preliminary Android 0.18.0 support

lostintangent:
This PR adds basic support for Android in React Native 0.18.0 by making the following changes:

1. Removing the restriction of the `MainActivitity.java` file to inherit from `FragmentActivity` in order to display a non-silent update. Since the new `ReactActivity` base class doesn't inherit from `FragmentActivity`, this limitation can't be enforced. 

2. Adds configuration instructions for 0.18.0, which are distinct from 0.15.0-0.17.0. Moving forward, I'd like to improve this setup experience, so this is just a temporary workaround to unblock the scenario.


geof90:
LGTM!

Issue #157, opened 2016-01-22
The guideline should state that `.bundle` or `.jsbundle` file extension of bundle output is required for Releasing Updates (JavaScript + images)


alexlau811:
Although the sample showed `--bundle-output ./release/main.jsbundle`, no documentation states that this does not accept other extensions besides `.bundle` or `.jsbundle`.

If a developer passed other extensions (e.g. `index.ios.js`), the update will stuck and no error messages are prompted nor logged.

(I have spent a morning to figure it out why my app won't update. 😭)

lostintangent:
This is a very good point! I apologize for the confusion and lack of docs here. We should accept files with the `.js`, `.jsbundle`, and `.bundle` extensions, so I'll use this bug to track adding support for `.js` files, as well as updating the docs to specify that the extension is relevant.

lostintangent:
Hey @alexlau811! Check out the PR I just submitted and let me know what you think. Thanks again for the feedback here! It means a lot to us since it can be easy to glaze over things that aren't documented well enough :)

alexlau811:
Wow! Your action is just so quick and that PR is super great. Thank you so much!

lostintangent:
Great! Thanks dude. I just merged this PR. If you're cool using the `.jsbundle` or `.bundle` extensions for now, I'll release an update to NPM later that adds the support for `.js` files as well.

PR #158, opened 2016-01-22
Adding support for .js extension

lostintangent:
This PR extends our current iOS support by allowing the JS bundle file to use a `.js` extension in addition to `.jsbundle` and `.bundle`. Additionally, I added documentation to clarify that the file extension is in fact relevant to the update process when shipping JS + images.

Both of these enhancements are meant to address issue #157.


geof90:
LGTM!

PR #162, opened 2016-01-25
Update README.md

talkol:
Make JS code examples consistent between CodePush and codePush (same case of the first letter)


lostintangent:
Good call @talkol! I'll merge this change right now.

Issue #161, opened 2016-01-24
React Native 0.19.0-RC changed the signature of RCTPromiseRejectBlock


lostintangent:
The RCTPromiseRejectBlock typedef was changed to accept three arguments, instead of one, and therefore, anyone trying to use CodePush with React Native 0.19.0-RC will have compilation errors on iOS. 

This is an obviously easy fix to make, but I'd like to add support for RN 0.19.0 without having to break compat with previous versions of RN. I validated the behavior of our plugin with this new platform version and everything works great, other than our need to respond to this typedef change.

respectTheCode:
Any progress on this? RN 0.19 has been released.

lostintangent:
@respectTheCode We'll be releasing the fix to NPM today. If you need to get unblocked sooner, you can take a look at PR #164 for the fix. Sorry for the delay!

lostintangent:
This fix is now in master and will be released to NPM in the morning.

PR #164, opened 2016-01-26
Use 3-argument RCTPromiseRejectBlock

jedlau:
Addresses https://github.com/Microsoft/react-native-code-push/issues/161.

### DIFF
--- a/CodePush.m
+++ b/CodePush.m
@@ -406,14 +406,14 @@ RCT_EXPORT_METHOD(downloadUpdate:(NSDictionary*)updatePackage
             NSDictionary *newPackage = [CodePushPackage getPackage:updatePackage[PackageHashKey] error:&err];
                 
             if (err) {
-                return reject(err);
+                return reject([NSString stringWithFormat: @"%lu", (long)err.code], err.localizedDescription, err);
             }
                 
             resolve(newPackage);
         }
         // The download failed
         failCallback:^(NSError *err) {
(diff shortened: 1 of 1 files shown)


lostintangent:
Thanks @jedlau! This is the perfect solution. The only reason I'm somewhat waiting to merge this in is because I would ideally love to have a solution that was backwards compatible with previous React Native versions. I'm not aware of a way to do this, so I posted the [question](https://github.com/facebook/react-native/issues/5524) on the RN repro to get some advice, and maybe even propose reverting the change in the first place.

ecjs:
Seeing how RN 0.19 was released, it would be awesome to see this merged.  Thanks for the great work on this, loving it so far.

lostintangent:
I've been waiting to merge it because we're planning to release an update tomorrow to NPM that addresses some bugs/etc., and I don't want that release to require RN 0.19.0. Additionally, I've been petitioning that this breaking change be reverted in RN, since it's a huge shame to start requiring RN 0.19 for our users moving forward. It's looking unlikely that that will happen, so we'll likely merge this change in sometime tomorrow and release a patch update to address RN 0.19.0.

Sorry for the delay!

ecjs:
No problem, sounds good!

geof90:
Hi @jedlau, I will close this because I have incorporated the required changes in https://github.com/Microsoft/react-native-code-push/pull/172.

PR #166, opened 2016-01-27
Metrics SDK Changes

geof90:
Dependency on https://github.com/Microsoft/code-push/pull/126

Sends previous version label and deployment key along with the deployment status report.

### DIFF
--- a/CodePush.h
+++ b/CodePush.h
@@ -91,6 +91,9 @@ failCallback:(void (^)(NSError *err))failCallback;
 @interface CodePushStatusReport : NSObject
 
 + (NSString *)getDeploymentKeyFromStatusReportIdentifier:(NSString *)statusReportIdentifier;
++ (NSDictionary *)getFailedUpdateStatusReport:(NSDictionary *)lastFailedPackage;
++ (NSDictionary *)getNewPackageStatusReport:(NSDictionary *)currentPackage;
++ (NSDictionary *)getNewAppVersionStatusReport:(NSString *)appVersion;
 + (NSString *)getPackageStatusReportIdentifier:(NSDictionary *)package;
 + (NSString *)getPreviousStatusReportIdentifier;
 + (NSString *)getVersionLabelFromStatusReportIdentifier:(NSString *)statusReportIdentifier;
--- a/CodePush.h
+++ b/CodePush.h
@@ -88,6 +88,17 @@ failCallback:(void (^)(NSError *err))failCallback;
 
 @end
 
+@interface CodePushStatusReport : NSObject
+
++ (NSString *)getDeploymentKeyFromStatusReportIdentifier:(NSString *)statusReportIdentifier;
++ (NSString *)getPackageStatusReportIdentifier:(NSDictionary *)package;
++ (NSString *)getPreviousStatusReportIdentifier;
++ (NSString *)getVersionLabelFromStatusReportIdentifier:(NSString *)statusReportIdentifier;
++ (BOOL)isStatusReportIdentifierCodePushLabel:(NSString *)statusReportIdentifier;
++ (void)recordDeploymentStatusReported:(NSString *)appVersionOrPackageIdentifier;
+
+@end
+
 typedef NS_ENUM(NSInteger, CodePushInstallMode) {
     CodePushInstallModeImmediate,
     CodePushInstallModeOnNextRestart,
--- a/CodePush.h
+++ b/CodePush.h
@@ -91,8 +91,8 @@ failCallback:(void (^)(NSError *err))failCallback;
 @interface CodePushTelemetryManager : NSObject
 
 + (NSDictionary *)getBinaryUpdateReport:(NSString *)appVersion;
-+ (NSDictionary *)getUpdateReport:(NSDictionary *)currentPackage;
 + (NSDictionary *)getRollbackReport:(NSDictionary *)lastFailedPackage;
++ (NSDictionary *)getUpdateReport:(NSDictionary *)currentPackage;
 
 @end
 
--- a/CodePush.h
+++ b/CodePush.h
@@ -88,17 +88,11 @@ failCallback:(void (^)(NSError *err))failCallback;
 
 @end
 
-@interface CodePushStatusReport : NSObject
+@interface CodePushTelemetryManager : NSObject
 
-+ (NSString *)getDeploymentKeyFromStatusReportIdentifier:(NSString *)statusReportIdentifier;
-+ (NSDictionary *)getFailedUpdateStatusReport:(NSDictionary *)lastFailedPackage;
-+ (NSDictionary *)getNewPackageStatusReport:(NSDictionary *)currentPackage;
-+ (NSDictionary *)getNewAppVersionStatusReport:(NSString *)appVersion;
-+ (NSString *)getPackageStatusReportIdentifier:(NSDictionary *)package;
-+ (NSString *)getPreviousStatusReportIdentifier;
-+ (NSString *)getVersionLabelFromStatusReportIdentifier:(NSString *)statusReportIdentifier;
-+ (BOOL)isStatusReportIdentifierCodePushLabel:(NSString *)statusReportIdentifier;
-+ (void)recordDeploymentStatusReported:(NSString *)appVersionOrPackageIdentifier;
++ (NSDictionary *)getBinaryUpdateReport:(NSString *)appVersion;
++ (NSDictionary *)getUpdateReport:(NSDictionary *)currentPackage;
++ (NSDictionary *)getRollbackReport:(NSDictionary *)lastFailedPackage;
 
 @end
 
--- a/CodePush.m
+++ b/CodePush.m
@@ -503,32 +503,35 @@ RCT_EXPORT_METHOD(notifyApplicationReady:(RCTPromiseResolveBlock)resolve
 RCT_EXPORT_METHOD(getNewStatusReport:(RCTPromiseResolveBlock)resolve
                             rejecter:(RCTPromiseRejectBlock)reject)
 {
-    if (needToReportRollback) {
-        needToReportRollback = NO;
-        NSUserDefaults *preferences = [NSUserDefaults standardUserDefaults];
-        NSMutableArray *failedUpdates = [preferences objectForKey:FailedUpdatesKey];
-        if (failedUpdates) {
-            NSDictionary *lastFailedPackage = [failedUpdates lastObject];
-            if (lastFailedPackage) {
-                resolve([CodePushTelemetryManager getRollbackReport:lastFailedPackage]);
+    
+    dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
+        if (needToReportRollback) {
+            needToReportRollback = NO;
+            NSUserDefaults *preferences = [NSUserDefaults standardUserDefaults];
+            NSMutableArray *failedUpdates = [preferences objectForKey:FailedUpdatesKey];
+            if (failedUpdates) {
+                NSDictionary *lastFailedPackage = [failedUpdates lastObject];
+                if (lastFailedPackage) {
+                    resolve([CodePushTelemetryManager getRollbackReport:lastFailedPackage]);
+                    return;
+                }
+            }
+        } else if (_isFirstRunAfterUpdate) {
+            NSError *error;
+            NSDictionary *currentPackage = [CodePushPackage getCurrentPackage:&error];
+            if (!error && currentPackage) {
+                resolve([CodePushTelemetryManager getUpdateReport:currentPackage]);
                 return;
             }
-        }
-    } else if (_isFirstRunAfterUpdate) {
-        NSError *error;
-        NSDictionary *currentPackage = [CodePushPackage getCurrentPackage:&error];
-        if (!error && currentPackage) {
-            resolve([CodePushTelemetryManager getUpdateReport:currentPackage]);
+        } else if (isRunningBinaryVersion || [_bridge.bundleURL.scheme hasPrefix:@"http"]) {
+            // Check if the current appVersion has been reported.
+            NSString *appVersion = [[CodePushConfig current] appVersion];
+            resolve([CodePushTelemetryManager getBinaryUpdateReport:appVersion]);
             return;
         }
-    } else if (isRunningBinaryVersion || [_bridge.bundleURL.scheme hasPrefix:@"http"]) {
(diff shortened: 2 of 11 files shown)


request-fetch-adapter.js
```diff
@@ -4,10 +4,11 @@ module.exports = {
...
       requestBody = null;
     }
-
-    var headers = {
+    
+    const headers = {
       "Accept": "application/json",
-      "Content-Type": "application/json"
+      "Content-Type": "application/json",
+      "X-SDK-Version": getSDKVersion()
```
lostintangent: Should we include "CodePush" in this name somewhere? (e.g. "X-CodePush-SDK-Version")

CodePush.m
```diff
@@ -161,6 +161,11 @@ - (void)dealloc
     [[NSNotificationCenter defaultCenter] removeObserver:self];
 }
 
+- (NSString *)getDeploymentKeyFromStatusReportIdentifier:(NSString *)statusReportIdentifier
```
lostintangent: What do you think about moving all these report status methods out into their own class? The CodePush class is getting kind of cluttered and I'd like to start refactoring it out a little bit more.

request-fetch-adapter.js
geof90: Added

CodePush.m
geof90: I think its a good idea, refactored.

CodePush.m
```diff
@@ -541,12 +511,11 @@ - (void)savePendingUpdate:(NSString *)packageHash
...
-                NSString *lastFailedPackageIdentifier = [self getPackageStatusReportIdentifier:lastFailedPackage];
-                if (lastFailedPackageIdentifier && [self isDeploymentStatusNotYetReported:lastFailedPackageIdentifier]) {
-                    [self recordDeploymentStatusReported:lastFailedPackageIdentifier];
-                    resolve(@{ @"package": lastFailedPackage, @"status": DeploymentFailed });
-                    return;
-                }
+                resolve(@{
+                          @"package": lastFailedPackage,
+                          @"status": DeploymentFailed
+                          });
```
lostintangent: Align the curly braces?

CodePush.m
```diff
@@ -554,20 +523,65 @@ - (void)savePendingUpdate:(NSString *)packageHash
...
-            if (currentPackageIdentifier && [self isDeploymentStatusNotYetReported:currentPackageIdentifier]) {
-                [self recordDeploymentStatusReported:currentPackageIdentifier];
-                resolve(@{ @"package": currentPackage, @"status": DeploymentSucceeded });
-                return;
+            NSString *currentPackageIdentifier = [CodePushStatusReport getPackageStatusReportIdentifier:currentPackage];
+            NSString *previousStatusReportIdentifier = [CodePushStatusReport getPreviousStatusReportIdentifier];
+            if (currentPackageIdentifier) {
+                if (previousStatusReportIdentifier == nil) {
+                    [CodePushStatusReport recordDeploymentStatusReported:currentPackageIdentifier];
+                    resolve(@{
```
lostintangent: Could we refactor this code a little bit to resolve/return in a single place? Maybe what you have is fine, it just seems like a little bit of duplication.

CodePushStatusReport.m
```diff
@@ -0,0 +1,51 @@
+#import "CodePush.h"
+
+static NSString *const LastDeploymentReportKey = @"CODE_PUSH_LAST_DEPLOYMENT_REPORT";
+static NSString *const DeploymentKeyKey = @"deploymentKey";
+static NSString *const LabelKey = @"label";
```
lostintangent: Sort?

CodePush.m
```diff
@@ -554,20 +523,65 @@ - (void)savePendingUpdate:(NSString *)packageHash
...
-        if ([self isDeploymentStatusNotYetReported:appVersion]) {
-            [self recordDeploymentStatusReported:appVersion];
+        NSString *previousStatusReportIdentifier = [CodePushStatusReport getPreviousStatusReportIdentifier];
+        if (previousStatusReportIdentifier == nil) {
+            [CodePushStatusReport recordDeploymentStatusReported:appVersion];
             resolve(@{ @"appVersion": appVersion });
             return;
+        } else if (![previousStatusReportIdentifier isEqualToString:appVersion]) {
+            [CodePushStatusReport recordDeploymentStatusReported:appVersion];
+            if ([CodePushStatusReport isStatusReportIdentifierCodePushLabel:previousStatusReportIdentifier]) {
```
lostintangent: I'm not sure I fully understand this code. If the `previousStatusReportIdentifier` isn't `nil`, when would it ever be set to an invalid report identifier? The code for this seems to simply check for `nil` (which you already do), and then check to see if it includes a colon. Do we need this check at all?

CodePush.m
```diff
@@ -554,20 +523,65 @@ - (void)savePendingUpdate:(NSString *)packageHash
...
+                    [CodePushStatusReport recordDeploymentStatusReported:currentPackageIdentifier];
+                    resolve(@{
+                              @"package": currentPackage,
+                              @"status": DeploymentSucceeded
+                              });
+                    return;
+                } else if (![previousStatusReportIdentifier isEqualToString:currentPackageIdentifier]) {
+                    [CodePushStatusReport recordDeploymentStatusReported:currentPackageIdentifier];
+                    if ([CodePushStatusReport isStatusReportIdentifierCodePushLabel:previousStatusReportIdentifier]) {
+                        NSString *previousDeploymentKey = [CodePushStatusReport getDeploymentKeyFromStatusReportIdentifier:previousStatusReportIdentifier];
```
lostintangent: I wonder whether it wouldn't be cleaner to have a single method that can retrieve the previous status report, and would return either `nil` or an object that already includes the parsed deployment key and label. It feels like this code knows more about how the status reports are persisted then it needs to.

CodePush.m
```diff
@@ -554,20 +523,65 @@ - (void)savePendingUpdate:(NSString *)packageHash
...
+            NSString *previousStatusReportIdentifier = [CodePushStatusReport getPreviousStatusReportIdentifier];
+            if (currentPackageIdentifier) {
+                if (previousStatusReportIdentifier == nil) {
+                    [CodePushStatusReport recordDeploymentStatusReported:currentPackageIdentifier];
+                    resolve(@{
+                              @"package": currentPackage,
+                              @"status": DeploymentSucceeded
+                              });
+                    return;
+                } else if (![previousStatusReportIdentifier isEqualToString:currentPackageIdentifier]) {
```
lostintangent: When would the current and previous update has the same identifier? Is that even possible?

CodePush.m
geof90: If it is an appVersion identifier, it does not include a colon. Hence the name `isStatusReportIdentifierCodePushLabel`, it tells you whether it is a CodePush label or it is just a native app version. I agree it would be much cleaner if all this was in the CodePushStatusReport class, i'll push a change in a bit to move that there.

CodePush.m
geof90: Yes if notifyApplicationReady() is called more than once.

CodePush.h
```diff
@@ -88,6 +88,20 @@ failCallback:(void (^)(NSError *err))failCallback;
 
 @end
 
+@interface CodePushStatusReport : NSObject
+
++ (NSString *)getDeploymentKeyFromStatusReportIdentifier:(NSString *)statusReportIdentifier;
```
lostintangent: We can probably remove all of the internal methods from the interface, since the `CodePush` class doesn't consume them.

CodePushStatusReport.m
```diff
@@ -0,0 +1,125 @@
...
+static NSString *const LastDeploymentReportKey = @"CODE_PUSH_LAST_DEPLOYMENT_REPORT";
+
+@implementation CodePushStatusReport
+
++ (NSString *)getDeploymentKeyFromStatusReportIdentifier:(NSString *)statusReportIdentifier
+{
+    return [[statusReportIdentifier componentsSeparatedByString:@":"] firstObject];
+}
+
++ (NSDictionary *)getFailedUpdateStatusReport:(NSDictionary *)lastFailedPackage
```
lostintangent: What about `getRollbackStatusReport`?

CodePushStatusReport.m
```diff
@@ -0,0 +1,125 @@
...
+{
+    return @{
+              @"package": lastFailedPackage,
+              @"status": DeploymentFailed
+            };
+}
+
++ (NSDictionary *)getNewPackageStatusReport:(NSDictionary *)currentPackage
```
lostintangent: What about `getUpdateStatusReport`?

CodePushStatusReport.m
```diff
@@ -0,0 +1,125 @@
...
+                         @"previousLabelOrAppVersion": previousStatusReportIdentifier
+                         };
+            }
+        }
+    }
+    
+    return nil;
+}
+
++ (NSDictionary *)getNewAppVersionStatusReport:(NSString *)appVersion
```
lostintangent: What about `getBinaryUpdateStatusReport`?

CodePushStatusReport.m
```diff
@@ -0,0 +1,125 @@
+#import "CodePush.h"
+
+static NSString *const DeploymentFailed = @"DeploymentFailed";
+static NSString *const DeploymentKeyKey = @"deploymentKey";
+static NSString *const DeploymentSucceeded = @"DeploymentSucceeded";
+static NSString *const LabelKey = @"label";
+static NSString *const LastDeploymentReportKey = @"CODE_PUSH_LAST_DEPLOYMENT_REPORT";
+
+@implementation CodePushStatusReport
```
lostintangent: What about calling this something like `CodePushTelemetryManager`?

geof90:
GitHub is back up!

CodePush.h
```diff
@@ -88,6 +88,14 @@ failCallback:(void (^)(NSError *err))failCallback;
 
 @end
 
+@interface CodePushTelemetryManager : NSObject
+
++ (NSDictionary *)getBinaryUpdateReport:(NSString *)appVersion;
++ (NSDictionary *)getUpdateReport:(NSDictionary *)currentPackage;
++ (NSDictionary *)getRollbackReport:(NSDictionary *)lastFailedPackage;
```
lostintangent: Move this above for sorting purposes?

CodePushTelemetryManager.m
```diff
@@ -0,0 +1,125 @@
...
+                         @"previousLabelOrAppVersion": previousStatusReportIdentifier
+                         };
+            }
+        }
+    }
+    
+    return nil;
+}
+
++ (NSString *)getDeploymentKeyFromStatusReportIdentifier:(NSString *)statusReportIdentifier
```
lostintangent: Since you're not sorting all methods together, what do you think about adding a "#pragma mark - Private methods" line above this group so it's clear what the two method groups are

android/app/src/main/java/com/microsoft/codepush/react/CodePush.java
```diff
@@ -51,24 +51,21 @@
...
-    private final String LAST_DEPLOYMENT_REPORT_KEY = "CODE_PUSH_LAST_DEPLOYMENT_REPORT";
 
     // This needs to be kept in sync with https://github.com/facebook/react-native/blob/master/ReactAndroid/src/main/java/com/facebook/react/devsupport/DevSupportManager.java#L78
     private final String REACT_DEV_BUNDLE_CACHE_FILE_NAME = "ReactNativeDevBundle.js";
 
+    // Helper classes.
     private CodePushPackage codePushPackage;
     private CodePushReactPackage codePushReactPackage;
+    private CodePushTelemetryManager codePushTelemetryManager;
     private CodePushNativeModule codePushNativeModule;
```
lostintangent: Move this to the top of the these four lines?

lostintangent:
LGTM!

CodePushTelemetryManager.m
```diff
@@ -0,0 +1,125 @@
...
+    } else if (![previousStatusReportIdentifier isEqualToString:appVersion]) {
+        [self recordDeploymentStatusReported:appVersion];
+        if ([self isStatusReportIdentifierCodePushLabel:previousStatusReportIdentifier]) {
+            NSString *previousDeploymentKey = [self getDeploymentKeyFromStatusReportIdentifier:previousStatusReportIdentifier];
+            NSString *previousLabel = [self getVersionLabelFromStatusReportIdentifier:previousStatusReportIdentifier];
+            return @{
+                     @"appVersion": appVersion,
+                     @"previousDeploymentKey": previousDeploymentKey,
+                     @"previousLabelOrAppVersion": previousLabel
+                     };
```
lostintangent: Align curly braces

CodePushTelemetryManager.m
```diff
@@ -0,0 +1,125 @@
...
+        } else {
+            // Previous status report was with a binary app version.
+            return @{
+                     @"appVersion": appVersion,
+                     @"previousLabelOrAppVersion": previousStatusReportIdentifier
+                     };
```
lostintangent: Align curly braces

CodePushTelemetryManager.m
```diff
@@ -0,0 +1,125 @@
...
+{
+    NSString *currentPackageIdentifier = [self getPackageStatusReportIdentifier:currentPackage];
+    NSString *previousStatusReportIdentifier = [self getPreviousStatusReportIdentifier];
+    if (currentPackageIdentifier) {
+        if (previousStatusReportIdentifier == nil) {
+            [self recordDeploymentStatusReported:currentPackageIdentifier];
+            return @{
+                     @"package": currentPackage,
+                     @"status": DeploymentSucceeded
+                     };
```
lostintangent: Align this and the other two curly braces

lostintangent:
This baby looks good to go! I really really love how clean this is. Awesoke job @geof90!

CodePush.m
```diff
@@ -523,55 +493,42 @@ - (void)savePendingUpdate:(NSString *)packageHash
...
 
 /*
- * This method is checks if a new status update exists (new version was installed, 
+ * This method is checks if a new status update exists (new version was installed,
  * or an update failed) and return its details (version label, status).
  */
 RCT_EXPORT_METHOD(getNewStatusReport:(RCTPromiseResolveBlock)resolve
                             rejecter:(RCTPromiseRejectBlock)reject)
 {
     if (needToReportRollback) {
```
lostintangent: Should we wrap this entire method in a call to "dispatch_async" (just like "installUpdate") so that we don't synchronously tie up the native module thread?

PR #168, opened 2016-01-29
Android Asset Updates 🎉

geof90:


android/app/src/main/java/com/microsoft/codepush/react/CodePushUtils.java
```diff
@@ -265,6 +271,108 @@ public static String tryGetString(ReadableMap map, String key) {
...
+                if (zis != null) zis.close();
+                if (bis != null) bis.close();
+                if (fis != null) fis.close();
+            } catch (IOException e) {
+                throw new CodePushUnknownException("Error closing IO resources.", e);
+            }
+        }
+    }
+
+    public static void mergeEntriesInFolder(String fromPath, String destinationPath) throws IOException {
```
lostintangent: How about "copyDirectoryContents"? And the params could be "sourceDirectoryPath" and "targetDiectoryPath"?

android/app/src/main/java/com/microsoft/codepush/react/CodePushUtils.java
```diff
@@ -265,6 +271,108 @@ public static String tryGetString(ReadableMap map, String key) {
...
+        File fromDir = new File(fromPath);
```
lostintangent: Should we go with "source" everywhere instead of "from"? I think it reads a little better.

android/app/src/main/java/com/microsoft/codepush/react/CodePushUtils.java
```diff
@@ -265,6 +271,108 @@ public static String tryGetString(ReadableMap map, String key) {
         }
     }
 
+    public static void unzipFile(File zipFile, String destination) throws IOException {
+        FileInputStream fis = null;
+        BufferedInputStream bis = null;
```
lostintangent: I'd prefer if we didn't use acronym identifiers. How about "inputStream", "bufferStream" and "zipStream"? It obviously looks repetitive at the point of decleation but I think it makes the code easier to read when you're using the variables latter on. Also, do we need to initialize these to null or would the compiler do that for us?

android/app/src/main/java/com/microsoft/codepush/react/CodePushPackage.java
```diff
@@ -178,6 +214,76 @@ public void downloadPackage(Context applicationContext, ReadableMap updatePackag
...
+            // File is not a zip.
+            File updateBundleFile = new File(newPackageFolderPath, UPDATE_BUNDLE_FILE_NAME);
+            downloadFile.renameTo(updateBundleFile);
+        }
+
+        String bundlePath = CodePushUtils.appendPathComponent(newPackageFolderPath, PACKAGE_FILE_NAME);
+        CodePushUtils.writeReadableMapToFile(updatePackage, bundlePath);
+    }
+
+    public String findMainBundleInFolder(String folderPath) {
```
lostintangent: How about "findJSBundleInUpdateContents"? The word "Main" doesn't seem necessary and I like adding "UpdateContents" since it adds a little more semantic value to what this function is trying to achieve.

android/app/src/main/java/com/microsoft/codepush/react/CodePushPackage.java
```diff
@@ -87,7 +104,13 @@ public String getCurrentPackageBundlePath() {
             return null;
         }
 
-        return CodePushUtils.appendPathComponent(packageFolder, UPDATE_BUNDLE_FILE_NAME);
+        WritableMap currentPackage = getCurrentPackage();
```
lostintangent: I understand what you're doing here, but I might now in a month :) Could we add a simple comment explaining that the relative path is used for zipped updates vs single file updates?

android/app/src/main/java/com/microsoft/codepush/react/CodePushUtils.java
geof90: Point taken about the naming. And no, local variables are not auto-initialized to null in Java. It only does that for class variables.

lostintangent: OK cool, thanks for confirming :)

lostintangent:
BTW, this isn't related to this change, but I just noticed that the `getNewStatusReport` method wasn't made to run Async like our other potentially expensive JS-exported methods. Should we change that like we did for iOS?

android/app/src/main/java/com/microsoft/codepush/react/CodePushInvalidPackageException.java
```diff
@@ -0,0 +1,7 @@
+package com.microsoft.codepush.react;
+
+public class CodePushInvalidPackageException  extends RuntimeException {
```
lostintangent: What do you think about replacing "Package" with "Update"? I'd like to start eradicating the use of the word "Package" in our codebase :)

android/app/src/main/java/com/microsoft/codepush/react/CodePushPackage.java
geof90: Added the comment

geof90:
Yup agreed, I just pushed a change to wrap the getNewStatusReport in an AsyncTask.

android/app/src/main/java/com/microsoft/codepush/react/CodePushUpdateUtils.java
```diff
@@ -0,0 +1,46 @@
...
+
+import com.facebook.react.bridge.ReadableArray;
+import com.facebook.react.bridge.WritableMap;
+
+import java.io.File;
+import java.io.IOException;
+
+public class CodePushUpdateUtils {
+
+    public static void copyNecessaryFilesFromCurrentPackage(String diffManifestFilePath, String currentPackageFolderPath, String newPackageFolderPath) throws IOException{
```
lostintangent: What about "processDiffUpdate" or something similar? Since you moved this into an update-specific utils, I like the idea of having the method names be as specific to the scenario as possible.

lostintangent:
LGTM!

Issue #44, opened 2015-11-10
CocoaPods configuration


nevir:
It'd be handy to also package this as a CocoaPods project for even easier setup

lostintangent:
This is definitely something we need to do! Thanks for bringing it up @nevir. We'll likely address this in the next week or so, as we light up a few other enhancements.

nevir:
:+1:

lostintangent:
We're going to evaluate using RNPM for making the setup of our plugin easier for both platforms, since the CocoaPods solution would only help iOS. Reopen this issue if you think CocoaPods support specifically would be valuable. Thanks!

PR #169, opened 2016-01-29
CocoaPods support

oblador:
This PR adds a `.podspec` file so that CodePush may be integrated via CocoaPods. I noted it handled here https://github.com/Microsoft/react-native-code-push/issues/44 where you guys preferred `rnpm`, but can't see any downsides to having it as an option. 

I opted to add the `SSZipArchive` as source files instead of as a proper CocoaPods dependency to streamline how that dependency is handled and avoid future upgrade bugs.

### DIFF
--- a/CodePush.podspec
+++ b/CodePush.podspec
@@ -1,0 +1,16 @@
+Pod::Spec.new do |s|
+
+  s.name             = 'CodePush'
+  s.version          = '1.6.0-beta'
+  s.summary          = 'React Native plugin for the CodePush service'
+  s.author           = 'Microsoft Corporation'
+  s.license          = 'MIT'
+  s.homepage         = 'http://microsoft.github.io/code-push/'
+  s.source           = { :git => 'https://github.com/Microsoft/react-native-code-push.git', :tag => "v#{s.version}" }
+  s.platform         = :ios, '7.0'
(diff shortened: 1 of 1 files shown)


lostintangent:
Nice!! This is unbelievably helpful @oblador! I agree that it makes total sense to have CocoaPods support, especially since RNPM seems to be going in the direction of relying on CocoaPods/Carthage for certain things (e.g. adding custom linking behavior ala libz). This change looks great to me.

oblador:
Thanks for the quick reply/merge. Do you guys have a release schedule (referencing the git repo doesn't work since it depends on an unreleased version of `code-push`)?

lostintangent:
We're planning to release an update to NPM on Monday that introduces support for updating images on Android, and includes a few bugs fixes (and your PR) :) Does that schedule work OK for you?

oblador:
Perfect, thanks :-)

lostintangent:
Awesome! Thanks again for the contribution. I'll ping you as soon as we release on Monday.

lostintangent:
@oblador We're still doing some testing for our upcoming release, but we released the code-push dependency that master has, so you should be able to install from GitHub in the meantime.

oblador:
That works :+1:

PR #170, opened 2016-01-30
Another way to init CodePush runtime for RN Android >=0.18

geof90:
This updates the docs to prescribe initializing the `_codePush` instance variable inside `onCreate`, which seems to be a common practice for initializing instance variables in Android, since it is guaranteed to be only called once early in the Activity lifecycle. Just a suggestion. @lostintangent

### DIFF
--- a/README.md
+++ b/README.md
@@ -220,7 +220,7 @@ After installing the plugin and syncing your Android Studio project with Gradle,
         // Activity lifecycle event using the right deployment key
         @Override
         protected void onCreate(Bundle savedInstanceState) {
-            codePush = new CodePush("0dsIDongIcoH0mqAmoR0CYb5FhBZNy1w4Bf-l", this, BuildConfig.DEBUG);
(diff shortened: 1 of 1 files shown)


lostintangent:
LGTM! This is definitely much cleaner than my previous recommendation :)

geof90:
Some caveats:

The `this._codePush = new CodePush("0dsIDongIcoH0mqAmoR0CYb5FhBZNy1w4Bf-l", this, BuildConfig.DEBUG);` line MUST be placed above the call to `super.onCreate(savedInstanceState);`. This is because the `ReactActivity` implementation of `onCreate()` will also initialize the 3rd party packages, which will result in a NPE if `_codePush` is not intialized before the call to `getPackages()`.

Do you think we will run into issues with this hidden requirement? Should we document this inside as a comment?

lostintangent:
Yeah I think it's fine to document this with a comment. Let's merge this change since I think it's a better solution in the interim.

lostintangent:
Now that we've broken compat with previous RN versions, I actually think we should refactor our API to be more accommodating of the new Android activity solution, and not worry about taking this change.

PR #176, opened 2016-02-02
Disable metrics in debug

geof90:
Reverts https://github.com/Microsoft/react-native-code-push/pull/175 and disables the call to getNewStatusReport from JS instead.


lostintangent:
LGTM!

PR #179, opened 2016-02-03
Version bump

lostintangent:
We never bumped the Podspec version as part of the 1.7.0 release, and so I'm bumping our package version again so that we can release an update to NPM which includes the appropriate version in the Podspec as well.


geof90:
LGTM!

PR #181, opened 2016-02-04
Linking via RNPM

geof90:
Simplifies the linking process by using RNPM. Updated docs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,10 +6,8 @@ This plugin provides client-side integration for the [CodePush service](http://c
 * [Supported React Native Platforms](#supported-react-native-platforms)
 * [Getting Started](#getting-started)
 * [iOS Setup](#ios-setup)
-    * [Plugin Installation](#plugin-installation-ios)
     * [Plugin Configuration](#plugin-configuration-ios)
 * [Android Setup](#android-setup)
-    * [Plugin Installation](#plugin-installation-android)
     * [Plugin Configuration (React Native < v0.18.0)](#plugin-configuration-android---react-native--v0180)
     * [Plugin Configuration (React Native 0.18.0+)](#plugin-configuration-android---react-native-v0180)
 * [Plugin Usage](#plugin-usage)
@@ -60,7 +58,7 @@ rnpm info Linking react-native-code-push ios dependency
 rnpm info iOS module react-native-code-push has been successfully linked 
 ```
 
-After installing the plugin via NPM, as with all other React Native plugins, the integration experience is different for iOS and Android, so perform the following setup steps depending on which platform(s) you are targetting.
+Next, as with all other React Native plugins, the integration experience is different for iOS and Android, so perform the following setup steps depending on which platform(s) you are targetting.
 
 ## iOS Setup
 
--- a/README.md
+++ b/README.md
@@ -7,8 +7,6 @@ This plugin provides client-side integration for the [CodePush service](http://c
 * [Getting Started](#getting-started)
 * [iOS Setup](#ios-setup)
 * [Android Setup](#android-setup)
-    * [Plugin Configuration (React Native < v0.18.0)](#plugin-configuration-android---react-native--v0180)
-    * [Plugin Configuration (React Native 0.18.0+)](#plugin-configuration-android---react-native-v0180)
 * [Plugin Usage](#plugin-usage)
 * [Releasing Updates (JavaScript-only)](#releasing-updates-javascript-only)
 * [Releasing Updates (JavaScript + images)](#releasing-updates-javascript--images)
@@ -124,12 +122,6 @@ To let the CodePush runtime know which deployment it should query for updates ag
 ## Android Setup
 
 *NOTE: these instructions only apply to version 1.7.2-beta and above of the plugin. For setup instructions pertaining to older versions, please refer to [this version of the docs](https://github.com/Microsoft/react-native-code-push/blob/28c4f2dd03448c393445800a5e0d828b69a614e3/README.md).*
-
-In order to integrate CodePush into your Android project, perform the following steps:
-
-### Plugin Configuration (Android - React Native v0.18.0+)
-
-*NOTE: These instructions are specific to apps that are using React Native v0.18.0+. If you are using v0.15.0-v0.17.0, then refer to the previous section.*
 
 After installing the plugin and syncing your Android Studio project with Gradle, you need to configure your app to consult CodePush for the location of your JS bundle, since it will "take control" of managing the current and all future versions. To do this, perform the following steps:
 
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ Next, as with all other React Native plugins, the integration experience is diff
 
 ## iOS Setup
 
-*NOTE: these instructions only apply to version 1.7.2-beta and above of the plugin. For setup instructions pertaining to older versions, please refer to [this version of the docs](https://github.com/Microsoft/react-native-code-push/blob/28c4f2dd03448c393445800a5e0d828b69a614e3/README.md).*
+*NOTE: these instructions only apply to version 1.7.2-beta and above of the plugin. For setup instructions pertaining to older versions, please refer to [this version of the docs](https://github.com/Microsoft/react-native-code-push/blob/28c4f2dd03448c393445800a5e0d828b69a614e3/README.md#ios-setup).*
 
 Once you've acquired the CodePush plugin, you need to integrate it into the Xcode project of your React Native app.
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -6,10 +6,7 @@ This plugin provides client-side integration for the [CodePush service](http://c
 * [Supported React Native Platforms](#supported-react-native-platforms)
 * [Getting Started](#getting-started)
 * [iOS Setup](#ios-setup)
-    * [Plugin Installation](#plugin-installation-ios)
```
lostintangent: Do we need to keep the old docs around for a little bit? Or are we expecting that anyone reading the getting started docs will be using the latest plugin version?

package.json
```diff
@@ -17,9 +17,13 @@
   },
   "dependencies": {
     "code-push": "1.5.2-beta",
-    "semver": "^5.1.0"
+    "semver": "^5.1.0",
+    "rnpm": "1.4.1"
   },
-  "devDependencies": {
+  "peerDependencies": {
     "react-native": "0.19.0"
```
lostintangent: Should this be >=0.19.0? We don't want to artificially block future versions that might just work.

README.md
lostintangent: Maybe we can just have a note about older releases and link to the readme in our existing 1.7.1 tag in GitHub.

package.json
```diff
@@ -17,9 +17,13 @@
...
+  },
+  "scripts": { 
+    "postinstall": "cd ../../ && rnpm link react-native-code-push"
```
lostintangent: Is it really necessary to cd up like this? Did you get this solution from the RNPM team? It just kind of seems strange. I would have expected the RNPM script to walk up the CWD until it found an RN app or something similar.

README.md
```diff
@@ -52,38 +49,31 @@ Once you've followed the general-purpose ["getting started"](http://codepush.too
 npm install --save react-native-code-push
 ```
 
-As with all other React Native plugins, the integration experience is different for iOS and Android, so perform the following setup steps depending on which platform(s) you are targetting.
+The React Native plugin for CodePush is configured to be linked with your project automatically by [RNPM](https://github.com/rnpm/rnpm) upon installation via an NPM postinstall hook. To verify that the linking has succeeded, you should see the following lines in the terminal after installation:
```
lostintangent: How about "...the following output in your terminal..."?

README.md
```diff
@@ -185,7 +157,9 @@ After installing the plugin and syncing your Android Studio project with Gradle,
                     // folder with the name of the bundle passed in, e.g. index.android.bundle
                     .setJSBundleFile(codePush.getBundleUrl("index.android.bundle"))
                     
-                    // 6. Expose the CodePush module to JavaScript.
+                    // 6. DELETE THIS LINE --> .addPackage(new CodePushReactPackage())
```
lostintangent: The version of the CodePush plugin that uses RNPM isn't going to support <RN 0.19.0 versions, so we presumably don't even need this section anymore right?

README.md
```diff
@@ -238,9 +212,11 @@ After installing the plugin and syncing your Android Studio project with Gradle,
             // 4. Instantiate an instance of the CodePush runtime, using the right deployment key
             this._codePush = new CodePush("0dsIDongIcoH0mqAmoR0CYb5FhBZNy1w4Bf-l", this, BuildConfig.DEBUG);
```
lostintangent: Could we change this constructor to take a ReactActvitty instance instead of just an Activity? Then we could remove the third parameter and determine debug mode by simply calling the ReactActivitiy's method.

README.md
```diff
@@ -238,9 +212,11 @@ After installing the plugin and syncing your Android Studio project with Gradle,
...
 
-            // 5. Add the CodePush package to the list of existing packages
+            // 5. Add the CodePush package to the list of existing packages,
+            // replacing the `new CodePushReactPackage()` added by RNPM
             return Arrays.<ReactPackage>asList(
-                new MainReactPackage(), this._codePush.getReactPackage());
+                new MainReactPackage(), /* REMOVE THIS -> new CodePushReactPackage() */
```
lostintangent: Would it be possible to support an API where the package constructor took the deployment key and activity reference, and there was a static method for getting the JS bundle URL? It would be cool if instead of requiring the dev to remove the code added by RNPM and then adding another line, they just simply had to pass two parameters to the constructor that RNPM already injected for them.

README.md
```diff
@@ -52,38 +49,33 @@ Once you've followed the general-purpose ["getting started"](http://codepush.too
...
 
-### Plugin Installation (iOS)
+Next, as with all other React Native plugins, the integration experience is different for iOS and Android, so perform the following setup steps depending on which platform(s) you are targetting.
 
-1. Open your app's Xcode project
-2. Find the `CodePush.xcodeproj` file within the `node_modules/react-native-code-push` directory, and drag it into the `Libraries` node in Xcode
+## iOS Setup
 
-    ![Add CodePush to project](https://cloud.githubusercontent.com/assets/516559/10322414/7688748e-6c32-11e5-83c1-00d3e6758df4.png)
+*NOTE: these instructions only apply to version 1.7.2-beta and above of the plugin. For setup instructions pertaining to older versions, please refer to [this version of the docs](https://github.com/Microsoft/react-native-code-push/blob/28c4f2dd03448c393445800a5e0d828b69a614e3/README.md).*
```
lostintangent: Should we hyperlink directly to the iOS setup section in the readme to make this easier?

README.md
```diff
@@ -131,80 +123,9 @@ To let the CodePush runtime know which deployment it should query for updates ag
...
-                    // folder with the name of the bundle passed in, e.g. index.android.bundle
-                    .setJSBundleFile(codePush.getBundleUrl("index.android.bundle"))
-                    
-                    // 6. Expose the CodePush module to JavaScript.
-                    .addPackage(codePush.getReactPackage())
-                    ...
-        }
-    }
-    ```
+*NOTE: these instructions only apply to version 1.7.2-beta and above of the plugin. For setup instructions pertaining to older versions, please refer to [this version of the docs](https://github.com/Microsoft/react-native-code-push/blob/28c4f2dd03448c393445800a5e0d828b69a614e3/README.md).*
```
lostintangent: Same comment as above.

package.json
```diff
@@ -17,9 +17,13 @@
...
+    "semver": "^5.1.0",
+    "rnpm": "1.4.1"
   },
-  "devDependencies": {
-    "react-native": "0.19.0"
+  "peerDependencies": {
+    "react-native": "^0.19.0"
+  },
+  "scripts": { 
+    "postinstall": "cd ../../ && rnpm link react-native-code-push"
```
lostintangent: How does this behave if you re-install the plugin to acquire a newer version, but the host project has already linked to CodePush? Does it effectively no-op? Or would the dev have to perform some steps again?

README.md
```diff
@@ -131,80 +123,9 @@ To let the CodePush runtime know which deployment it should query for updates ag
...
-            ...
-            versionName "1.0.0"
-            ...
-        }
-        ...
-    }
-    ```
+In order to integrate CodePush into your Android project, perform the following steps:
 
 ### Plugin Configuration (Android - React Native v0.18.0+)
```
lostintangent: We probably don't need this header anymore once we make this readme specific to RN 0.19.0+.

README.md
```diff
@@ -6,10 +6,7 @@ This plugin provides client-side integration for the [CodePush service](http://c
 * [Supported React Native Platforms](#supported-react-native-platforms)
 * [Getting Started](#getting-started)
 * [iOS Setup](#ios-setup)
-    * [Plugin Installation](#plugin-installation-ios)
-    * [Plugin Configuration](#plugin-configuration-ios)
 * [Android Setup](#android-setup)
-    * [Plugin Installation](#plugin-installation-android)
     * [Plugin Configuration (React Native < v0.18.0)](#plugin-configuration-android---react-native--v0180)
```
lostintangent: You can remove this link since the section it points to has been removed.

package.json
geof90: It is definitely supposed to no-op and show you this message - 

```
rnpm info Linking react-native-code-push android dependency 
rnpm info Android module react-native-code-push is already linked 
rnpm info Linking react-native-code-push ios dependency 
rnpm info iOS module react-native-code-push is already linked 
```

That said, there's some weird issue with android where it is inserting the same lines in gradle twice.

geof90: Actually, if the original project was linked with RNPM too, everything works fine. If the original project was manually linked via our previous instructions before, the RNPM plugin gets confused.

lostintangent: What about if the project is using other custom modules? Would the first and/or subsequent links of CodePush screw up the build file?

lostintangent: I noticed that your RNPM PR just got merged, but I'm not sure if that would address the scenario I mentioned above. If a dev manually linked another RN module by appending the project name to an existing include statement in their build.gradle file, when they installed CodePush, it would then break their build file.

geof90: Yup, that was fixing the issue with detecting that a project has already been linked. I have another PR there now that fixes that issue with single line multi project import statements. https://github.com/rnpm/rnpm-plugin-link/pull/31

geof90: Update: That PR got merged too, just waiting for them to publish the latest version to NPM.

geof90: @lostintangent The RNPM guys have published those latest changes to NPM. Should we merge this? :D

geof90:
Closing due to change in RNPM design.

PR #184, opened 2016-02-04
Prevent crashes after downloading zip updates containing no valid jsbundle extension

geof90:


android/app/src/main/java/com/microsoft/codepush/react/CodePush.java
```diff
@@ -379,6 +379,10 @@ public void call(DownloadProgress downloadProgress) {
                     } catch (IOException e) {
                         e.printStackTrace();
                         promise.reject(e.getMessage());
+                    } catch (CodePushInvalidUpdateException e) {
+                        e.printStackTrace();
+                        saveFailedUpdate(updatePackage);
+                        promise.reject(e.getMessage());
```
lostintangent: What will this error message be? It would be great if we rejected this with a clear message ala "An update was downloaded, but ignored, because it didn't include a valid JS bundle file."

geof90: It currently says this: "Update is invalid - no files with extension .bundle, .js or .jsbundle were found in the update package."

geof90: Take a look at https://github.com/Microsoft/react-native-code-push/blob/master/android/app/src/main/java/com/microsoft/codepush/react/CodePushInvalidUpdateException.java

lostintangent: Ah OK cool.

lostintangent:
LGTM!

Issue #180, opened 2016-02-04
Icons disappearing


maxdjohnson:
Hi there, just deployed CodePush in our app but we're starting to notice missing icons on some devices. 

Comparing two phones running the same app store version and the same bundle downloaded from codepush, one shows all the icons while the other is missing a few (the icons on a TabBarIOS component).

The working phone is running ios 9.2, the other is running 9.3 beta 2.
The app is using code-push-react-native 1.6.0-beta and react-native 0.18.1

lostintangent:
Hmm interesting. Apologies for that! So the only devices that exhibit this issue are pre-release versions of iOS? On those devices, are you seeing all assets that you require via the RN assets system disappear after the CodePush update or just the tab bar icons?

We haven't done any testing with iOS 9.3 yet, so it's possible there is a breaking change in either RN or the CodePush plugin we need to respond to. We can take a look at this ASAP tomorrow if you could confirm that the issue appears to be limited to that iOS version.

Thanks so much for reporting this!

maxdjohnson:
No worries thanks for the quick reply!

It seems like it's only happening for the 9.3 prerelease and the TabBarIOS images. The odd thing is that the images were working on the prerelease phone before the bundle was updated via codepush. I'll do some more investigation tomorrow and try to reproduce it.

lostintangent:
That is really bizarre! I would expect the images to either work or not, so it's really intriguing to me that only the tab bar icons disappear after the CodePush update. Are you using custom icons or system icons? I just ask because I'll try to reproduce the issue on my end and I want to make sure I know the exact setup.

maxdjohnson:
We're using custom icons. Thanks for digging in!

lostintangent:
Hey @maxdjohnson, so I haven't been able to repro this issue unfortunately :( I upgraded my iPhone to 9.3 beta 2, and added a `TabBarIOS` component to a demo app, and I'm not seeing the icons disappear after doing a CodePush update. I tried doing an update that did and didn't modify the actual icon to see whether there was an issue with the diffing logic on the server, but it worked in both cases. Can you confirm a couple of things for me, just to make sure I'm testing the right condtions:

1. Are you referencing your `TabBarIOS.Item` icons via a `require` statement or are you relying on referencing them within an xcassets file/etc.? I'm currently testing the `require` strategy and it seems to be working fine.

    ```javascript 
    <TabBarIOS.Item title="Please work!" icon={require("./sample.png")}>
    ...
[...]

PR #186, opened 2016-02-05
Attach binary bundle date to package metadata on download

geof90:
Use the binary bundle date string attached to the package metadata for comparison in [CodePush bundleUrl] so as to avoid date synchronization issues (eg bundle date generated by the PC may be out of sync with device date). @lostintangent

### DIFF
--- a/CodePush.m
+++ b/CodePush.m
@@ -18,7 +18,6 @@ RCT_EXPORT_MODULE()
 static BOOL needToReportRollback = NO;
 static BOOL isRunningBinaryVersion = NO;
 static BOOL testConfigurationFlag = NO;
-static NSString *binaryBundleDate = nil;
 
 // These constants represent valid deployment statuses
 static NSString *const DeploymentFailed = @"DeploymentFailed";
@@ -39,26 +38,39 @@ static NSString *const PendingUpdateIsLoadingKey = @"isLoading";
 static NSString *const PackageHashKey = @"packageHash";
 static NSString *const PackageIsPendingKey = @"isPending";
 
+// These values are used to save the bundleURL and extension for the JS bundle
+// in the binary.
+static NSString *binaryJsName = @"main";
+static NSString *binaryJsExtension = @"jsbundle";
+
+
 #pragma mark - Public Obj-C API
+
++ (NSURL *)binaryJsBundleUrl
+{
+    return [[NSBundle mainBundle] URLForResource:binaryJsName withExtension:binaryJsExtension];
+}
 
 + (NSURL *)bundleURL
 {
-    return [self bundleURLForResource:@"main"];
+    return [self bundleURLForResource:binaryJsName];
 }
 
 + (NSURL *)bundleURLForResource:(NSString *)resourceName
 {
+    binaryJsName = resourceName;
     return [self bundleURLForResource:resourceName
-                        withExtension:@"jsbundle"];
+                        withExtension:binaryJsExtension];
 }
 
 + (NSURL *)bundleURLForResource:(NSString *)resourceName
                   withExtension:(NSString *)resourceExtension
 {
+    binaryJsName = resourceName;
+    binaryJsExtension = resourceExtension;
     NSError *error;
     NSString *packageFile = [CodePushPackage getCurrentPackageBundlePath:&error];
-    NSURL *binaryJsBundleUrl = [[NSBundle mainBundle] URLForResource:resourceName withExtension:resourceExtension];
-    [self setBinaryBundleDate:binaryJsBundleUrl];
+    NSURL *binaryJsBundleUrl = [self binaryJsBundleUrl];
     
     NSString *logMessageFormat = @"Loading JS bundle from %@";
     
@@ -79,7 +91,7 @@ static NSString *const PackageIsPendingKey = @"isPending";
     NSString *packageDate = [currentPackageMetadata objectForKey:BinaryBundleDateKey];
     NSString *packageAppVersion = [currentPackageMetadata objectForKey:@"appVersion"];
     
-    if ([binaryBundleDate isEqualToString:packageDate] && ([CodePush isUsingTestConfiguration] ||[binaryAppVersion isEqualToString:packageAppVersion])) {
+    if ([[self modifiedDateStringFromFileUrl:binaryJsBundleUrl] isEqualToString:packageDate] && ([CodePush isUsingTestConfiguration] ||[binaryAppVersion isEqualToString:packageAppVersion])) {
         // Return package file because it is newer than the app store binary's JS bundle
         NSURL *packageUrl = [[NSURL alloc] initFileURLWithPath:packageFile];
         NSLog(logMessageFormat, packageUrl);
@@ -112,13 +124,17 @@ static NSString *const PackageIsPendingKey = @"isPending";
 }
 
 /*
- * This caches the binary's jsbundle modified date in memory as a string.
+ * This returns the modified date as a string for a given file URL.
  */
-+ (void)setBinaryBundleDate:(NSURL *)binaryJsBundleUrl
++ (NSString *)modifiedDateStringFromFileUrl:(NSURL *)fileUrl
 {
-    NSDictionary *binaryFileAttributes = [[NSFileManager defaultManager] attributesOfItemAtPath:[binaryJsBundleUrl path] error:nil];
-    NSDate *binaryDate = [binaryFileAttributes objectForKey:NSFileModificationDate];
-    binaryBundleDate = [NSString stringWithFormat:@"%f", [binaryDate timeIntervalSince1970]];
+    if (fileUrl != nil) {
+        NSDictionary *fileAttributes = [[NSFileManager defaultManager] attributesOfItemAtPath:[fileUrl path] error:nil];
+        NSDate *modifiedDate = [fileAttributes objectForKey:NSFileModificationDate];
+        return [NSString stringWithFormat:@"%f", [modifiedDate timeIntervalSince1970]];
+    } else {
+        return nil;
+    }
 }
 
 + (void)setDeploymentKey:(NSString *)deploymentKey
@@ -377,9 +393,10 @@ RCT_EXPORT_METHOD(downloadUpdate:(NSDictionary*)updatePackage
                         rejecter:(RCTPromiseRejectBlock)reject)
 {
     dispatch_async(dispatch_get_main_queue(), ^{
-        NSDictionary* mutableUpdatePackage = [updatePackage mutableCopy];
-        if (binaryBundleDate != nil) {
-            [mutableUpdatePackage setValue:binaryBundleDate
+        NSDictionary *mutableUpdatePackage = [updatePackage mutableCopy];
+        NSURL *binaryJsBundleUrl = [CodePush binaryJsBundleUrl];
+        if (binaryJsBundleUrl != nil) {
+            [mutableUpdatePackage setValue:[CodePush modifiedDateStringFromFileUrl:binaryJsBundleUrl]
                                     forKey:BinaryBundleDateKey];
         }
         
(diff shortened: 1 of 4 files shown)


CodePush.m
```diff
@@ -362,11 +366,19 @@ - (void)savePendingUpdate:(NSString *)packageHash
 /*
  * This is native-side of the RemotePackage.download method
  */
-RCT_EXPORT_METHOD(downloadUpdate:(NSDictionary*)updatePackage
+RCT_EXPORT_METHOD(downloadUpdate:(NSDictionary*)immutableUpdatePackage
```
lostintangent: I'd rather leave this as `updatePackage` and change the name of the mutable one to `mutablePackage`.

CodePush.m
```diff
@@ -362,11 +366,19 @@ - (void)savePendingUpdate:(NSString *)packageHash
...
                         resolver:(RCTPromiseResolveBlock)resolve
                         rejecter:(RCTPromiseRejectBlock)reject)
 {
     dispatch_async(dispatch_get_main_queue(), ^{
+        NSDictionary* updatePackage = [immutableUpdatePackage mutableCopy];
+        NSUserDefaults *preferences = [NSUserDefaults standardUserDefaults];
```
lostintangent: Why are we pulling this from the defaults as opposed to just getting the current binary timestamp and storing that in the package?

CodePush.m
```diff
@@ -56,6 +57,11 @@ + (NSURL *)bundleURLForResource:(NSString *)resourceName
     NSError *error;
     NSString *packageFile = [CodePushPackage getCurrentPackageBundlePath:&error];
     NSURL *binaryJsBundleUrl = [[NSBundle mainBundle] URLForResource:resourceName withExtension:resourceExtension];
+    NSDictionary *binaryFileAttributes = [[NSFileManager defaultManager] attributesOfItemAtPath:[binaryJsBundleUrl path] error:nil];
```
lostintangent: Could we refactor this logic out into a utils method called `[CodePushUtils binaryBundleModifiedDateString]` or something? That way, this code and the download code can both use them. I'm assuming you're storing this date as a way to share this value between both methods?

CodePushPackage.m
```diff
@@ -285,6 +285,11 @@ + (void)downloadPackage:(NSDictionary *)updatePackage
                 [CodePushPackage copyEntriesInFolder:unzippedFolderPath
                                           destFolder:newPackageFolderPath
                                                error:&error];
+                if (error) {
```
lostintangent: Was this added for good measure or is there a scenario you thought of that this could cause problems with?

geof90: 1. This is the same behaviour as Android https://github.com/Microsoft/react-native-code-push/blob/master/android/app/src/main/java/com/microsoft/codepush/react/CodePush.java#L379: If the error is an IOException (which should not happen), we don't mark the update failed because the problem is with our plugin, not with the update. I don't want to overengineer the thing by adding retry logic. I think the simplest solution is to reject the sync promise, and get the error printed to the console so that users can report it. I think that if this is in place, we would at least get more information on https://github.com/Microsoft/react-native-code-push/issues/180

2. Good point, and totally agree, we should delete it before returning.

CodePush.m
geof90: The way the 
```
+ (NSURL *)bundleURLForResource:(NSString *)resourceName
                  withExtension:(NSString *)resourceExtension
```
is designed, it looks like the resource name and resource extension is variable and may change. Therefore the only place that knows how to find the binary timestamp is in that method, that's why I thought of saving the timestamp in the defaults inside that method so that other methods can retrieve it.

lostintangent: Ah right, that's a good point. I guess this is yet another example of why our current API has some odd consequences as opposed to having a more intuitive entry point for the plugin than `bundleURL`.

CodePushPackage.m
geof90: I pushed a change to always delete the folder at the destination before unzipping so that leftover folders will not cause issues.

CodePush.m
```diff
@@ -367,7 +387,14 @@ - (void)savePendingUpdate:(NSString *)packageHash
                         rejecter:(RCTPromiseRejectBlock)reject)
 {
     dispatch_async(dispatch_get_main_queue(), ^{
-        [CodePushPackage downloadPackage:updatePackage
+        NSDictionary* mutableUpdatePackage = [updatePackage mutableCopy];
+        NSString *binaryBundleDate = [CodePush getBinaryBundleDateString];
+        if (binaryBundleDate != nil) {
```
lostintangent: When would this be `nil`? Only in debug where a bundle wasn't generated?

CodePush.m
```diff
@@ -102,6 +102,17 @@ + (NSString *)getApplicationSupportDirectory
     return applicationSupportDirectory;
 }
 
++ (NSString *)getBinaryBundleDateString
```
lostintangent: Isn't it typically Objective-C convention to leave off the `get` prefix for accessor methods?

CodePush.m
```diff
@@ -56,6 +57,7 @@ + (NSURL *)bundleURLForResource:(NSString *)resourceName
     NSError *error;
     NSString *packageFile = [CodePushPackage getCurrentPackageBundlePath:&error];
     NSURL *binaryJsBundleUrl = [[NSBundle mainBundle] URLForResource:resourceName withExtension:resourceExtension];
+    [self saveBinaryBundleDate:binaryJsBundleUrl];
```
lostintangent: Does this really need to happen on every app launch? Maybe it would make more sense to use an ivar to carry this value? The write/synchronize to user defaults is causing disk I/O, so I'm just wondering if that is the best solution for this afterall.

CodePush.m
geof90: Yup!

CodePush.m
```diff
@@ -36,26 +36,41 @@ @implementation CodePush {
 // that is associated with an update's package.
 static NSString *const PackageHashKey = @"packageHash";
 static NSString *const PackageIsPendingKey = @"isPending";
+static NSString *const BinaryBundleDateKey = @"binaryDate";
+
+// These values are used to save the bundleURL and extension for the JS bundle
+// in the binary.
+static NSString *binaryJsName = @"main";
```
lostintangent: How about `bundleResourceName` and `bundleResourceExtension`? Those match the parameter names we use for `bundleURL`.

CodePush.m
```diff
@@ -36,26 +36,41 @@ @implementation CodePush {
 // that is associated with an update's package.
 static NSString *const PackageHashKey = @"packageHash";
 static NSString *const PackageIsPendingKey = @"isPending";
+static NSString *const BinaryBundleDateKey = @"binaryDate";
```
lostintangent: Sort

CodePush.m
```diff
@@ -36,26 +36,41 @@ @implementation CodePush {
...
+
+// These values are used to save the bundleURL and extension for the JS bundle
+// in the binary.
+static NSString *binaryJsName = @"main";
+static NSString *binaryJsExtension = @"jsbundle";
+
 
 #pragma mark - Public Obj-C API
 
++ (NSURL *)binaryJsBundleUrl
```
lostintangent: Why is this part of the public API section?

CodePush.m
```diff
@@ -36,26 +36,41 @@ @implementation CodePush {
...
 
 + (NSURL *)bundleURLForResource:(NSString *)resourceName
                   withExtension:(NSString *)resourceExtension
 {
+    binaryJsName = resourceName;
+    binaryJsExtension = resourceExtension;
     NSError *error;
     NSString *packageFile = [CodePushPackage getCurrentPackageBundlePath:&error];
-    NSURL *binaryJsBundleUrl = [[NSBundle mainBundle] URLForResource:resourceName withExtension:resourceExtension];
+    NSURL *binaryJsBundleUrl = [self binaryJsBundleUrl];
```
lostintangent: It feels a little weird to set these two variables and then use the `binaryJsBundleUrl` to then access them. I like the previous line better since it's clearer what is happening.

CodePush.m
lostintangent: Maybe call this `binaryBundleURL`? We should definitely at least capitalize the `URL` part since that is the Cocoa Touch convention.

CodePush.m
lostintangent: Hmm, maybe it's OK after all. I don't feel strongly about it :)

CodePush.m
```diff
@@ -34,52 +34,59 @@ @implementation CodePush {
...
 // These keys are used to inspect/augment the metadata
 // that is associated with an update's package.
+static NSString *const BinaryBundleDateKey = @"binaryDate";
 static NSString *const PackageHashKey = @"packageHash";
 static NSString *const PackageIsPendingKey = @"isPending";
 
+// These values are used to save the bundleURL and extension for the JS bundle
+// in the binary.
+static NSString *bundleResourceName = @"main";
+static NSString *bundleResourceExtension = @"jsbundle";
```
lostintangent: Sort these two and remove the extra newline below.

CodePush.m
```diff
@@ -34,52 +34,58 @@ @implementation CodePush {
 
 // These keys are used to inspect/augment the metadata
 // that is associated with an update's package.
+static NSString *const BinaryBundleDateKey = @"binaryDate";
 static NSString *const PackageHashKey = @"packageHash";
 static NSString *const PackageIsPendingKey = @"isPending";
 
+// These values are used to save the bundleURL and extension for the JS bundle
+// in the binary.
+static NSString *bundleResourceExtension = @"jsbundle";
```
lostintangent: I know this is unrelated to your PR, but could you create a new pragma section called "Static variables" that includes these two and the three above? I just noticed that I grouped all of these into a "Private constants" section, which isn't accurate.

CodePush.m
```diff
@@ -111,6 +117,20 @@ + (BOOL)isUsingTestConfiguration
     return testConfigurationFlag;
 }
 
+/*
+ * This returns the modified date as a string for a given file URL.
+ */
++ (NSString *)modifiedDateStringFromFileUrl:(NSURL *)fileUrl
```
lostintangent: How about `modifiedDateOfFileAtURL`? We should definitely capitalize the `URL` part (in the parameter as well), and I feel like the "Of[X]At[Y]" format is more idiomatic of Cococa Touch APIs (e.g. `attributesOfItemAtPath`).

CodePush.m
```diff
@@ -111,6 +119,20 @@ + (BOOL)isUsingTestConfiguration
     return testConfigurationFlag;
 }
 
+/*
+ * This returns the modified date as a string for a given file URL.
+ */
++ (NSString *)modifiedDateStringOfFileAtUrl:(NSURL *)fileUrl
```
lostintangent: This should be `*URL` (including the `fileURL` parameter name. Additionally, this should be in the private API section right?

CodePushPackage.m
```diff
@@ -228,8 +228,16 @@ + (void)downloadPackage:(NSDictionary *)updatePackage
             NSString * unzippedFolderPath = [CodePushPackage getUnzippedFolderPath];
             NSMutableDictionary * mutableUpdatePackage = [updatePackage mutableCopy];
             if (isZip) {
-                NSError *nonFailingError = nil;
+                if ([[NSFileManager defaultManager] fileExistsAtPath:unzippedFolderPath]) {
```
lostintangent: Can you add a brief comment here explaining why you're doing this? It would be helpful to remember that this is for catching the scenario were a previous download failed. BTW, could you also confirm that this folder would get removed in the future if a new binary was discovered, as opposed to waiting for a new CodePush update?

geof90: Added the comment. Also [CodePushPackage clearUpdates] deletes the entire CodePush folder which contains everything that we use including this unzip folder destination path.

lostintangent:
LGTM!

PR #187, opened 2016-02-06
Sync Twice issue

geof90:
Fixes https://github.com/Microsoft/react-native-code-push/issues/110

### DIFF
--- a/CodePush.js
+++ b/CodePush.js
@@ -191,7 +191,7 @@ const sync = (() => {
   
   return (options = {}, syncStatusChangeCallback, downloadProgressCallback) => {
     if (syncInProgress) {
-      typeof syncStatusChangeCallback == "function"
+      typeof syncStatusChangeCallback === "function"
         ? syncStatusChangeCallback(CodePush.SyncStatus.SYNC_IN_PROGRESS)
         : log("Sync already in progress.");
       return Promise.resolve(CodePush.SyncStatus.SYNC_IN_PROGRESS);
@@ -228,7 +228,7 @@ async function syncInternal(options = {}, syncStatusChangeCallback, downloadProg
     ...options 
   };
   
-  syncStatusChangeCallback = typeof syncStatusChangeCallback == "function"
+  syncStatusChangeCallback = typeof syncStatusChangeCallback === "function"
     ? syncStatusChangeCallback
     : (syncStatus) => {
         switch(syncStatus) {
@@ -267,7 +267,7 @@ async function syncInternal(options = {}, syncStatusChangeCallback, downloadProg
         }
       };
     
-  downloadProgressCallback = typeof downloadProgressCallback == "function" 
+  downloadProgressCallback = typeof downloadProgressCallback === "function" 
     ? downloadProgressCallback 
     : (downloadProgress) => {
         log(`Expecting ${downloadProgress.totalBytes} bytes, received ${downloadProgress.receivedBytes} bytes.`);
--- a/CodePush.js
+++ b/CodePush.js
@@ -139,7 +139,20 @@ function log(message) {
   console.log(`[CodePush] ${message}`)
 }
 
-async function notifyApplicationReady() {
+// This ensures that notifyApplicationReadyInternal is only called once 
+// in the lifetime of this module instance.
+const notifyApplicationReady = (() => {
+  let notifyApplicationReadyPromise;
+  return () => {
+    if (!notifyApplicationReadyPromise) {
+      notifyApplicationReadyPromise = notifyApplicationReadyInternal();
+    }
+    
+    return notifyApplicationReadyPromise;
+  };
+})();
+
+async function notifyApplicationReadyInternal() {
   await NativeCodePush.notifyApplicationReady();  
   const statusReport = await NativeCodePush.getNewStatusReport();
   if (statusReport) {
@@ -170,15 +183,26 @@ function setUpTestDependencies(testSdk, providedTestConfig, testNativeBridge) {
   if (testNativeBridge) NativeCodePush = testNativeBridge;
 }
 
-// This function allows calls to syncInternal to be chained on to each other so that they do not
-// interleave in the event that sync() is called multiple times.
+// This function allows only one syncInternal operation to proceed at any given time.
+// Parallel calls to sync() while one is ongoing yields CodePush.SyncStatus.SYNC_IN_PROGRESS.
 const sync = (() => {
-  let syncPromiseChain = Promise.resolve();
+  let syncInProgress = false;
+  const setSyncCompleted = () => { syncInProgress = false; };
+  
   return (options = {}, syncStatusChangeCallback, downloadProgressCallback) => {
-    syncPromiseChain = syncPromiseChain
-      .catch(() => {})
-      .then(() => syncInternal(options, syncStatusChangeCallback, downloadProgressCallback));
-    return syncPromiseChain;
+    if (syncInProgress) {
+      syncStatusChangeCallback(CodePush.SyncStatus.SYNC_IN_PROGRESS);
+      return Promise.resolve(CodePush.SyncStatus.SYNC_IN_PROGRESS);
+    } 
+    
+    syncInProgress = true;
+    const syncInternalPromise = syncInternal(options, syncStatusChangeCallback, downloadProgressCallback);
+    syncInternalPromise
+      .then(setSyncCompleted)
+      .catch(setSyncCompleted)
+      .done();
+
+    return syncInternalPromise;
   };
 })();
 
@@ -350,6 +374,7 @@ const CodePush = {
     UP_TO_DATE: 4, // The running app is up-to-date
     UPDATE_IGNORED: 5, // The app had an optional update and the end-user chose to ignore it
     UPDATE_INSTALLED: 6, // The app had an optional/mandatory update that was successfully downloaded and is about to be installed.
+    SYNC_IN_PROGRESS: 7, // There is an ongoing "sync" operation in progress.
     UNKNOWN_ERROR: -1
   },
   DEFAULT_UPDATE_DIALOG: {
--- a/CodePush.js
+++ b/CodePush.js
@@ -170,8 +170,20 @@ function setUpTestDependencies(testSdk, providedTestConfig, testNativeBridge) {
   if (testNativeBridge) NativeCodePush = testNativeBridge;
 }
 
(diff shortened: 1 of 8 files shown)


CodePush.js
```diff
@@ -179,7 +191,7 @@ function setUpTestDependencies(testSdk, providedTestConfig, testNativeBridge) {
  * releases, and displaying a standard confirmation UI to the end-user
  * when an update is available.
  */
-async function sync(options = {}, syncStatusChangeCallback, downloadProgressCallback) {  
+async function syncOperation(options = {}, syncStatusChangeCallback, downloadProgressCallback) {  
   const syncOptions = {
```
lostintangent: Maybe call this "syncInternal"?

CodePushPackage.m
```diff
@@ -458,19 +459,32 @@ + (void)installPackage:(NSDictionary *)updatePackage
...
-    NSString *previousPackageHash = [self getPreviousPackageHash:error];
-    if (!*error && previousPackageHash && ![previousPackageHash isEqualToString:packageHash]) {
-        NSString *previousPackageFolderPath = [self getPackageFolderPath:previousPackageHash];
-        // Error in deleting old package will not cause the entire operation to fail.
-        NSError *deleteError;
-        [[NSFileManager defaultManager] removeItemAtPath:previousPackageFolderPath
-                                                   error:&deleteError];
-        if (deleteError) {
-            NSLog(@"Error deleting old package: %@", deleteError);
+    if (!removePendingUpdate) {
```
lostintangent: I'm confused by this because it looks like the "removePendingUpdate" argument is set to true if there is a pending update, but you're only deleting the current update if there isn't a pending update. Does this need to be changed to "if (removePendingUpdate)" or am I reading it wrong?

CodePush.m
```diff
@@ -505,7 +506,12 @@ - (void)savePendingUpdate:(NSString *)packageHash
 RCT_EXPORT_METHOD(notifyApplicationReady:(RCTPromiseResolveBlock)resolve
                                 rejecter:(RCTPromiseRejectBlock)reject)
 {
-    [CodePush removePendingUpdate];
+    // We only mark a pending update as succeeded only if it has been booted up
+    // successfully, during which `didUpdate` is set to true.
```
lostintangent: You used the Java variable name :) "didUpdate" isn't defined here.

CodePush.js
```diff
@@ -170,8 +170,20 @@ function setUpTestDependencies(testSdk, providedTestConfig, testNativeBridge) {
   if (testNativeBridge) NativeCodePush = testNativeBridge;
 }
 
+// This function allows calls to syncInternal to be chained on to each other so that they do not
+// interleave in the event that sync() is called multiple times.
+const sync = (() => {
```
lostintangent: I'm still not sure how I feel about this. I think the fix you made to notifyApplicationReady and the change you made to remove a pending update when installing a new update assessed two core issues, but I'm still not convinced that we should allow queuing subsequent calls to sync as opposed to rejecting the second with a new status (e.g. SYNC_IN_PROGRESS). If a sync call was setup to display a UI when it was checking or finished checking for the update, it wouldn't add any value to run both, since that would be a poor user experience. Since we don't know when sync will be called and how it will be presented, I think it's a much more safe and simple solution to just not allow sync to be called while another call is in progress.

Can you think of a single reason why calling it twice in parallel would actually be valuable? Calling sync twice within a single user session makes total sense, which is why we need the logic to clear a pending update when downloading, but I don't see the value in allowing calling sync concurrently, especially since the side effects would be unknown.

CodePush.m
```diff
@@ -505,7 +506,12 @@ - (void)savePendingUpdate:(NSString *)packageHash
 RCT_EXPORT_METHOD(notifyApplicationReady:(RCTPromiseResolveBlock)resolve
                                 rejecter:(RCTPromiseRejectBlock)reject)
 {
-    [CodePush removePendingUpdate];
+    // We only mark a pending update as succeeded only if it has been booted up
+    // successfully, during which `_isFirstRunAfterUpdate` is set to true.
+    if (_isFirstRunAfterUpdate) {
```
lostintangent: How would this behave in the following scenario:

1. App is restarted to apply a new update and isFirsrRun is set to run
2. App calls sync and installs a new update that will be applied on next restart
3. App calls sync again

Wouldn't the second sync call remove the pending update from #2 because #1 set isFirstRun to true? This may be a fairly rare scenario, but it seems like one that we should accommodate.

geof90: Do you mean the third sync call?

CodePush.js
geof90: I definitely don't see any point in allowing parallel requests. I also think rejecting the sync is better. The metaphor I'm thinking of is when an email client is syncing with the email server, when you do a pull-to-refresh in the app for example, it would block you from doing the pull-to-refresh again until the existing one is completed.

lostintangent: I like that metaphor / rationale :) +1

lostintangent: I'm cool with SYNC_IN_PROGRESS. I can't think of anything much better :)

CodePush.m
geof90: Actually, maybe should we just on the JS side noOp notifyApplicationReady after the first time it is called? I feel like we are complicating the native notifyApplicationReady logic just to accomodate a scenarion that doesn't make sense to allow happen anyway

geof90: That way I can completely undo this change. Plus it could simplify the reportStatus logic too.

lostintangent: Yep that's much simpler and better :+1:

geof90: It would work under the assumption that nobody will `require` the same `react-native-code-push` module twice in the same app, thereby creating two instances. Doing so would defeat the work we are doing to prevent parallel `sync` requests as well as the work to noOp notifyApplicationReady after the first time. Seems like there's no singleton pattern in JS. But I think it is a valid assumption.

lostintangent: Are you sure that would create two instances? JS modules are supposed to be singletons already, so this pattern should work just fine. It would be good to verify that Babel / RN abide by this behavior, but I'm pretty sure they do.

CodePush.js
```diff
@@ -170,8 +183,31 @@ function setUpTestDependencies(testSdk, providedTestConfig, testNativeBridge) {
...
+  
+  return (options = {}, syncStatusChangeCallback, downloadProgressCallback) => {
+    if (syncInProgress) {
+      syncStatusChangeCallback(CodePush.SyncStatus.SYNC_IN_PROGRESS);
+      return Promise.resolve(CodePush.SyncStatus.SYNC_IN_PROGRESS);
+    } 
+    
+    syncInProgress = true;
+    const syncPromise = syncInternal(options, syncStatusChangeCallback, downloadProgressCallback);
+    syncPromise
```
lostintangent: Don't you need to assign the result of this expression back to "syncPromise" before returning it?

CodePush.js
```diff
@@ -170,8 +183,31 @@ function setUpTestDependencies(testSdk, providedTestConfig, testNativeBridge) {
...
+      .then(setSyncCompleted)
+      .catch(setSyncCompleted)
+      .done();
```
lostintangent: Doesn't calling done here end the Promise chain so that devs couldn't call them on it?

CodePush.js
```diff
@@ -170,8 +183,31 @@ function setUpTestDependencies(testSdk, providedTestConfig, testNativeBridge) {
   if (testNativeBridge) NativeCodePush = testNativeBridge;
 }
 
+// This function allows only one syncInternal operation to proceed at any given time.
+// Parallel calls to sync() while one is ongoing yields CodePush.SyncStatus.SYNC_IN_PROGRESS.
+const sync = (() => {
+  let syncInProgress = false;
+  const setSyncCompleted = () => { syncInProgress = false; };
+  
+  return (options = {}, syncStatusChangeCallback, downloadProgressCallback) => {
```
lostintangent: Could we do "return (...args) {" and then do "syncInternal(...args)" below? That way we don't have to define the signature/default values twice.

CodePush.js
```diff
@@ -170,8 +183,31 @@ function setUpTestDependencies(testSdk, providedTestConfig, testNativeBridge) {
...
+    if (syncInProgress) {
+      syncStatusChangeCallback(CodePush.SyncStatus.SYNC_IN_PROGRESS);
```
lostintangent: Don't you need to check for null here since the callback is optional? Additionally, can we add logging here to indicate the state?

CodePush.js
lostintangent: Actually, never mind, you need access to the sync callback, so what you have is probably easier.

CodePush.js
geof90: Yes, thats why i didn't do the assignment as per your previous comment. The one im returning is the un-doned promise

CodePush.js
lostintangent: Never mind, I see what you're doing now :)

CodePush.js
lostintangent: Ah right, got it. Thanks for clarifying that!

CodePush.js
```diff
@@ -170,8 +183,33 @@ function setUpTestDependencies(testSdk, providedTestConfig, testNativeBridge) {
...
+// Parallel calls to sync() while one is ongoing yields CodePush.SyncStatus.SYNC_IN_PROGRESS.
+const sync = (() => {
+  let syncInProgress = false;
+  const setSyncCompleted = () => { syncInProgress = false; };
+  
+  return (options = {}, syncStatusChangeCallback, downloadProgressCallback) => {
+    if (syncInProgress) {
+      typeof syncStatusChangeCallback == "function"
+        ? syncStatusChangeCallback(CodePush.SyncStatus.SYNC_IN_PROGRESS)
+        : log("Sync already in progress.");
```
lostintangent: I'd prefer always logging this, instead of only when there isn't a sync callback. That way, basic diagnostics could still be done via the console as opposed to having to set break points or expect the dev to add their own logging.

lostintangent:
LGTM!

lostintangent:
Can you also update the docs to include the new sync status and mention

lostintangent:
Sorry! I accidentally fat fingered the close button on my phone :/

lostintangent:
Can you also update the docs in two places:

1. Add the new sync status to the list of options for the enum
2. Add the new sync status as a possible resolved value for the promise returned by sync

lostintangent:
Awesome! Let' merge this baby :)

PR #189, opened 2016-02-09
Fix failing tests

geof90:
Tiny PR, fixes some issues caused by the latest RN dependency version of Babel not being happy with the use of "export default" and the "package" keyword.


lostintangent:
LGTM!

PR #191, opened 2016-02-09
Removing the NDK section from our build.gradle file

lostintangent:
Our plugin doesn't include any C++ code so we shouldn't need to specify any NDK config of our own. We had a user report issues with integrating CodePush into their NDK app, and it was because of the superflous `android.defaultConfig.ndk` section in our `build.grade` file. This fixes that issue by simply removing it entirely, since we don't actually need it.


basketofsoftkittens:
+1

geof90:
LGTM!

Issue #195, opened 2016-02-11
CodePush.getCurrentPackage() not found.


livioso:
Hello everybody. 

I am encountering the following problem when I run my application on the device *(simulator works fine)*.  I am using **React Native 0.19** and **react-native-code-push 1.7.2-beta**. Not sure but I think I had the same error before *(with React Native 0.18 and react-native-code-push 1.6.0-beta).*

```
2016-02-11 16:37:40.259 Fashwell[778:256664] Loading JS bundle from file:///var/mobile/Containers/Bundle/Application/C840798E-87AA-4700-BC2A-1A7A89453BC7/Fashwell.app/main.jsbundle
2016-02-11 16:37:40.676 [warn][tid:com.facebook.React.JavaScript] Callback with id 14: CodePush.getCurrentPackage() not found
2016-02-11 16:37:40.683 [error][tid:com.facebook.React.RCTExceptionsManagerQueue] Unhandled JS Exception: Callback with id 14: CodePush.getCurrentPackage() not found
2016-02-11 16:37:40.683 Fashwell[778:256712] *** Terminating app due to uncaught exception 'RCTFatalException: Unhandled JS Exception: Callback with id 14: CodePush.getCurrentPackage() not found', reason: 'Unhandled JS Exception: Callback with id 14: CodePush.getCurrentPackage() n..., stack:
value@1:28822
<unknown>@1:26937
_@1:25866
value@1:26907
'
*** First throw call stack:
(0x183999900 0x183007f80 0x183999848 0x10010902c 0x100103a1c 0x18399f430 0x18389ceb4 0x1838a17e0 0x100107180 0x100127010 0x100126a00 0x1003f1bf0 0x1003f1bb0 0x1003fe6c8 0x1003f58a0 0x1003f1bb0 0x100400e10 0x1004004d8 0x183601470 0x183601020)
libc++abi.dylib: terminating with uncaught exception of type NSException
(lldb) 
```

In my main component's `componentWillMount()` I simply do the following:
```
codePush.sync();
```

#### Update

1. Apparently, I get the same error in the simulator (after / while closing the app).

2. When compiled with Debug I get a slightly more verbose output:

```
2016-02-11 17:13:24.235 [trace][tid:com.facebook.React.JavaScript] [CodePush] Checking for update.
2016-02-11 17:13:24.261 [trace][tid:com.facebook.React.JavaScript] [CodePush] An unknown error occurred.
[...]

Issue #182, opened 2016-02-04
Updating new release causes crash


oney:
First, it never happened in my test. Today I released new update via CodePush, in the Crashlytics, about 200 users (25%) encountered the crash. In last email sent by @lostintangent to me, he said that he got two update prompts in a row, and then the app worked fine, but after he restarted it, it seemed to keep crashing. I guess this crash is related to it. Unfortunately, I can't reproduce it in my test.
The crash log is
```
Thread : Fatal Exception: RCTFatalException: Unhandled JS Exception: Invariant Violation: Callback with id 22: CodePush.getCurrentPackage() not found
0  CoreFoundation                 0x18369d900 __exceptionPreprocess
1  libobjc.A.dylib                0x182d0bf80 objc_exception_throw
2  CoreFoundation                 0x18369d848 -[NSException initWithCoder:]
3  GoodnightReact                 0x100137860 RCTFatal
4  GoodnightReact                 0x1001326c0 -[RCTExceptionsManager reportFatalException:stack:exceptionId:]
5  CoreFoundation                 0x1836a3430 __invoking___
6  CoreFoundation                 0x1835a0eb4 -[NSInvocation invoke]
7  CoreFoundation                 0x1835a57e0 -[NSInvocation invokeWithTarget:]
8  GoodnightReact                 0x100135f58 -[RCTModuleMethod invokeWithBridge:module:arguments:]
9  GoodnightReact                 0x100155724 -[RCTBatchedBridge _handleRequestNumber:moduleID:methodID:params:]
10 GoodnightReact                 0x100155114 __33-[RCTBatchedBridge handleBuffer:]_block_invoke388
11 libdispatch.dylib              0x1830f1630 _dispatch_call_block_and_release
12 libdispatch.dylib              0x1830f15f0 _dispatch_client_callout
13 libdispatch.dylib              0x1830fd634 _dispatch_queue_drain
14 libdispatch.dylib              0x1830f50f4 _dispatch_queue_invoke
15 libdispatch.dylib              0x1830f15f0 _dispatch_client_callout
16 libdispatch.dylib              0x1830ffa88 _dispatch_root_queue_drain
17 libdispatch.dylib              0x1830ff224 _dispatch_worker_thread3
18 libsystem_pthread.dylib        0x183305470 _pthread_wqthread
19 libsystem_pthread.dylib        0x183305020 start_wqthread
```

[...]

PR #196, opened 2016-02-12
Fix getCurrentPackage

geof90:
When the native `getCurrentPackage` API rejects the promise, it does not return from the function, resulting in the `Callback with id xx: CodePush.getCurrentPackage() not found` errors reported in https://github.com/Microsoft/react-native-code-push/issues/195 and https://github.com/Microsoft/react-native-code-push/issues/182.

More investigation needs to be done to ascertain why the promise gets rejected in the first place, e.g. why the `app.json` file could go missing with the rest of the files in the update folder still remains intact.

This PR also modifies the `downloadUpdate` native APIs to clear away any existing files in a new update folder in order to prevent any chance of an update being corrupted. Also, on iOS, this changes `downloadUpdate` to use the `NSFileManager`'s `copyItemAtPath` method instead of the custom copy method when copying from an old update to a new one in the case of a diff update, with the assumption that the stock Objective-C API is more reliable.

### DIFF
--- a/CodePushPackage.m
+++ b/CodePushPackage.m
@@ -142,7 +142,7 @@ NSString * const UnzippedFolderName = @"unzipped";
     NSString *folderPath = [CodePushPackage getCurrentPackageFolderPath:error];
     if (!*error) {
         if (!folderPath) {
-            return [NSDictionary dictionary];
+            return nil;
         }
         
         NSString *packagePath = [folderPath stringByAppendingPathComponent:@"app.json"];
@@ -159,7 +159,7 @@ NSString * const UnzippedFolderName = @"unzipped";
         }
     }
     
-    return NULL;
+    return nil;
 }
 
 + (NSDictionary *)getPackage:(NSString *)packageHash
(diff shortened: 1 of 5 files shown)


android/app/src/main/java/com/microsoft/codepush/react/CodePush.java
```diff
@@ -411,6 +411,10 @@ public void getCurrentPackage(final Promise promise) {
                 @Override
                 protected Void doInBackground(Object... params) {
                     WritableMap currentPackage = codePushPackage.getCurrentPackage();
+                    if (currentPackage == null) {
+                        promise.resolve("");
+                        return null;
```
lostintangent: Does this need to return null when the method has a void return type?

geof90: Its Void, not void. So yes

CodePushPackage.m
```diff
@@ -259,9 +266,9 @@ + (void)downloadPackage:(NSDictionary *)updatePackage
                         return;
                     }
                     
-                    [CodePushPackage copyEntriesInFolder:currentPackageFolderPath
-                                              destFolder:newPackageFolderPath
-                                                   error:&error];
+                    [[NSFileManager defaultManager] copyItemAtPath:currentPackageFolderPath
+                                                            toPath:newPackageFolderPath
```
lostintangent: Does this create the directory if it doesn't exist?

CodePushPackage.m
```diff
@@ -204,15 +204,22 @@ + (void)downloadPackage:(NSDictionary *)updatePackage
            failCallback:(void (^)(NSError *err))failCallback
 {
     NSString *newPackageFolderPath = [self getPackageFolderPath:updatePackage[@"packageHash"]];
-    NSError *error = nil;
+    NSError *error;
     
-    if (![[NSFileManager defaultManager] fileExistsAtPath:newPackageFolderPath]) {
-        [[NSFileManager defaultManager] createDirectoryAtPath:newPackageFolderPath
+    if (![[NSFileManager defaultManager] fileExistsAtPath:[self getCodePushPath]]) {
```
lostintangent: Why this change?

CodePushPackage.m
geof90: It creates the destination folder if it doesnt exist, and fails if it already does

CodePushPackage.m
geof90: Previously, we depended on the fact that we create the newPackageFolderPath if it doesnt exist to create the intermediate directories. Now, we only delete it if it exists, so we need this addition to create the intermediate directories, otherwise, the download handler (NSURLConnection) will fail

lostintangent: Would it make sense to make this an `else if` after the following clause?

android/app/src/main/java/com/microsoft/codepush/react/CodePush.java
lostintangent: Ah OK, interesting. Another Java thing I didn't know about :)

CodePushPackage.m
geof90: Well, yes. Though the expectation is that the `else if` only ever saves 1 file existence check during the first time this is called. Once the CodePush path is created, it is never removed unless a new binary is deployed.

PR #197, opened 2016-02-12
Version bump and doc updates

lostintangent:
This bumps the version so that we can release a couple of key bug fixes, and also updates the docs for `getCurrentPackage` to reflect the fact that it can now return `null`.


geof90:
LGTM!

PR #198, opened 2016-02-12
Fix single JS bundle release

geof90:
https://github.com/Microsoft/react-native-code-push/pull/196 regressed the single JSbundle update scenario because the `newPackageFolderPath` is not created.

### DIFF
--- a/CodePushPackage.m
+++ b/CodePushPackage.m
@@ -348,14 +348,10 @@ NSString * const UnzippedFolderName = @"unzipped";
                     return;
                 }
             } else {
-                if ([[NSFileManager defaultManager] fileExistsAtPath:bundleFilePath]) {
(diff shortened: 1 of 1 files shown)


CodePushPackage.m
```diff
@@ -348,14 +348,10 @@ + (void)downloadPackage:(NSDictionary *)updatePackage
...
             } else {
-                if ([[NSFileManager defaultManager] fileExistsAtPath:bundleFilePath]) {
-                    [[NSFileManager defaultManager] removeItemAtPath:bundleFilePath error:&error];
-                    if (error) {
-                        failCallback(error);
-                        return;
-                    }
-                }
-                
+                [[NSFileManager defaultManager] createDirectoryAtPath:newPackageFolderPath
```
lostintangent: Don't we want to only do this if it doesn't already exist?

lostintangent:
LGTM!

PR #202, opened 2016-02-17
Updating docs

lostintangent:
This PR does two things:

1. Updates our RN support matrix to include 0.20.0 and also mentions the disclaimer about future releases.
2. Adds a new section enumerating OSS projects that use CodePush, that way, we can give them credit for being awesome, and also point new users at them as examples.


shishirx34:
LGTM

PR #205, opened 2016-02-19
Verify hash after download

geof90:
As suggested in https://github.com/Microsoft/react-native-code-push/issues/180, This change configures the plugin to calculate and verify the hash of a downloaded update after it is unzipped. This is especially useful in a diff update so that the client can verify that the merged contents from the diff update and the previous update is the same as what the server would send in a full update.

Also, some other changes:

1. Moved all the iOS files out of the project root into an iOS folder.
2. Created a `CodePushUtils.m` file to contain some of the grunt logic originally performed in `CodePushPackage.m` so as to avoid code clutter.


iOS/CodePushPackage.m
```diff
@@ -288,18 +290,24 @@ + (void)downloadPackage:(NSDictionary *)updatePackage
...
-                            NSLog(@"Error deleting file from current package: %@", nonFailingError);
-                            nonFailingError = nil;
+                                                                   error:&error];
+                        if (error) {
+                            failCallback(error);
+                            return;
                         }
                     }
+                    
+                    [[NSFileManager defaultManager] removeItemAtPath:diffManifestFilePath
```
lostintangent: Is this important now because of the hash validation?

iOS/CodePushPackage.m
```diff
@@ -288,18 +290,24 @@ + (void)downloadPackage:(NSDictionary *)updatePackage
...
                     for (NSString *deletedFileName in deletedFiles) {
                         [[NSFileManager defaultManager] removeItemAtPath:[newPackageFolderPath stringByAppendingPathComponent:deletedFileName]
-                                                                   error:&nonFailingError];
-                        
-                        if (nonFailingError) {
-                            NSLog(@"Error deleting file from current package: %@", nonFailingError);
-                            nonFailingError = nil;
+                                                                   error:&error];
+                        if (error) {
+                            failCallback(error);
```
lostintangent: Will this cause the update to be black listed? Do we need it to?

iOS/CodePushPackage.m
geof90: Yep.

iOS/CodePush.h
```diff
@@ -104,6 +104,24 @@ failCallback:(void (^)(NSError *err))failCallback;
 
 @end
 
+@interface CodePushUtils : NSObject
+
++ (void)addContentsOfFolderToManifest:(NSString *)folderPath
```
lostintangent: Does this need to be part of the interface?

iOS/CodePushPackage.m
geof90: This will not.. i think we should not. However, maybe I can tweak this slightly to check whether the file exists before deleting it.

1. In the event the file doesn't exist, we should just carry on, because we achieved the effect we want of ensuring the file is removed. The fact that we received a manifest asking us to delete a file that doesnt exist is suspect, but so long as the final hash matches, I think we're good.

2. In the event an IO error occurs in deleting it, we should not blacklist it because it is not a problem with the update.

iOS/CodePushUtils.m
```diff
@@ -0,0 +1,156 @@
...
+                   [[fileName pathExtension] isEqualToString:@"jsbundle"] ||
+                   [[fileName pathExtension] isEqualToString:@"js"]) {
+            return fileName;
+        }
+    }
+    
+    return nil;
+}
+
++ (BOOL)verifyHashForZipUpdate:(NSString *)finalUpdateFolder
```
lostintangent: How about "isValidDiffUpdate"? I prefer using the term "Diff" as opposed to "Zip".

iOS/CodePush.h
```diff
@@ -104,6 +104,24 @@ failCallback:(void (^)(NSError *err))failCallback;
...
+                           pathPrefix:(NSString *)pathPrefix
+                             manifest:(NSMutableArray *)manifest
+                                error:(NSError **)error;
++ (NSString *)computeHash:(NSData *)inputData;
```
lostintangent: I don't think this needs to be part of the public interface either

android/app/src/main/java/com/microsoft/codepush/react/CodePushUpdateUtils.java
```diff
@@ -3,11 +3,66 @@
...
+            try {
+                if (digestInputStream != null) digestInputStream.close();
+                if (dataStream != null) dataStream.close();
+            } catch (IOException e) {
+                e.printStackTrace();
+            }
+        }
+
+        byte[] hash = messageDigest.digest();
+        return String.format("%064x", new java.math.BigInteger(1, hash));
```
lostintangent: Could we use the "android.util.Base64" class here instead of doing this? This article seems to use that solution: http://karanbalkar.com/2013/05/tutorial-28-implement-sha1-and-md5-hashing-in-android/

iOS/CodePushUtils.m
geof90: So we dont want to compute the hash if it is a zip update but not a diff update? I guess i'm fairly confident the unzipping process won't go wrong.

iOS/CodePush.h
```diff
@@ -104,6 +104,24 @@ failCallback:(void (^)(NSError *err))failCallback;
 
 @end
 
+@interface CodePushUtils : NSObject
```
lostintangent: Should we call this `CodePushUpdateUtils`? The current name implies it is more generic than it actually is.

iOS/CodePushPackage.m
```diff
@@ -203,7 +203,9 @@ + (void)downloadPackage:(NSDictionary *)updatePackage
            doneCallback:(void (^)())doneCallback
            failCallback:(void (^)(NSError *err))failCallback
 {
-    NSString *newPackageFolderPath = [self getPackageFolderPath:updatePackage[@"packageHash"]];
+    NSString *newPackageHash = updatePackage[@"packageHash"];
```
lostintangent: While we're making some refactoring changes, should we replace the use of the word "Package" for "Update" (e.g. `newUpdateHash`, `newUpdateDirectoryPath`).

lostintangent:
Could we call the new directory `ios` instead of `iOS`?

android/app/src/main/java/com/microsoft/codepush/react/CodePushPackage.java
```diff
@@ -153,12 +153,13 @@ public WritableMap getPackage(String packageHash) {
 
     public void downloadPackage(Context applicationContext, ReadableMap updatePackage,
                                 DownloadProgressCallback progressCallback) throws IOException {
-
-        String newPackageFolderPath = getPackageFolderPath(CodePushUtils.tryGetString(updatePackage, PACKAGE_HASH_KEY));
-        if (FileUtils.fileAtPathExists(newPackageFolderPath)) {
+        String newUpdateHash = CodePushUtils.tryGetString(updatePackage, PACKAGE_HASH_KEY);
+        String newUpdateFolderPath = getPackageFolderPath(newUpdateHash);
```
dtivel: I notice you're renaming "package" to "update" here and there.  You could also rename "updatePackage", "PACKAGE_HASH_KEY", "PACKAGE_FILE_NAME" and "getPackageFolderPath."

There are probably more symbols that could be renamed, but these are what I saw right here.

It need not be done with this pull request.

android/app/src/main/java/com/microsoft/codepush/react/CodePushUpdateUtils.java
```diff
@@ -1,13 +1,70 @@
...
+        File[] folderFiles = folder.listFiles();
+        for (File file : folderFiles) {
+            String fullFilePath = file.getAbsolutePath();
+            String relativePath = (pathPrefix.isEmpty() ? "" : (pathPrefix + "/")) + file.getName();
+            if (file.isDirectory()) {
+                addContentsOfFolderToManifest(fullFilePath, relativePath, manifest);
+            } else {
+                try {
+                    // Substring 1 because appendPathComponent creates a "/" prefix to the relativePath.
+                    manifest.add(relativePath + ":" + computeHash(new FileInputStream(file)));
```
dtivel: I can't make sense of the comment in context.  Should the comment be removed?

dtivel:
Looks good.

android/app/src/main/java/com/microsoft/codepush/react/CodePushPackage.java
geof90: Yep. I think for ease of review I'll do it in another PR.

Some of them also involve changes in the API, e.g getCurrentPackage -> getCurrentUpdate, which may break users, so I want to leave it off till we have discussed this further internally.

android/app/src/main/java/com/microsoft/codepush/react/CodePushUpdateUtils.java
geof90: oops! that was leftover from when I was trying to iterate on the code. will remove.

PR #209, opened 2016-02-25
Adding note about App Store tip

lostintangent:
Now that we've had a few incidents of App Store rejections due to displaying the update dialog (that were easily mitigated by hiding it), I want to make this prescription very clear to our users, so that they know we recommend not showing it for App Store-distributed apps, whereas all other distribution channels can freely choose to show it as desired.


geof90:
LGTM!

PR #211, opened 2016-02-25
Hash binary resources 

geof90:
This change allows the plugin to obtain a hash of the binary bundled resources, to be used during `updateCheck` requests, so that we can reject updates from the server if the hash is the same. The implementation differs between iOS and Android due to technical reasons.

On iOS:

The hash is generated dynamically during every call to `getCurrentPackage`, if the app is currently running the version bundled with the binary. This is hash is then used to query against the server in `checkForUpdate`. If the hash is the same as the latest one on the server, the server will return no update as per normal. If the server returns a diff update, the `downloadPackage` method will check to see if the currently running version is the one in the binary. If so, it will proceed to copy the necessary files from the binary in order to apply the diff update.

On Android:

Because there is no way to obtain all of the drawable assets bundled into the APK and calculate a hash dynamically, the hash is generated during build time as a build task, and saved in the assets folder. In order for this to work, devs will have to add the `codepush.gradle` task into their `build.gradle` files, and this additional setup step is updated in the README. The task simply hooks into the existing build steps, and takes a snapshot of the built resources before and after the `bundleJsAndAssets` task is run, to figure out which files were generated by React Native, and compute the hash using those files.

Again because there is no way to obtain all of the drawable assets bundled into the APK, we don't have the capability to perform diff updates against the binary version in Android. Therefore, in Android `getCurrentPackage` will not return this hash, so that we do not query the server with the hash and cause the server to return a diff update if there is one. We allow the server to return the full update if there is one, then make a native call to `getBinaryHash` to see if it is the same as the binary version, and if so, reject it.

Also bumped the plugin version to v1.7.4.


scripts/build-and-test.sh
```diff
@@ -1,45 +0,0 @@
-#!/bin/bash
```
geof90: Deleted this script since it is unused.

ios/CodePush/CodePush.m
```diff
@@ -458,11 +459,33 @@ - (void)savePendingUpdate:(NSString *)packageHash
                            rejecter:(RCTPromiseRejectBlock)reject)
 {
     NSError *error;
+    if (isRunningBinaryVersion) {
+        // isRunningBinaryVersion will not get set to "YES" if running against the packager.
+        NSString *binaryHash = [CodePushUpdateUtils getHashForBinaryContents:[CodePush binaryBundleURL] error:&error];
```
lostintangent: Could we save this hash somewhere so that at most, we only have to compute the hash on the first run after a binary installation? I'd prefer to keep app start as light as possible.

ios/CodePush/CodePush.m
```diff
@@ -458,11 +459,33 @@ - (void)savePendingUpdate:(NSString *)packageHash
...
+        if (error) {
+            NSLog(@"Error obtaining hash for binary contents: %@", error);
+            resolve(nil);
```
lostintangent: This behavior is kind of add. Now this method can return null, the metadata about a CodePush update, or the metadata about the binary contents. Should we change it to just always return an object that is either the binary or a CodePush update, and then reject the promise in error conditions?

ios/CodePush/CodePush.m
```diff
@@ -458,11 +459,33 @@ - (void)savePendingUpdate:(NSString *)packageHash
...
+            return;
+        } else if (binaryHash == nil) {
+            resolve(nil);
```
lostintangent: Is this only to accommodate the packager?

ios/CodePush/CodePush.m
```diff
@@ -47,6 +48,11 @@ @implementation CodePush {
 
 #pragma mark - Public Obj-C API
 
++ (NSURL *)binaryBundleURL
```
lostintangent: Do we need to make this part of our public/documented Obj-C API?

ios/CodePush/CodePushUpdateUtils.m
```diff
@@ -3,6 +3,10 @@
 
 @implementation CodePushUpdateUtils
 
+NSString * const ManifestFolderPrefix = @"CodePush";
+NSString * const DefaultJsBundleName = @"main.jsbundle";
+NSString * const DefaultAssetsFolderName = @"assets";
```
lostintangent: Isn't the "assets" folder name static/fixed? If so, the name "Default" is maybe a little misleading.

ios/CodePush/CodePushUpdateUtils.m
```diff
@@ -3,6 +3,10 @@
 
 @implementation CodePushUpdateUtils
 
+NSString * const ManifestFolderPrefix = @"CodePush";
+NSString * const DefaultJsBundleName = @"main.jsbundle";
```
lostintangent: This seems a little brittle since we know people can rename it. Shouldn't we try to get the actual bundle file name from the CodePush class?

ios/CodePush/CodePushUpdateUtils.m
```diff
@@ -127,6 +131,55 @@ + (NSString *)findMainBundleInFolder:(NSString *)folderPath
...
+{
+    return DefaultAssetsFolderName;
+}
+
++ (NSString *)getDefaultJsBundleName
+{
+    return DefaultJsBundleName;
+}
+
++ (NSString *)getHashForBinaryContents:(NSURL *)binaryBundleUrl
```
lostintangent: There seems to be a decent amount of duplication between this and verifyHashForDiffUodate. Is it possible to refactor them to share some of this logic?

ios/CodePush/CodePush.m
lostintangent: Actually, I wonder if it wouldn't make more sense for getConfigiration to return the binary hash and update checkForUpdate to detect a null current package and use the config' hash and version instead. This seems to provided a cleaner seperate on between getCurrentPackage and getConfigiration. Since we don't document getConfiguration, it also allows us to add this new behavior without having to modify the public JS API.

ios/CodePush/CodePush.h
```diff
@@ -109,8 +111,18 @@ failCallback:(void (^)(NSError *err))failCallback;
 + (void)copyEntriesInFolder:(NSString *)sourceFolder
                  destFolder:(NSString *)destFolder
                       error:(NSError **)error;
+
 + (NSString *)findMainBundleInFolder:(NSString *)folderPath
                                error:(NSError **)error;
+
++ (NSString *)getDefaultAssetsFolderName;
++ (NSString *)getDefaultJsBundleName;
```
lostintangent: Are either of these two methods currently used outside of its enclosing class? If not, I'd rather omit them from the interface.

geof90: They are used in CodePushPackage, thats why I exposed them

ios/CodePush/CodePush.m
geof90: It is used in both CodePushPackage and CodePushUpdateUtils to get the binaryURL to 1. calculate the hash 2. apply diff updates

ios/CodePush/CodePush.m
geof90: Sounds like a good idea!

ios/CodePush/CodePush.m
geof90: Yup.

ios/CodePush/CodePushUpdateUtils.m
geof90: My thought was, even if they rename it locally, the one deployed to our server will be called `main.jsbundle` if they use the `release-react` command, so in order to generate a matching hash we should rename it for them in the manifest.

ios/CodePush/CodePushUpdateUtils.m
geof90: Refactored

ios/CodePush/CodePush.h
geof90: Got rid of getDefaultJsBundleName.

ios/CodePush/CodePushUpdateUtils.m
geof90: Removed this completely.

ios/CodePush/CodePushUpdateUtils.m
geof90: renamed to "AssetsFolderName"

ios/CodePush/CodePush.m
```diff
@@ -448,7 +435,29 @@ - (void)savePendingUpdate:(NSString *)packageHash
...
+    if (isRunningBinaryVersion) {
+        // isRunningBinaryVersion will not get set to "YES" if running against the packager.
+        NSString *binaryHash = [CodePushUpdateUtils getHashForBinaryContents:[CodePush binaryBundleURL] error:&error];
+        if (error) {
+            NSLog(@"Error obtaining hash for binary contents: %@", error);
+            resolve(configuration);
+            return;
+        }
+        
+        if (binaryHash == nil) {
```
lostintangent: Is this accommodating the scenario where you built the app against the simulator and it didn't create the bundle? I would love to add a simple comment to these kinds of statements, since it will be hard to remember what it's purpose is in the future.

ios/CodePush/CodePushUpdateUtils.m
```diff
@@ -127,6 +154,76 @@ + (NSString *)findMainBundleInFolder:(NSString *)folderPath
...
++ (NSString *)getAssetsFolderName
+{
+    return AssetsFolderName;
+}
+
++ (NSString *)getHashForBinaryContents:(NSURL *)binaryBundleUrl
+                                 error:(NSError **)error
+{
+    if (!didLoadBinaryHash) {
+        didLoadBinaryHash = YES;
```
lostintangent: Doesn't the JS side already cache the results of calling getConfiuration the first time it does? If so, we wouldn't need to cache this on the native side as well. I'd still love to think about how to cache this in user defaults as well, so we don't have to recompile this every time the app is started.

geof90: Oh, it is also caching in user defaults, if you scroll down a little.

ios/CodePush/CodePush.m
geof90: Added a comment below

scripts/getFilesInFolder.js
```diff
@@ -0,0 +1,18 @@
+var fs = require("fs");
+
+// Utility function that collects the stats of every file in a directory
+// as well as in its subdirectories.
+function getFilesInFolder(folderName, fileList) {
+    var folderFiles = fs.readdirSync(folderName);
+    folderFiles.forEach(function(file) {
+        var fileStats = fs.statSync(folderName + "/" + file);
```
lostintangent: Would it be better to use `path.join` here and the two places below? To fully accommodate this being run on Windows/OS X/Linux?

scripts/generateBundledResourcesHash.js
```diff
@@ -0,0 +1,93 @@
...
+var crypto = require("crypto");
+var fs = require("fs");
+var path = require("path");
+
+var getFilesInFolder = require("./getFilesInFolder");
+
+var CODE_PUSH_FOLDER_PREFIX = "CodePush";
+var CODE_PUSH_HASH_FILE_NAME = "CodePushHash.json";
+var HASH_ALGORITHM = "sha256";
+var JS_BUNDLE_FILE_NAME = "main.jsbundle";
```
lostintangent: We don't know for certain that this is what the file will be named do we?

CodePush.js
```diff
@@ -56,14 +64,19 @@ async function checkForUpdate(deploymentKey = null) {
...
    *    bug in the server, but we're adding this check just to double-check that the
    *    client app is resilient to a potential issue with the update check.
+   * 4) The server said there is an update, but the update's hash is the same as that
+   *    of the binary's currently running version. This should only happen in Android -
+   *    unlike iOS, we don't attach the binary's hash to the updateCheck request
+   *    because we want to avoid having to install diff updates against the binary's
+   *    version, which we can't do yet on Android.
    */
-  if (!update || update.updateAppVersion || localPackage && (update.packageHash === localPackage.packageHash)) {
+  if (!update || update.updateAppVersion || localPackage && (update.packageHash === localPackage.packageHash) || config.packageHash === update.packageHash) {
```
lostintangent: For the last expression, don't you also want to check that `localPackage` is `null`? Since we wouldn't care about the binary's hash if the user was running on a CodePush update?

ios/CodePush/CodePushUpdateUtils.m
```diff
@@ -127,6 +150,70 @@ + (NSString *)findMainBundleInFolder:(NSString *)folderPath
     return nil;
 }
 
++ (NSString *)getAssetsFolderName
```
lostintangent: We don't need the `get` prefix.

ios/CodePush/CodePushUpdateUtils.m
```diff
@@ -127,6 +150,70 @@ + (NSString *)findMainBundleInFolder:(NSString *)folderPath
...
+    
+    // Cache the hash in user preferences. This assumes that the modified date for the
+    // JS bundle changes every time a new bundle is generated by the packager.
+    [binaryHashDictionary setObject:binaryHash forKey:binaryModifiedDate];
+    [preferences setObject:binaryHashDictionary forKey:BinaryHashKey];
+    [preferences synchronize];
+    return binaryHash;
+}
+
++ (NSString *)getManifestFolderPrefix
```
lostintangent: We don't need the `get` prefix.

PR #212, opened 2016-02-25
Removing JS bundle modified date mutation

lostintangent:
This PR removes the logic which explicitly sets the modified date of the JS bundle file when installing a ZIP update. This behavior was originally needed to determine precedence between a CodePush update and the binary contents, but we actually stopped relying on this a while ago, and therefore, it isn't necessary.

Additionally, this behavior causes an issue when the user releases an update whose JS bundle is read-only, and therefore, we can't mutate. By removing this logic, we can accommodate that scenario.


geof90:
LGTM

Issue #214, opened 2016-02-27
Install Mode Error!


DHidee66:
Hi,

I just got this error when my app was closed.
![screen shot 2016-02-27 at 4 24 35 pm](https://cloud.githubusercontent.com/assets/6751025/13373690/cb094558-dd6e-11e5-849d-f5922146b854.png)

Thank You!

lostintangent:
@DHidee66 That error indicates that the CodePush module hasn't been fully installed, and therefore, the "proxy" object isn't being exposed to JS. Could you double check the install instructions to make sure everything is setup?

DHidee66:
Hey @lostintangent , 
I have a GCM module that does something in the background when my app is closed and it receives a notification. Then is when the error appears. 
When i run the project normally I get the usual CodePush logs that it checked for a new version and there was no updates.

lostintangent:
Ah OK that's interesting. Are you calling any CodePush APIs in response to the notification?

lostintangent:
Also out of curiosity, are you using this module for handling the notifications? https://github.com/oney/react-native-gcm-android?

DHidee66:
@lostintangent Yes that's the one. And the sync() method is not called.

lostintangent:
Hmm OK. I just pinged Howard who is the owner of that module (and also uses CodePush in his apps) to see if he has any insight. I'll investigate into it as well to see what might be going on. We haven't seen this issue before so I appreciate you reporting it!

Just to make sure I'm trying to repro the same scenario, could you let me know which React Natice you're using? RN 0.20.?

Additionally, does the JS file that includes your notification handler also include a require/import statement for the "react-native-code-push" module? The error you're seeing would actually come from simply loading our JS module, as opposed to calling sync.

lostintangent:
It seems possible that this issue isn't related to CodePush, but rather, is being caused by this bug in the RN GCM module: https://github.com/oney/react-native-gcm-android/issues/26.

If you removed any imports/require statements for CodePush and any sync calls from your project, does this issue go away entirely? Or do you see a similar error but for a different module? I'm just wondering if the CodePush error you're seeing is just a symptom of an underlying issue.

DHidee66:
[...]

PR #215, opened 2016-02-27
Make loading our JS module more permissive

lostintangent:
Currently, simply `import`ing or `require`ing the `react-native-code-push` JS module requires that the native-end of the module also be fully installed/initialized. This change simply removes this dependency by only constructing the JS API in the event that the native module was initialized, which allows an app to load our module without getting a cryptic error. This addresses issue #214.

If an app tried to access any members on the module export (e.g. `codePush.sync()`), without the native module being installed properly or initialized, that would still error out, but at least it would be a more intuitive/explicit action that led to the error, as opposed to simply importing the module, which is arguably weird to have result in an app crash.

Additionally, I added some new logging in the event that the native module isn't initialized, which should allow for easier diagnostics when users haven't properly installed the plugin.


geof90:
LGTM!

PR #218, opened 2016-02-29
Handle uncleared updates during debug

geof90:
This fixes an issue in debug builds, where we do not clear "outdated" updates that were installed in an older version of the binary. This causes `getCurrentPackage` to return a non null value, which results in `checkForUpdate` not rejecting updates if they have the same hash as the currently running binary version.

This adds an (undocumented) `isRunningBinaryVersion` field to the current update returned to JS to indicate to JS that although there exists a current update, it is outdated and not running.


CodePush.js
```diff
@@ -70,7 +70,12 @@ async function checkForUpdate(deploymentKey = null) {
    *    because we want to avoid having to install diff updates against the binary's
    *    version, which we can't do yet on Android.
    */
-  if (!update || update.updateAppVersion || localPackage && (update.packageHash === localPackage.packageHash) || !localPackage && config.packageHash === update.packageHash) {
+  if (
+    !update ||
```
lostintangent: What if you moved this expression to the same line as the `(`? That might make this look a little cleaner, and then align the other three with it.

ios/CodePush/CodePush.m
```diff
@@ -480,6 +480,13 @@ - (void)savePendingUpdate:(NSString *)packageHash
         return;
     }
     
+    if (isRunningBinaryVersion) {
+        // This only matters in Debug builds. Since we do not clear "outdated" updates,
+        // we need to indicate to the JS side that somehow we have a current update on
+        // disk that is not actually running.
+        [package setObject:@(isRunningBinaryVersion) forKey:@"_isDebugOnly"];
```
lostintangent: Would it be easier to read this if we just set it to true instead of relying on the `isRunningBinaryVersion` variable?

android/app/src/main/java/com/microsoft/codepush/react/CodePush.java
```diff
@@ -423,6 +423,10 @@ protected Void doInBackground(Object... params) {
                         return null;
                     }
 
+                    if (isRunningBinaryVersion) {
+                        currentPackage.putBoolean("_isDebugOnly", isRunningBinaryVersion);
```
lostintangent: Same comment as with the iOS version

lostintangent:
LGTM!

PR #219, opened 2016-03-01
Update Readme: codepush.gradle should be build.gradle

dejakob:


geof90:
Hi @dejakob, codepush.gradle is the correct file. The docs in master apply to a version of the plugin that is not released yet, so it might be confusing that the latest version from NPM doesnt contain the codepush.gradle file.

dejakob:
Oh, I see. So, I should be using the build.gradle file in the latest release?

geof90:
You can just ignore and skip step 3 of https://github.com/microsoft/react-native-code-push#plugin-installation-android. So, no need to use any of our gradle files.

dejakob:
Ok, thank you :)

Issue #210, opened 2016-02-25
Mandatory updates + install mode when calling sync


lostintangent:
Currently, the Cordova plugin forces `InstallMode.IMMEDIATE` whenever it encounters a mandatory update, regardless which `InstallMode` option was actually passed to `sync`. It's possible that this behavior makes sense, since currently, the mandatory flag has no meaning to a silent install (for example), and therefore, the developer has no way of enforcing this policy from the server. This seems less than ideal since I've heard of a few instances already where the Cordova behavior was actually beneficial for folks (e.g. forcing a critical bug fix when the potentially jarring restart was worth it).

Since we're prescribing silent installs for App Store apps, I think we should rationalize this scenario to enable truly mandatory updates. One simple way I can think about handling this is by adding a new option to `sync` called `mandatoryInstallMode`, which defaults to `InstallMode.IMMEDIATE`, and controls which `InstallMode` to use for mandatory updates. That way, users can effectively control how to install mandatory vs. optional updates separately, without there being any odd "magic". This would be extremely simple to implement as well :)

Thoughts?

*NOTE: This wouldn't and shouldn't impact `LocalPackage.install` since we don't apply any opinion for this workflow. The developer could easily inspect the `mandatory` flag on the `LocalPackage` before calling `install`, and therefore, they would already have full control.*

PR #221, opened 2016-03-02
mandatoryInstallMode

lostintangent:
This PR introduces a new option for the `sync` method called `mandatoryInstallMode` which allows devs to determine the `InstallMode` which should be used for mandatory updates. It defaults to `InstallMode.IMMEDIATE` in order to reflect the behavior that most of our users expect, but this can be customized (e.g. if you want optional updates to install on restart and mandatory on resume to be a little less invasive).

Without this behavior, devs don't have an "escape hatch" if they use a silent update via `sync`, and latter need to force a critical bug fix in the wild. With this change, they could rely on optional updates for most things, and have the option for pushing mandatory updates into production if/when they need it.

This addresses #210 and achieves parity with the existing Cordova behavior.


geof90:
LGTM!

PR #225, opened 2016-03-03
Remove semver use

geof90:
Removed the use of semver, which complains about version values missing a patch version (e.g "1.0"). Updated the docs to state that missing patch versions will be assumed as "0", i.e. ("1.0" -> "1.0.0"). Also updated the docs to reflect the fact that the target binary version can now be a range expression as per https://github.com/Microsoft/code-push/pull/149, and also update the docs to include the `release-react` command.


README.md
```diff
@@ -345,7 +345,26 @@ And that's it! for more information regarding the CLI and how the release (or pr
...
+```
+code-push release-react Foo ios
+```
+
+Once the command has finished execution, your latest update should be released to CodePush! For more information about the `release-react` command and its parameters, refer to the [CLI documentation](http://codepush.tools/docs/cli.html#releasing-updates-to-a-react-native-app).
+
+Additionally, the CodePush client supports differential updates, so even though you are releasing your JS bundle and assets on every update, your end users will only actually download the files they need. The service handles this automatically so that you can focus on creating awesome apps and we can worry about optimizing end user downloads.
+
+*Note: Releasing assets via CodePush requires that you're using React Native v0.15.0+ and CodePush v1.4.0+ (for iOS) and React Native v0.19.0+ and CodePush v1.7.0+ (Android). If you are using assets and an older version of the CodePush plugin, you should not release updates via CodePush, because it will break your app's ability to load images from the binary. Please test and release appropriately!*
```
shishirx34: *Note: Releasing assets via CodePush requires that you're using React Native v0.15.0+ and CodePush v1.4.0+ (for iOS) and React Native v0.19.0+ and CodePush v1.7.0+ (Android). If you are using assets and an older version of the CodePush plugin, you should not release updates **with assets** via CodePush, because it will break your app's ability to load images from the binary. Please test and release appropriately!*

- You can still use older plugins to release, just not assets, right?

geof90:
@lostintangent I removed all the doc changes.

CodePush.js
```diff
@@ -41,7 +40,7 @@ async function checkForUpdate(deploymentKey = null) {
    * in any updates for current app store version, regardless of hash.
    */
   let queryPackage;
-  if (localPackage && localPackage.appVersion && semver.compare(localPackage.appVersion, config.appVersion) === 0) {
+  if (localPackage && localPackage.appVersion && localPackage.appVersion === config.appVersion) {
```
lostintangent: I actually wonder whether we shouldn't just remove the second and third conditions entirely, and rely on the native side to determine whether a local package is still valid or not. I believe we had this check in back before we started clearing updates on disk, so it's potentially just legacy cruft that we don't need anymore. I don't believe that the Cordova plugin does any validation like this before using a local package to query the server with.

lostintangent: As I think about this more, since we already clear updates when a new binary is installed in release, the only time that this last expression would matter is after a binary update in debug mode. In those scenarios, it's likely that the dev didn't bump the version anyways, and therefore, this version check would only prevent using an update in debug mode, where the dev actually bumped the version. If we care about this scenario, I would argue that it might be a cleaner solution to change our logic to clear updates in debug mode if the version has changed.

lostintangent: I just quickly looked at the iOS implementation, and we already know the update and binary version, so it would be super simple to change the logic to clear updates in debug mode is the versions are different. That way, our assumption is that if you're not bumping the version in debug mode, we keep old updates so you don't get unnecessary CodePush updates, but as soon as you bump the version, even in debug mode, we clear updates. We actually need to do this anything to accommodate the new range feature, since the newly bumped version may no longer be eligible for the locally installed updates.

geof90: I've updated the logic according to your suggestion

PR #228, opened 2016-03-04
Fixing Podspec for Swift consumption

### LINKED ISSUE #226: Unable to build via CocoaPods in project using Swift
I'm using the version 1.7.3-beta and am also installing React 0.20.0 via CocoaPods. When I install CodePush via cocoapods I am unable to build my project. 

The error I get is: "RCTBridgeModule.h file not found".

lostintangent:
This fixes #226 by making the following changes:

1. Moves the use of the `RCTBridgeModule` protocol from the public interface of our `CodePush` class, to the private implementation (via a class extension). This allows the class to have the same behavior, but the public header no longer needs to import `RCTBridgeModule.h`, which doesn't work when we are built as a framework via CocoaPods (since frameworks can't use quoted imports to headers that aren't embedded within them)

2. Explicitly sets the public header in our Podspec, so that CocoaPods doesn't include the SSArchive headers into it when it generates the module map when building as a framework

Additionally, unrelated to this bug, I updated the Podspec to accomodate the Xcode project refactoring we did, and I also bumped the version for our upcoming release.

### DIFF
--- a/ios/CodePush/CodePush.h
+++ b/ios/CodePush/CodePush.h
@@ -1,6 +1,6 @@
-#import "RCTBridgeModule.h"
+#import <Foundation/Foundation.h>
 
-@interface CodePush : NSObject <RCTBridgeModule>
+@interface CodePush : NSObject
 
 + (NSURL *)binaryBundleURL;
 /*
--- a/CodePush.podspec
+++ b/CodePush.podspec
@@ -1,16 +1,17 @@
 Pod::Spec.new do |s|
 
-  s.name             = 'CodePush'
-  s.version          = '1.7.3-beta'
-  s.summary          = 'React Native plugin for the CodePush service'
-  s.author           = 'Microsoft Corporation'
-  s.license          = 'MIT'
-  s.homepage         = 'http://microsoft.github.io/code-push/'
-  s.source           = { :git => 'https://github.com/Microsoft/react-native-code-push.git', :tag => "v#{s.version}" }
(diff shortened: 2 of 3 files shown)


jkhustler:
:+1: thanks for the quick fix! :)

lostintangent:
@jkhustler Would you be able to NPM install the bits from this commit and verifying that it works for you? Our testing shows that it's working, but I would love to get a sanity check from you as well if that wouldn't be an inconvenience. Thanks!

jkhustler:
I've just loaded the `cocoapods` branch directly via cocoapods as a dependency and put `import CodePush

lostintangent:
Awesome! Thank you so much for checking that :) I'll just wait for a sign-off from @geof90 before merging this, but we'll get it out ASAP. In the meantime, are you cool with depending on the `cocoapods` branch as you currently are?

jkhustler:
yes, I can wait.

ios/CodePush/CodePush.h
```diff
@@ -1,6 +1,6 @@
-#import "RCTBridgeModule.h"
+#import <Foundation/Foundation.h>
```
lostintangent: BTW, this change was necessary because we're using `NSObject` and were getting away with not importing this header because `RCTBridgeModule.h` imported it for us. Now that we've removed that dependency, we now need to explicitly import the `Foundation.h` file to ensure `NSObject` resolves correctly.

geof90:
LGTM!

PR #236, opened 2016-03-10
Minimum background duration

lostintangent:
This introduces a new parameter to `sync` and `LocalPackage.install` called `minimumBackgroundDuration` that allows you to perform resume-based installs (which get the update in front of users much sooner than restart-based), without being too obtrusive. This way, if you're comfortable assuming that after 10 minutes of inactivity (for example), the end user wouldn't expect to pick up on their previous operation, then you can choose to restart the app. However, if the app is only in the background for a few seconds or a couple minutes, that might represent the user quickly switching contexts (e.g. responding to a text message), and it would be disruptive to restart the app when they return.


shishirx34:
Cool feature :+1: LGTM!

ios/CodePush/CodePush.m
```diff
@@ -12,6 +12,8 @@ @interface CodePush () <RCTBridgeModule>
 @implementation CodePush {
     BOOL _hasResumeListener;
     BOOL _isFirstRunAfterUpdate;
+    int _minimumBackgroundDuration;
+    NSDate * _lastResignedDate;
```
geof90: no space between the * and the _?

lostintangent: Ah good catch.

CodePush.js
```diff
@@ -273,7 +274,11 @@ async function syncInternal(options = {}, syncStatusChangeCallback, downloadProg
             if (resolvedInstallMode == CodePush.InstallMode.ON_NEXT_RESTART) {
               log("Update is installed and will be run on the next app restart.");
             } else {
-              log("Update is installed and will be run when the app next resumes.");
+              if (syncOptions.minimumBackgroundDuration > 0) {
```
geof90: Can we Math.floor() this value before using it and logging to the user since we are not supporting decimal values?

CodePush.js
```diff
@@ -302,7 +307,7 @@ async function syncInternal(options = {}, syncStatusChangeCallback, downloadProg
       resolvedInstallMode = localPackage.isMandatory ? syncOptions.mandatoryInstallMode : syncOptions.installMode;
       
       syncStatusChangeCallback(CodePush.SyncStatus.INSTALLING_UPDATE);
-      await localPackage.install(resolvedInstallMode, () => {
+      await localPackage.install(resolvedInstallMode, syncOptions.minimumBackgroundDuration, () => {
```
geof90: Should we do

`typeof syncOptions.minimumBackgroundDuration === "number" ? syncOptions.minimumBackgroundDuration : 0`

to prevent an exception?

geof90:
LGTM!

Issue #237, opened 2016-03-11
One of the dependency is throwing timeout error


mohanrohith:
`\m/⚡  ===> npm install --save react-native-code-push
npm ERR! git fetch -a origin (git://github.com/visionmedia/superagent.git) fatal: unable to connect to github.com:
npm ERR! git fetch -a origin (git://github.com/visionmedia/superagent.git) github.com[0: 192.30.252.131]: errno=Operation timed out
npm ERR! Darwin 14.5.0
npm ERR! argv "/usr/local/bin/node" "/usr/local/bin/npm" "install" "--save" "react-native-code-push"
npm ERR! node v4.2.3
npm ERR! npm  v3.8.0
npm ERR! code 128

npm ERR! Command failed: git fetch -a origin
npm ERR! fatal: unable to connect to github.com:
npm ERR! github.com[0: 192.30.252.131]: errno=Operation timed out
npm ERR! 
npm ERR! 
npm ERR! 
npm ERR! If you need help, you may report this error at:
npm ERR!     <https://github.com/npm/npm/issues>

npm ERR! Please include the following file with any support request:

dtivel:
Hi, mohanrohith.

Did any other dependencies download successfully?  If you run the npm command again, does it succeed in downloading this dependency (superagent)?

I wonder if this was a temporary network issue.  Can you ping github.com successfully?

lostintangent:
@mohanrohith We just submitted a fix via #242 that removes this GitHub-based dependency. I'm going to close this for now, assuming that you aren't experiencing issues with downloading NPM-based dependencies. Please re-open this if that isn't the case, and we can investigate it further. Thanks for reporting this!

PR #242, opened 2016-03-12
Upgrading SDK dependency

lostintangent:
Simply upgrading our SDK dependency, which among other things, removes the transitive dependency on the GitHub-based `superagent` version. Issue #237 could still exist depending on the developer's network conditions, but we mine as well consolidate the number of locations that dependencies are pulled from to just NPM.


geof90:
LGTM!

PR #243, opened 2016-03-12
Simplify JS bundle discovery strategy

lostintangent:
This PR removes the "fuzzy logic" for locating the JS bundle in a downloaded update, and instead, looks for the exact same file name that the plugin already looks for in the binary. This simplifies things in the following ways:

1. Looking for `.js` files is potentially error prone, since we've seen cases where apps ship `WebView` hosted content, including `.js` files. We want to allow app devs to use `.js` instead of `.jsbundle`, but only if they explicitly tell us they are using that extension via `[CodePush bundleURLForResource:withExtension]`. If an app is configured to use `.jsbundle` files, it doesn't make sense for us to look for or consider `.js` files as valid app bundles.

2. Looking for the first file that uses any of these extensions makes it easier for our logic to have a false positive, if the update includes both `.js` and `.jsbundle` files.

3. If a dev wants to call their JS bundle file `foo.bar`, this change would allow it, since we're no longer restricted to a few extension types, as opposed to simply looking for the same file that was shipped with the binary

As long as everything uses the same bundle file name, then the E2E should be more reliable, while still being flexible enough to allow devs to configure a different name and/or extension when necessary.

### DIFF
--- a/android/app/src/main/java/com/microsoft/codepush/react/CodePush.java
+++ b/android/app/src/main/java/com/microsoft/codepush/react/CodePush.java
@@ -365,7 +365,7 @@ public class CodePush {
                     try {
                         WritableMap mutableUpdatePackage = CodePushUtils.convertReadableMapToWritableMap(updatePackage);
                         mutableUpdatePackage.putString(BINARY_MODIFIED_TIME_KEY, "" + getBinaryResourcesModifiedTime());
-                        codePushPackage.downloadPackage(applicationContext, mutableUpdatePackage, new DownloadProgressCallback() {
+                        codePushPackage.downloadPackage(applicationContext, mutableUpdatePackage, CodePush.this.assetsBundleFileName, new DownloadProgressCallback() {
                             @Override
                             public void call(DownloadProgress downloadProgress) {
                                 getReactApplicationContext()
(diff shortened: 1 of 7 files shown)


geof90:
LGTM!

android/app/src/main/java/com/microsoft/codepush/react/CodePush.java
```diff
@@ -365,7 +365,7 @@ protected Void doInBackground(Object... params) {
                     try {
                         WritableMap mutableUpdatePackage = CodePushUtils.convertReadableMapToWritableMap(updatePackage);
                         mutableUpdatePackage.putString(BINARY_MODIFIED_TIME_KEY, "" + getBinaryResourcesModifiedTime());
-                        codePushPackage.downloadPackage(applicationContext, mutableUpdatePackage, new DownloadProgressCallback() {
+                        codePushPackage.downloadPackage(applicationContext, mutableUpdatePackage, CodePush.this.assetsBundleFileName, new DownloadProgressCallback() {
```
ericmuyser: Smooth moves

ericmuyser:
Sweet. Thanks for this @lostintangent!

lostintangent:
@ericmuyser Absolutely! If you have the time to put your fork up somewhere where we can review it, I would be happy to make the other API adjustments to fully accommodate RN apps using the WebView component.

Issue #238, opened 2016-03-11
iOS crash on app boot


avishayil:
Since upgrading to 1.8.0 I encounter the following error.
I tried to re-manual link the library, checked build settings & build phases for the correct settings.
````
2016-03-11 13:07:25.408 Geektime[38245:350139] BuddybuildSDK : Disabled in the simulator
**2016-03-11 13:07:25.409 Geektime[38245:350139] Loading JS bundle from (null)**
2016-03-11 13:07:25.452 Geektime[38245:350139] [Crashlytics] Version 3.7.0 (102)
2016-03-11 13:07:25.476 [fatal][tid:com.facebook.React.JavaScript] Unable to execute JS call: __fbBatchedBridge is undefined
2016-03-11 13:07:25.573 Geektime[38245:350177] *** Terminating app due to uncaught exception 'RCTFatalException: Unable to execute JS call: __fbBatchedBridge is undefined', reason: 'Unable to execute JS call: __fbBatchedBridge is undefined'
*** First throw call stack:
(
	0   CoreFoundation                      0x000000010a908e65 __exceptionPreprocess + 165
	1   libobjc.A.dylib                     0x0000000109b0cdeb objc_exception_throw + 48
	2   CoreFoundation                      0x000000010a908d9d +[NSException raise:format:] + 205
	3   Geektime                            0x0000000106d7363f RCTFatal + 404
	4   Geektime                            0x0000000106d91237 __69-[RCTBatchedBridge _actuallyInvokeAndProcessModule:method:arguments:]_block_invoke + 55
	5   Geektime                            0x0000000106d8bf9e __52-[RCTJSCExecutor _executeJSCall:arguments:callback:]_block_invoke + 1398
	6   Geektime                            0x0000000106d8c5da -[RCTJSCExecutor executeBlockOnJavaScriptQueue:] + 188
	7   Geektime                            0x0000000106d8b9d4 -[RCTJSCExecutor _executeJSCall:arguments:callback:] + 180
	8   Geektime                            0x0000000106d8b80c -[RCTJSCExecutor callFunctionOnModule:method:arguments:callback:] + 166
	9   Geektime                            0x0000000106d911ca -[RCTBatchedBridge _actuallyInvokeAndProcessModule:method:arguments:] + 189
	10  Geektime                            0x0000000106d906ae __39-[RCTBatchedBridge enqueueJSCall:args:]_block_invoke + 360
[...]

PR #244, opened 2016-03-14
Checking for null binary bundle

lostintangent:
This PR attempts to address #238 by adding a better error message. We can't resolve this issue, since it's a React Native behavior, but we can definitely display a clear message with possible next steps.

I also created a new `#pragma` section to distinguish the public API from the test-only methods, which happen to be public.

### DIFF
--- a/ios/CodePush/CodePush.m
+++ b/ios/CodePush/CodePush.m
@@ -221,8 +221,7 @@ static NSString *bundleResourceName = @"main";
         errorMessage = [NSString stringWithFormat:@"The specified JS bundle file wasn't found within the app's binary. Is \"%@\" the correct file name?", [bundleResourceName stringByAppendingPathExtension:bundleResourceExtension]];
 #endif
         
-        RCTFatal([NSError errorWithDomain:@"CodePushError" code:-1
-                                 userInfo:@{ NSLocalizedDescriptionKey: NSLocalizedString(errorMessage, nil) }]);
+        RCTFatal([CodePushErrorUtils errorWithMessage:errorMessage]);
     }
 }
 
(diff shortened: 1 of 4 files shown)


ios/CodePush/CodePush.m
```diff
@@ -190,6 +196,36 @@ - (void)dealloc
...
+        "2. Force the JS bundle to be generated in simulator builds by removing the if block which echoes "
+        "\"Skipping bundling for Simulator platform\" in the \"node_modules/react-native/packager/react-native-xcode.sh\" file.\n\n"
+        
+        "3. Deploy a release build to the simulator, which unlike debug builds, will generate the JS bundle (React Native >=0.22.0 only).";
+#else
+        errorMessage =  = [NSString stringWithFormat:@"The specified JS bundle file wasn't found within the app's binary. Is \"%@\" the correct file name?", [bundleResourceName stringByAppendingPathExtension:bundleResourceExtension]];
+#endif
+        
+        RCTFatal([NSError errorWithDomain:@"CodePushError" code:-1
+                                 userInfo:@{ NSLocalizedDescriptionKey: NSLocalizedString(errorMessage, nil) }]);
```
geof90: Probably some future refactoring work we could do is to have a utility method "createCodePushError", like how we have "isCodePushError". The "CodePushError" domain string and code is also defined in the `CodePushPackage` class as constants and it would be great if we could reuse those / keep those defined in one place.

geof90:
Should we do something similar for Android? We could point users to either edit the `react.gradle` file https://github.com/facebook/react-native/blob/master/local-cli/generator-android/templates/src/app/react.gradle#L84 or run their app in release mode.

geof90:
LGTM!

lostintangent:
@geof90 I created a new `CodePushErrorUtils` class and refactored the `NSError` creation and detection code into it. While I was at it, I re-organized the `CodePushPackage` class so that the methods were alpha-sorted.

Issue #247, opened 2016-03-15
[Android] Rnpm link and Manual code in README doesn't match.


BilalBudhani:
Code generated by `rnpm link` and the code provided for manually linking the library in README doesn't match for Android platform.

lostintangent:
@BilalBudhani Yeah that's why we don't currently document the use of RNPM for Android, since our plugin isn't particularly compatible with the expectations that it has. iOS works perfectly, but Android has some snags.

You should be able to fix it though if you perform the manual steps, and do the following additional items that back out of some changes that RNPM made:

1. Renove the import statement for "com.microsoft.codepush.react.CodePushReactPackage".

3. Remove the "new CodePushReactPackage()" expression that RNPM added to the getPackages method. 

And you

lostintangent:
@BilalBudhani Any luck? :)

lostintangent:
We just updated the docs to call out these changes for RNPM. I'll close this for now but please re-open this if this workaround didn't work for you. Thanks for reporting it!

BilalBudhani:
@lostintangent Sorry for replying late I was traveling for last couple of days. 

I did get the plugin setup done manually on Android and it works perfectly fine. Thanks for guiding me out there. I went through the doc changes in regards to `rnpm` for Android as well. It sounds good to me. :+1:

lostintangent:
Awesome! Thanks for confirming :)

PR #250, opened 2016-03-16
Docs

lostintangent:
This addresses #247 by adding the necessary detail to our Android installation steps for devs that are using RNPM. Hopefully we can remove these docs at some point, once we better support RNPM's assumptions, but until then, this is necessary.


geof90:
LGTM!

PR #252, opened 2016-03-17
Binary hash w/o assets

lostintangent:
The binary hashing logic assumes an app is using assets, and if it isn't (the `assets` folder doesn't exist), then it short circuits generating the hash. This PR simply checks for the existence of the assets, so that the hash can still be generated even if they don't exist.


geof90:
LGTM!

PR #255, opened 2016-03-19
Fix bundle file name to support multi-bundles

Jackong:


lostintangent:
Hey @Jackong! Thanks for submitting this PR. Could you provide a little context about this fix? Did we cause a regression, or does this add support for a new scenario?

Jackong:
@lostintangent I'm integrating react-native with an existing project. Each team is developing for different modules, so I need multiple bundles for different modules on my project.

lostintangent:
So you have multiple different React Native views in the same Android app? Does each view use a different JS bundle file name?

Jackong:
Yes, I have multiple different React-Native activities/views in same APP. And every activities is developed by different teams. So When I use different bundle file name with different code-push APP on my project. I found it do not work because of the hard-code `UPDATE_BUNDLE_FILE_NAME`.

lostintangent:
Got it. So you created different CodePush apps and configured each React Native view with a different deployment key and JS bundle name? And you'll be releasing individual updates to CodePush for each activity?

Jackong:
Yes. But to be honest, I'm not sure whether is right to use different bundle name and deployment key. Will this lead to some issue?

lostintangent:
I'm not sure if this PR alone would help support this scenario. Each downloaded CodePush update is stored in a directory called "CodePush/{HASH}" within the app's local storage. Because of that, the two JS bundles for your different Activities shouldn't actually conflict, unless they had the exact same hash, which would seem pretty rare. I think this PR is a good change, but I'm actually surprised that the JS bundle name was relevant.

However, one problem you'll encounter is related to the fact that CodePush maintains a JSON file in the root of the "CodePush" directory that determines which update is the current. Since there is only one of these files, a single Android app couldn't include multiple Acitivites, each of which had an active CodePush update, regardless what the JS bundle file name is.

We would presumably need to allow the CodePush class in our Java API to allow you to specify an "AppName" as well that forces it to use its own directory within the "CodePush" directory. That way, each Activity would be fully isolated from each other.

lostintangent:
There may be other work needed to support this scenario, but I'm cool with this change. @geof90?

geof90:
Yup, I'm okay with this change.

philip-tong:
@Jackong 
Have you successfully load different bundle with codepush?

PR #257, opened 2016-03-22
Simplify android setup

geof90:
This makes our plugin more compatible with RNPM by making the `CodePush` class itself implement `ReactPackage`. That way, users don't have to modify the auto-generated import statement, or keep a reference to the CodePush instance in their MainActivity.

### DIFF
--- a/README.md
+++ b/README.md
@@ -297,12 +297,7 @@ After installing the plugin and syncing your Android Studio project with Gradle,
     import com.microsoft.codepush.react.CodePush;
 
     public class MainActivity extends ReactActivity {
-        // 2. Define a private field to hold the CodePush runtime instance
-        private CodePush _codePush;
-
-        ...
-
-        // 3. Override the getJSBundleFile method in order to let
+        // 2. Override the getJSBundleFile method in order to let
         // the CodePush runtime determine where to get the JS
         // bundle location from on each app start
         @Override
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -312,41 +307,20 @@ After installing the plugin and syncing your Android Studio project with Gradle,
...
-    public class MainActivity extends ReactActivity {
-        ...
-        @Override
-        protected List<ReactPackage> getPackages() {
-            return Arrays.<ReactPackage>asList(
-                ...
-                new CodePushReactPackage() // 2. Remove this line
-                ...
+                new MainReactPackage(), 
+                new CodePush("0dsIDongIcoH0mqAmoR0CYb5FhBZNy1w4Bf-l", this, BuildConfig.DEBUG)
```
dtivel: Should this be a placeholder, like "your deployment key goes here"?

README.md
```diff
@@ -312,41 +307,20 @@ After installing the plugin and syncing your Android Studio project with Gradle,
...
             );
         }
+
         ...
     }
     ```
 
-3. Ensure that the `android.defaultConfig.versionName` property in your `android/app/build.gradle` file is set to a semver compliant value. Note that if the value provided is missing a patch version, the CodePush server will assume it is `0`, i.e. `1.0` will be treated as `1.0.0`.
+2. Ensure that the `android.defaultConfig.versionName` property in your `android/app/build.gradle` file is set to a semver compliant value. Note that if the value provided is missing a patch version, the CodePush server will assume it is `0`, i.e. `1.0` will be treated as `1.0.0`.
```
dtivel: semver-compliant

android/app/src/main/java/com/microsoft/codepush/react/CodePush.java
```diff
@@ -100,6 +101,11 @@ public CodePush(String deploymentKey, Activity mainActivity, boolean isDebugMode
         }
 
         initializeUpdateAfterRestart();
+        if (currentInstance != null) {
+            CodePushUtils.log("More than one CodePush instances has been initialized, please use the non-static method codePush.getBundleUrlInternal() to get the correct bundleURL for a particular instance.");
```
dtivel: "More than one CodePush instance has been initialized.  Please use the instance method codePush.getBundleUrlInternal() to get the correct bundleURL for a particular instance."

dtivel: Why is one CodePush instance insufficient?  Why not make this a singleton?

android/app/src/main/java/com/microsoft/codepush/react/CodePushNotInitializedException.java
```diff
@@ -0,0 +1,12 @@
+package com.microsoft.codepush.react;
+
+public class CodePushNotInitializedException extends RuntimeException {
```
dtivel: final?

android/app/src/main/java/com/microsoft/codepush/react/CodePush.java
geof90: We know of one user eg. https://github.com/Microsoft/react-native-code-push/pull/255 that is thinking of having multiple ReactActivity in their app. While we don't currently support this scenario now, I wanted to keep the architecture flexible for enabling that in the future.

PR #261, opened 2016-03-23
Problem with a quote in Podfile section

xcarpentier:
Problem with a quote in Podfile section


lostintangent:
@xcarpentier Good catch! Thanks for fixing this.

PR #264, opened 2016-03-25
Introduce simpler getBundleUrl

lostintangent:
This PR simply introduces a version of the `CodePush.getBundleUrl` method that defaults to assuming you're JS bundle file is named `index.android.bundle`, instead of requiring you to explicitly specify it. This provides parity with the Objective-C API, which already has a no-parameter method which assumes your JS bundle is called `main.jsbundle`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -771,7 +771,7 @@ Constructs the CodePush client runtime and includes methods for integrating Code
     
 ##### Methods
 
-- __getBundleUrl()__ - Returns the path to the most recent version of your app's JS bundle file, assuming that the resource name is `index.android.bundle`. If your app is using a different bundle name, then use the overloaded verison of this method which allows specifying it. This method has the same resolution behavior as the Objective-C equivalent described above.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -771,6 +771,8 @@ Constructs the CodePush client runtime and includes methods for integrating Code
     
 ##### Methods
 
+- __getBundleUrl()__ - Returns the path to the most recent version of your app's JS bundle file, assuming that the resource name is `index.android.bundle`. If your app is using a different bundle name, then use the overloaded verison of this method which allows specifying it. This method has the same resolution behavior as the Objective-C equivalent described above.
```
shishirx34: overloaded __version__

android/app/src/main/java/com/microsoft/codepush/react/CodePush.java
```diff
@@ -135,6 +135,10 @@ private long getBinaryResourcesModifiedTime() {
         }
     }
 
+    public static String getBundleUrl() {
+        return getBundleUrl("index.android.bundle");
```
shishirx34: not super important, but may be define `"index.android.bundle"` as a static field above.

shishirx34:
Looks good! :shipit:

geof90:
Looks awesome

PR #270, opened 2016-03-31
Podspec version bump

mrspeaker:


lostintangent:
@mrspeaker Thanks for catching this!

Issue #273, opened 2016-04-01
unable to link app after installing w/ npm and cocoapods


charlieschwabacher:
Hi - appreciate all your work on code-push!  I ran into an issue when installing into my react-native app.  Starting from a working project, I added `"react-native-code-push": "^1.9.0-beta"` to `package.json`, and `pod 'CodePush', :path => '../node_modules/react-native-code-push'` to `Podfile`, and then ran `npm install` and `pod install`.  After making only these changes, the project fails to link w/ the following error:

```
duplicate symbol _main in:
    /Users/username/Library/Developer/Xcode/DerivedData/MyProject-aqrwunwlhomiwrdpixfnqxvvzbwt/Build/Intermediates/MyProject.build/Test-iphonesimulator/MyProject.build/Objects-normal/x86_64/main.o
    /Users/username/Library/Developer/Xcode/DerivedData/MyProject-aqrwunwlhomiwrdpixfnqxvvzbwt/Build/Products/Test-iphonesimulator/libCodePush.a(pwd2key.o)
ld: 1 duplicate symbol for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

Have you heard of anything like this before?  Or is there a step that I am missing?

geof90:
`pwd2key` belongs to the SSZipArchive library that we are shipping together with this plugin. My guess is that one of your other plugins in the project are also using the SSZipArchive library, resulting in a duplicate symbol. 

There used to be a similar issue https://github.com/Microsoft/react-native-code-push/issues/241 reported here previously, except that user was linking the project manually (not using CocoaPods).

If you followed the workaround in that issue, and manually linked our plugin instead of using CocoaPods, you should be able to get this to work. I am not too sure about the best workaround from a CocoaPods workflow, but I will be happy to investigate this more tomorrow if you are still blocked. My guess is that if you removed the SSZipArchive files from our plugin in the node_modules folder, you should also be able to get this to just work, with some additional effort.

charlieschwabacher:
[...]

PR #274, opened 2016-04-03
remove test code from aes pwd2key.c

charlieschwabacher:
This addresses https://github.com/Microsoft/react-native-code-push/issues/273 - the test code in this file causes an error by defining a duplicate symbol `main` when running an app in the test scheme.


lostintangent:
Looks good to me. I'll wait for @geof90 to sign off as well before merging, and I'll do some basic smoke testing as well. Thanks for submitting this!

geof90:
This change looks good to me.

ios/CodePush/SSZipArchive/aes/pwd2key.c
```diff
@@ -101,93 +101,3 @@ void derive_key(const unsigned char pwd[],  /* the PASSWORD     */
...
-                printf("\n");
-            printf("0x%02x%02x%02x%02x ", key[j], key[j + 1], key[j + 2], key[j + 3]);
-        }
-        printf(j < key_len ? " ... \n" : "\n");
-    }
-    printf("\n");
-    return 0;
-}
-
-#if defined(__cplusplus)
```
lostintangent: Does this #if block need to actually be moved outside of the `TEST` section? I just noticed that it seems to add the closing brace for the `{` added at the very top of this file.

PR #275, opened 2016-04-05
Android reload refactor

lostintangent:
This PR implements a new app reloading solution on Android that doesn't require restarting the `Activity`, and therefore, is less obtrusive to the user-experience. It works based on private reflection, and therefore, in order to be resilient to breaking changes, it will fall back to the old `Activity` restart logic if it fails for any reason.

After making this change, the `initializeUpdateAfterRestart` method needs to be called whenever the module is initialized, not when the `CodePush` class is constructed, since the `CodePush` class would only be constructed on the app load. Therefore, I moved the call to this method to the module's `initialize` method, which will be appropriately triggered on app start as well as after a programmatic restart. This change ends up mirroring the way that iOS already works.

Finally, I re-organized the methods a little bit in the module, which is why there is some churn on unrelated methods (e.g. `getConstants`, `getName`).

### DIFF
--- a/android/app/src/main/java/com/microsoft/codepush/react/CodePush.java
+++ b/android/app/src/main/java/com/microsoft/codepush/react/CodePush.java
@@ -396,7 +396,7 @@ public class CodePush implements ReactPackage {
             try {
                 // #1) Get the private ReactInstanceManager, which is what includes
                 //     the logic to reload the current React context.
-                Field instanceManagerField = ReactActivity.class.getDeclaredField("mReacInstanceManager");
+                Field instanceManagerField = ReactActivity.class.getDeclaredField("mReactInstanceManager");
                 instanceManagerField.setAccessible(true); // Make a private field accessible
                 final Object instanceManager = instanceManagerField.get(mainActivity);
 
--- a/android/app/src/main/java/com/microsoft/codepush/react/CodePush.java
+++ b/android/app/src/main/java/com/microsoft/codepush/react/CodePush.java
@@ -152,7 +152,7 @@ public class CodePush implements ReactPackage {
         return currentInstance.getBundleUrlInternal(assetsBundleFileName);
     }
     
-    private String getBundleUrlInternal(String assetsBundleFileName) {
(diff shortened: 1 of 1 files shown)


android/app/src/main/java/com/microsoft/codepush/react/CodePush.java
```diff
@@ -147,8 +151,8 @@ public static String getBundleUrl(String assetsBundleFileName) {
 
         return currentInstance.getBundleUrlInternal(assetsBundleFileName);
     }
-
-    public String getBundleUrlInternal(String assetsBundleFileName) {
+    
+    private String getBundleUrlInternal(String assetsBundleFileName) {
```
geof90: Should we keep this `public`? Otherwise, the logging [here](https://github.com/Microsoft/react-native-code-push/pull/275/files#diff-7a4664ad8b157a07a7a6cf0a7de26861R110) no longer makes sense.

android/app/src/main/java/com/microsoft/codepush/react/CodePush.java
```diff
@@ -356,13 +360,74 @@ public void clearUpdates() {
...
 
-        private void loadBundle() {
+        @Override
+        public String getName() {
+            return "CodePush";
+        }
+        
+        @Override
+        public void initialize() {
+            currentInstance.initializeUpdateAfterRestart();
```
geof90: Should we use `CodePush.this` instead?

geof90:
LGTM!

android/app/src/main/java/com/microsoft/codepush/react/CodePush.java
lostintangent: Good call, I'll fix this.

android/app/src/main/java/com/microsoft/codepush/react/CodePush.java
lostintangent: Yeah that's a much better solution.

PR #279, opened 2016-04-06
Adding RNPM config

lostintangent:
This change primarily introduces RNPM config to the module which will enable the following scenarios:

1. Generating the right constructor call within the `MainActivity` file for Android projects, which sources the deployment key from the end user via a prompt 

2. Automatically linking the `libz` dependency for iOS projects. This feature isn't enabled yet in RNPM, so I'm adding this as an optimistic feature that will light up soon

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,12 +18,9 @@
   "dependencies": {
     "code-push": "1.8.0-beta"
   },
-  "peerDependencies": {
-    "react-native": ">=0.19.0"
-  },
   "rnpm": {
       "android": {
-          "packageInstance": "new CodePush('${androidDeployment}', this, BuildConfig.DEBUG)"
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -18,7 +18,20 @@
...
+      "android": {
+          "packageInstance": "new CodePush('${androidDeployment}', this, BuildConfig.DEBUG)"
+      },
+      "ios": {
+        "sharedLibraries": ["libz"]  
+      },
+      "params": [{
+        "type": "input",
+        "name": "androidDeploymentKey",
+        "message": "What is your CodePush deployment key for Android"
```
shishirx34: may be `?` in the end?

shishirx34:
LGTM

package.json
lostintangent: The UI will actually display a `?` already.

PR #280, opened 2016-04-06
Fix Android clearing dev cache on reload

lostintangent:
When I changed the way we reload on Android, I broke the logic that clears the cached JS bundle in debug mode. I fixed this by clearing the cache when `loadBundle` is called, which I actually think makes more sense then where we were previously doing it.


geof90:
LGTM!

PR #281, opened 2016-04-08
Update `android` part when using `rnpm`

grabbou:
Hey,

So recent PR (#279) added params to be used with `rnpm` - that means we no longer generate an empty `new CodePush`, but ask user for development token instead as well.

I am thinking we might either remove that comment or mention `if you are using RNPM < 1.6`

What you think?


lostintangent:
LGTM! I'm cool with removing this line.

PR #283, opened 2016-04-08
Updating alert adapter

lostintangent:
This PR removes our dependency on `AlertIOS` on iOS since the standard `Alert` module will work fine. Additionally, I updated some of the syntax to match the ES6 conventions we use in the rest of the repo.


geof90:
LGTM!

PR #285, opened 2016-04-12
remote "-beta" from podspec -- itunesconnect error

triage:
Hi there
Tried submitting a build w/ react-native-code-push (installed via Cocoapods, tag: 1.1.0-beta) to iTunesConnect, and got an error:
`[Transporter Error Output]: ERROR ITMS-90060: "This bundle is invalid. The value for key CFBundleShortVersionString '1.10.1-beta' in the Info.plist file must be a period-separated list of at most three non-negative integers."`

Renaming the version in the podspec to remove `-beta` resolved this error.

### DIFF
--- a/CodePush.podspec
+++ b/CodePush.podspec
@@ -6,7 +6,7 @@ Pod::Spec.new do |s|
   s.author              = 'Microsoft Corporation'
   s.license             = 'MIT'
   s.homepage            = 'http://microsoft.github.io/code-push/'
-  s.source              = { :git => 'https://github.com/Microsoft/react-native-code-push.git', :tag => "v#{s.version}" }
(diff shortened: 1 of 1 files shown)


CodePush.podspec
```diff
@@ -1,7 +1,7 @@
 Pod::Spec.new do |s|
 
   s.name                = 'CodePush'
-  s.version             = '1.10.1-beta'
```
lostintangent: Since the `s.source` property specifies the Git tag using the `s.version` property, would it cause any issues for folks if we removed the `-beta` suffix here since the actual Git tag would include it?

triage: thank you @lostintangent ... I've updated this to append '-beta' to the tag. now the tags will be maintained but the version that shows up in the Info.plist will be in the proper format.

Issue #286, opened 2016-04-13
java.lang.VerifyError using latest version (v1.10.1-beta)


jamieathans:
I just tried upgrading to v1.10.1-beta and now the app crashes on startup with the following:

*********************************
`637-31650/? I/dalvikvm: Could not find method android.webkit.CookieManager.setCookie, referenced from method com.facebook.react.modules.network.ForwardingCookieHandler.addCookieAsync
04-13 10:40:29.924 31637-31650/? I/dalvikvm: Could not find method android.webkit.CookieManager.removeAllCookies, referenced from method com.facebook.react.modules.network.ForwardingCookieHandler.clearCookiesAsync
04-13 10:40:29.924 31637-31650/? I/dalvikvm: Could not find method android.webkit.CookieManager.flush, referenced from method com.facebook.react.modules.network.ForwardingCookieHandler$CookieSaver.flush
04-13 10:40:29.929 31637-31650/? W/dalvikvm: VFY:  rejected Lcom/microsoft/codepush/react/CodePush$CodePushNativeModule;.loadBundle ()V
04-13 10:40:29.929 31637-31650/? W/dalvikvm: VFY:  rejected Lcom/microsoft/codepush/react/CodePush$CodePushNativeModule;.loadBundle ()V
04-13 10:40:29.929 31637-31650/? W/dalvikvm: Verifier rejected class Lcom/microsoft/codepush/react/CodePush$CodePushNativeModule;
04-13 10:40:29.934 31637-31650/? E/AndroidRuntime: FATAL EXCEPTION: AsyncTask #1
                                                   java.lang.RuntimeException: An error occured while executing doInBackground()
                                                       at android.os.AsyncTask$3.done(AsyncTask.java:299)
                                                       at java.util.concurrent.FutureTask.finishCompletion(FutureTask.java:352)
                                                       at java.util.concurrent.FutureTask.setException(FutureTask.java:219)
                                                       at java.util.concurrent.FutureTask.run(FutureTask.java:239)
                                                       at android.os.AsyncTask$SerialExecutor$1.run(AsyncTask.java:230)
                                                       at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1080)
                                                       at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:573)
                                                       at java.lang.Thread.run(Thread.java:841)
[...]

PR #288, opened 2016-04-14
Removing use of ReflectiveOperationException

lostintangent:
The `ReflectiveOperationException` class was introduced in Android KitKat (API level 19), and therefore, we can't catch exceptions of that type and expect the code to work on older versions of Android. Since we want to support JellyBean+, this PR replaces the use of this exception type in order to resolve issue #286.

`ReflectiveOperationException` was the nearest common ancestor of all of the checked exceptions that the reflection APIs used in `loadBundle` can throw, and therefore, once we aren't catching that, we would either need to catch a handful of individual exception types, or just add a catch all. Since we want to fall back on the legacy restart logic in all failure cases, at this point, I feel more comfortable with just catching `Exception`, but am open to other opinions.

### DIFF
--- a/android/app/src/main/java/com/microsoft/codepush/react/CodePush.java
+++ b/android/app/src/main/java/com/microsoft/codepush/react/CodePush.java
@@ -30,7 +30,6 @@ import org.json.JSONArray;
 import org.json.JSONException;
 import org.json.JSONObject;
 
-import java.lang.Exception;
 import java.lang.reflect.Field;
 import java.lang.reflect.Method;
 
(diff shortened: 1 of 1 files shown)


android/app/src/main/java/com/microsoft/codepush/react/CodePush.java
```diff
@@ -30,7 +30,7 @@
 import org.json.JSONException;
 import org.json.JSONObject;
 
-import java.lang.ReflectiveOperationException;
+import java.lang.Exception;
```
geof90: I think we can get rid of this, it should be imported by default.

geof90:
LGTM!

android/app/src/main/java/com/microsoft/codepush/react/CodePush.java
lostintangent: Ah right, good catch.

PR #294, opened 2016-04-16
codePush.getUpdateMetadata

lostintangent:
This PR implements a new JS API method called `getUpdateMetadata` which allows an app to request the metadata specifically for either the currently running update, a pending update or the latest update (regardless if it is pending or not). The primary motivation behind this is to provide a deterministic way to retrieve the metadata of the currently running update, regardless if there is a pending update or not (which is currently a flaw with `getCurrentPackage`).

Example usages:

```javascript
import codePush, { UpdateState } from "react-native-code-push";

// Retrieve the metadata of the running update. 
// This is the primary benefit of this new method, since `getCurrentPackage`
// will return a pending update if there is one, so you can no longer retrieve
// the metadata of the currently running update
codePush.getUpdateMetadata(); 

// Same as above, since `RUNNING` is the default update state
codePush.getUpdateMetadata(UpdateState.RUNNING); 

// Retrieve the metadata of the pending update, or null if there isn't a pending update. 
// This replaces the need to call `getCurrentPackage` and then check `sPending` simply
// in order to determine if there is a pending update or not
codePush.getUpdateMetadata(UpdateState.PENDING); 

// Retrieve the metadata of the latest update, regardless if it's pending or not. 
// This behaves in the same way that `getCurrentPackage` currently does.
codePush.getUpdateMetadata(UpdateState.LATEST);
```

Note that I chose not to remove `getCurrentPackage` since I'm a little cautious about introducing breaking changes moving forward. I think we can remove it from our docs, and generally not prescribe it anymore, but it seemed largely harmless to me to simply make `getCurrentPackage` an alias for calling `getUpdateMetadata(codePush.UpdateState.LATEST)`. Additionally, I left the `isPending` property on the `*Package` objects, for three reasons:

1. Backwards compatibility (primary motivation)
2. It enables calling `getUpdateMetadata` with the `LATEST` state, and differentiating between whether the update is pending or not
3. Regardless which `UpdateState` you request, if you pass the `Package` object as an argument to another function, that function may want/need to understand if it's pending or not.

### Misc. Changes

1. Removing the logic that deletes the status file in the `clearUpdates` method of the Java/Obj-C `CodePushPackage` classes, since it actually isn't needed (deleting the entire CodePush folder will delete this file as well)

2. I introduced an alias for `restartApp` called `restartApplication`, which is the name that Cordova uses, and is admittedly more consistent with `notifyApplicationReady`

### Misc. Changes (iOS)

1. Replacing occurrences of `NULL` with `nil` (consistency!)

2. Extracting the `app.json` literal into a static const within the `CodePushPackage` class (Android already had this)

3. Re-organizaing the files in our `xcodeproj` alphabetically (this is the reason for some of the churn in this file)

### DIFF
--- a/CodePush.js
+++ b/CodePush.js
@@ -387,52 +387,52 @@ let CodePush;
 // and therefore, it doesn't make sense initializing 
 // the JS interface when it wouldn't work anyways.
 if (NativeCodePush) {
-    CodePush = {
-        AcquisitionSdk: Sdk,
-        checkForUpdate,
-        getConfiguration,
-        getCurrentPackage,
-        getUpdateMetadata,
-        log,
-        notifyApplicationReady,
-        restartApp,
-        restartApplication: restartApp,
-        setUpTestDependencies,
-        sync,
-        InstallMode: {
-            IMMEDIATE: NativeCodePush.codePushInstallModeImmediate, // Restart the app immediately
-            ON_NEXT_RESTART: NativeCodePush.codePushInstallModeOnNextRestart, // Don't artificially restart the app. Allow the update to be "picked up" on the next app restart
-            ON_NEXT_RESUME: NativeCodePush.codePushInstallModeOnNextResume // Restart the app the next time it is resumed from the background
-        },
-        SyncStatus: {
-            CHECKING_FOR_UPDATE: 0,
-            AWAITING_USER_ACTION: 1,
-            DOWNLOADING_PACKAGE: 2,
-            INSTALLING_UPDATE: 3,
-            UP_TO_DATE: 4, // The running app is up-to-date
-            UPDATE_IGNORED: 5, // The app had an optional update and the end-user chose to ignore it
-            UPDATE_INSTALLED: 6, // The app had an optional/mandatory update that was successfully downloaded and is about to be installed.
-            SYNC_IN_PROGRESS: 7, // There is an ongoing "sync" operation in progress.
-            UNKNOWN_ERROR: -1
-        },
-        UpdateState: {
-          RUNNING: NativeCodePush.codePushUpdateStateRunning,
-          PENDING: NativeCodePush.codePushUpdateStatePending,
-          LATEST: NativeCodePush.codePushUpdateStateLatest
-        },
-        DEFAULT_UPDATE_DIALOG: {
-            appendReleaseDescription: false,
-            descriptionPrefix: " Description: ",
-            mandatoryContinueButtonLabel: "Continue",
-            mandatoryUpdateMessage: "An update is available that must be installed.",
-            optionalIgnoreButtonLabel: "Ignore",
-            optionalInstallButtonLabel: "Install",
-            optionalUpdateMessage: "An update is available. Would you like to install it?",
-            title: "Update available"
-        }
+  CodePush = {
+    AcquisitionSdk: Sdk,
+    checkForUpdate,
+    getConfiguration,
+    getCurrentPackage,
+    getUpdateMetadata,
+    log,
+    notifyAppReady: notifyApplicationReady,
+    notifyApplicationReady,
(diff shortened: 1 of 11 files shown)


CodePush.js
```diff
@@ -100,12 +100,16 @@ const getConfiguration = (() => {
...
-      localPackage.failedInstall = await NativeCodePush.isFailedUpdate(localPackage.packageHash);
-      localPackage.isFirstRun = await NativeCodePush.isFirstRun(localPackage.packageHash);
+  return await getUpdateMetadata(CodePush.UpdateState.LATEST);
+}
+
+async function getUpdateMetadata(updateState) {
+  const updateMetadata = await NativeCodePush.getUpdateMetadata(updateState || CodePush.UpdateState.RUNNING);
+  if (updateMetadata) {
+      updateMetadata.failedInstall = await NativeCodePush.isFailedUpdate(updateMetadata.packageHash);
+      updateMetadata.isFirstRun = await NativeCodePush.isFirstRun(updateMetadata.packageHash);
```
geof90: I think we need to agree on whether we are using 2 spaces or 4 spaces for tabs at some point :) The above are using 4 spaces, but everywhere else is 2 spaces.

lostintangent: Totally agreed :) I personally prefer 4 spaces, but for now, I'll make sure we use 2 for consistency.

README.md
```diff
@@ -417,33 +417,41 @@ codePush.checkForUpdate()
...
-codePush.getCurrentPackage()
-.then((update) => {
-    // If the current app "session" represents the first time
-    // this update has run, and it had a description provided
-    // with it upon release, let's show it to the end user
-    if (update.isFirstRun && update.description) {
-        // Display a "what's new?" modal
+// Check if there is currently a CodePush update running, and if
+// so, register it with the HockeyApp SDK (https://github.com/slowpath/react-native-hockeyapp)
+//  so that crash reports will correctly display the JS bundle version the user was running.
```
geof90: There's an extra leading space here is that intentional?

lostintangent: It was probably an artifact of me splitting up a long comment line into two.

CodePush.js
geof90: I wouldn't mind if you changed everything to 4!

android/app/src/main/java/com/microsoft/codepush/react/CodePush.java
```diff
@@ -481,31 +487,53 @@ public void getConfiguration(Promise promise) {
...
+                        promise.resolve("");
+                    } else if (updateState == CodePushUpdateState.RUNNING.getValue() && currentUpdateIsPending) {
+                        // The caller wants the running update, but the current
+                        // one is pending, so we need to grab the previous.
+                        promise.resolve(codePushPackage.getPreviousPackage());
+                    } else {
+                        // The current package satisfies the request:
+                        // 1) Caller wanted a pending, and there is a pending update
+                        // 2) Caller wanted the running update, and there isn't a pending
+                        // 3) Calers wants the latest update, regardless if it's pending or not
```
geof90: Caller

CodePush.js
```diff
@@ -409,6 +415,11 @@ if (NativeCodePush) {
             SYNC_IN_PROGRESS: 7, // There is an ongoing "sync" operation in progress.
             UNKNOWN_ERROR: -1
         },
+        UpdateState: {
+          RUNNING: NativeCodePush.codePushUpdateStateRunning,
+          PENDING: NativeCodePush.codePushUpdateStatePending,
+          LATEST: NativeCodePush.codePushUpdateStateLatest
+        },
```
geof90: These too (here it's 2 spaces, but the CodePush object is defined with 4-spaced tabs)

geof90:
Looks awesome!

PR #298, opened 2016-04-19
Fixing restart in debug

lostintangent:
My change to the Android reload logic regressed the behavior that cleared the JS bundle cache when the  app is "naturally" restarted, as opposed to synthetically. This PR fixes it by simply attempting the clear the   cache when the app is first started if neccessary.


geof90:
LGTM!

Issue #297, opened 2016-04-18
Current package targeting older app version prevents update


johnwayner:
I think that perhaps [this change](https://github.com/Microsoft/react-native-code-push/commit/89d0ffaeb88e609671e83ab4acb0d47a77122289#diff-fdfb3b2552c5b0ea96bacdf49a2581c7L43) may have caused a bug.

If the application is currently using a package targeting an earlier version of the application, any updates that target the application's real version are ignored.

This happens because the update check will use the current package's target app version (localPackage.appVersion) instead of the real app version (config.appVersion).

geof90:
@johnwayner thanks for reporting!

If the package targets an earlier version of the application, it should not be run and be deleted due to [this change](https://github.com/Microsoft/react-native-code-push/commit/89d0ffaeb88e609671e83ab4acb0d47a77122289#diff-7a4664ad8b157a07a7a6cf0a7de26861R165). Have you actually encountered this problem yourself?

johnwayner:
I think I may have had this issue because I was calling sync even in my development environment (using the simulator).  When I updated the app version it would continue to use the downloaded package for the previous version.

I also misused the app version specification for the deployments.  I was thinking I was specifying a minimum version rather than a npm-style version specification.

I'm fine with closing this out as user error.  It just seemed a bit suspicious to request updates using the previous package's appVersion, so I thought I'd bring it up.

geof90:
No worries, it's a perfectly legitimate scenario to be calling sync and testing CodePush with the simulator. 

The fact that updates stick around when you deploy a new binary to the simulator is by design, because we think it is a bad experience for users to keep having to re-download the same update every time a new binary is deployed. Does this behaviour make sense to you?

Also, if you deploy a new binary to the simulator **with a newer/different appVersion**, the original CodePush updates on disk should be cleared. The SDK would not re-download the CodePush updates because they would be targeting an older version of the binary. I will be double-testing this to ensure that this indeed the behaviour, but I just want to ensure that you are indeed seeing this behaviour, or are you seeing something different?

> When I updated the app version it would continue to use the downloaded package for the previous version.

This doesn't sound right. If you updated the app version and deployed a new binary to the simulator, it should ignore the updates present on disk and load the one in the binary.

lostintangent:
[...]

PR #300, opened 2016-04-19
Check and clear updates if needed during init if running JS bundle from dev server

geof90:
This fixes an issue brought up in https://github.com/Microsoft/react-native-code-push/issues/297. If the app is configured to run the version from the dev server, `[CodePush bundleURL]` is never called, and outdated updates are hence never cleared. This causes `checkForUpdate` requests to still use outdated updates which could be installed under a different appVersion from the actual binary, and updates that actually apply to the binary will not be acquired.

### DIFF
--- a/ios/CodePush/CodePush.m
+++ b/ios/CodePush/CodePush.m
@@ -173,6 +173,29 @@ static NSString *bundleResourceName = @"main";
 @synthesize methodQueue = _methodQueue;
 
 /*
+ * This method is used to clear updates that are installed
+ * under a different app version and hence don't apply anymore,
+ * during a debug run configuration and when the bridge is 
+ * running the JS bundle from the dev server.
+ */
+- (void)clearDebugUpdates
+{
+    dispatch_async(dispatch_get_main_queue(), ^{
+        if ([_bridge.bundleURL.scheme hasPrefix:@"http"]) {
+            NSError *error;
+            NSString *binaryAppVersion = [[CodePushConfig current] appVersion];
+            NSDictionary *currentPackageMetadata = [CodePushPackage getCurrentPackage:&error];
+            if (currentPackageMetadata) {
+                NSString *packageAppVersion = [currentPackageMetadata objectForKey:AppVersionKey];
+                if (![binaryAppVersion isEqualToString:packageAppVersion]) {
+                    [CodePush clearUpdates];
+                }
+            }
+        }
+    });
+}
+
+/*
  * This method is used by the React Native bridge to allow
  * our plugin to expose constants to the JS-side. In our case
  * we're simply exporting enum values so that the JS and Native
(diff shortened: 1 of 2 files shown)


ios/CodePush/CodePush.m
```diff
@@ -247,6 +247,20 @@ - (instancetype)init
  */
 - (void)initializeUpdateAfterRestart
 {
+    dispatch_async(dispatch_get_main_queue(), ^{
+        if ([_bridge.bundleURL.scheme hasPrefix:@"http"]) {
```
shishirx34: does `hasPrefix` do ignoreCase comparison? Do you think the casing could be an issue?

shishirx34:
LGTM

ios/CodePush/CodePush.m
geof90: I don't think it does ignoreCase comparison, but because we have been using this same check elsewhere, e.g. https://github.com/Microsoft/react-native-code-push/blob/clear-updates-init/ios/CodePush/CodePush.m#L346 and it has not ever caused any problems in practice. The provided URL is generated by RN's app template and there is rarely any need for users to modify it or capitalize the scheme. Hence, I would prefer to leave it as-is.

ios/CodePush/CodePush.m
```diff
@@ -247,6 +247,20 @@ - (instancetype)init
  */
 - (void)initializeUpdateAfterRestart
 {
+    dispatch_async(dispatch_get_main_queue(), ^{
```
lostintangent: Should we move this out to another method? Somehing like "clearDebugUpdates" or whatever? It might help keep this core method slim.

ios/CodePush/CodePush.m
```diff
@@ -208,45 +208,64 @@ + (void)ensureBinaryBundleExists
...
+}
+
 /*
  * This method is used when the app is started to either
  * initialize a pending update or rollback a faulty update
  * to the previous version.
  */
 - (void)initializeUpdateAfterRestart
 {
+    [self clearDebugUpdates];
```
lostintangent: Should we actually wrap this in an #ifdef DEBUG statement? That way it's limited to just debug builds? I can't imagine why someone would create a release build that is using the packager, and I'm in favor of optimizing release builds as much as possible.

ios/CodePush/CodePush.m
```diff
@@ -208,45 +208,64 @@ + (void)ensureBinaryBundleExists
...
+
     if (self) {
         [self initializeUpdateAfterRestart];
     }
-    
+
     return self;
 }
 
+- (void)clearDebugUpdates
```
lostintangent: Do we want to keep this pragma group sorted?

ios/CodePush/CodePush.m
```diff
@@ -208,45 +208,64 @@ + (void)ensureBinaryBundleExists
...
+{
+    dispatch_async(dispatch_get_main_queue(), ^{
```
lostintangent: Are there any chances of this not working due to a timing issue? It just seems a little brittle to expect the bridge to be set when this is run. I'd almost be in favor of create a setBridge method to hook the exact moment that we know the bridge is available.

geof90: This same technique has been used in the loadBundle method in production with no issues. I am pretty sure it will be set because the bridge is initialized serially but i can double check

lostintangent: LoadBundle is a little different since it's not called as part of the init phase, so it wouldn't be possible for there to be a timing issue. 

That said, I just took a look again at the code in RN that creates modules and this should be fine. I forgot that it will create the module on the main thread if it has an init method, so dispatching this block will indeed always happen after the bridge property is set.

Thanks for humoring me!

lostintangent:
I'd vote to do the DEBUG check, but I'm cool either way. LGTM! Did you want to bump the version to 1.10.3 and I'll release it later this afternoon?

PR #308, opened 2016-04-25
Syncing Podspec version with package.json

lostintangent:
This updates our `Podspec` to pull its version from the `package.json` file. This way we can simply update a single version whenever we rev the module. I was inspired to do this after reading the `Podspec` of the `react-native-video` module, and realized how useful this is since we've occasionally forgotten to version the `Podspec` when shipping an update to NPM.

The "-beta" stripping logic is meant to maintain the fix for iTunes Connect not supporting frameworks whose version name includes a pre-release tag. I tested this change out in a native iOS project and it works beautifully :)

### DIFF
--- a/CodePush.podspec
+++ b/CodePush.podspec
@@ -1,8 +1,12 @@
+require 'json'
+
+package = JSON.parse(File.read(File.join(__dir__, 'package.json')))
+
 Pod::Spec.new do |s|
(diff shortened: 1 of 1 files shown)


geof90:
nice!! now we don't have to keep updating this. LGTM!

Issue #309, opened 2016-04-25
[Android] NullPointerException in CodePushUtils


dhrrgn:
Seeing a a good number of crash reports from CodePushUtils.

```
Fatal Exception: java.lang.RuntimeException: Unable to start activity ComponentInfo{com.tackk.squad/com.tackk.squad.MainActivity}: java.lang.NullPointerException
       at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:2449)
       at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:2509)
       at android.app.ActivityThread.access$900(ActivityThread.java:172)
       at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1308)
       at android.os.Handler.dispatchMessage(Handler.java:102)
       at android.os.Looper.loop(Looper.java:146)
       at android.app.ActivityThread.main(ActivityThread.java:5694)
       at java.lang.reflect.Method.invokeNative(Method.java)
       at java.lang.reflect.Method.invoke(Method.java:515)
       at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:1291)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1107)
       at dalvik.system.NativeStart.main(NativeStart.java)
Caused by java.lang.NullPointerException
       at com.microsoft.codepush.react.CodePushUtils.tryGetString(CodePushUtils.java:219)
       at com.microsoft.codepush.react.CodePushPackage.getCurrentPackageBundlePath(CodePushPackage.java:104)
       at com.microsoft.codepush.react.CodePush.getBundleUrlInternal(CodePush.java:150)
       at com.microsoft.codepush.react.CodePush.getBundleUrl(CodePush.java:141)
       at com.microsoft.codepush.react.CodePush.getBundleUrl(CodePush.java:133)
       at com.tackk.squad.MainActivity.getJSBundleFile(MainActivity.java:53)
       at com.facebook.react.ReactActivity.createReactInstanceManager(ReactActivity.java:108)
       at com.facebook.react.ReactActivity.onCreate(ReactActivity.java:140)
       at android.app.Activity.performCreate(Activity.java:5541)
       at android.app.Instrumentation.callActivityOnCreate(Instrumentation.java:1093)
       at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:2413)
[...]

PR #311, opened 2016-04-25
Fix rollback android

geof90:
Please ignore all the whitespace changes: https://github.com/Microsoft/react-native-code-push/pull/311/files?w=1

This PR: 

1. Fixes a regression with rollbacks. They no longer work after we migrated the new reflection based bundle loading strategy because `CodePushNativeModule.initialize()` gets called after the bundle is already loaded, which means that any rollbacks that happen would leave the app in a limbo state where it loads up a bundle that no longer exists on disk later.

2. Fixes https://github.com/Microsoft/react-native-code-push/issues/309 by checking that the current package is not null. It's a temporary fix before we find the root cause as to why the status file would have a current hash but getCurrentPackage() returns null in the first place.

3. Removes all instances of throwing CodePushUnknownException or CodePushMalformedDataException as runtime exceptions, as the plugin should not ever be causing crashes on the host app, even though we have a high amount of confidence those exceptions should never be thrown.


lostintangent:
LGTM!

shishirx34:
LGTM

PR #314, opened 2016-04-26
Fixing isFirstRun on Android

lostintangent:
This change ensures that `isFirstRun` is set to `false` after a rollback happens (which is how iOS behaves). Now that we call `initializeUpdateAfterRestart` in both the cold start and restart scenarios, I'm using it to manage the state for `isFirstRun`. Now, `didUpdate` is always reset to `false` every time the app is restarted for any reason, and it is only ever set to `true` is an update is actually running for the first time. Therefore, if a rollback happens, `didUpdate` would be reset to `false` and not set to `true` since there wouldn't be a pending update.


geof90:
LGTM!

PR #316, opened 2016-04-27
Improve download progress perf

### LINKED ISSUE #303: Downloading in the mainThread
When I want to the second page，why do I have to wait for the download is complete。It is not asynchronous ？

geof90:
This PR fixes #303 by mitigating the perf impact caused by the firing of download progress events.

1. If no download progress handler is passed to the call to CodePush.sync, the native side will not send the progress events.

2. Throttles the firing of the events to be equal or less than the UI framerate.

3. Handle the download progress event handling on the methodQueue instead of the main thread (iOS)

### DIFF
--- a/android/app/src/main/java/com/microsoft/codepush/react/CodePush.java
+++ b/android/app/src/main/java/com/microsoft/codepush/react/CodePush.java
@@ -238,7 +238,7 @@ public class CodePush implements ReactPackage {
         // Reset the state which indicates that
         // the app was just freshly updated.
         didUpdate = false;
-        
+
         JSONObject pendingUpdate = getPendingUpdate();
         if (pendingUpdate != null) {
             try {
@@ -253,7 +253,7 @@ public class CodePush implements ReactPackage {
                     // There is in fact a new update running for the first
                     // time, so update the local state to ensure the client knows.
                     didUpdate = true;
-                    
+
                     // Mark that we tried to initialize the new update, so that if it crashes,
                     // we will know that we need to rollback when the app next starts.
                     savePendingUpdate(pendingUpdate.getString(PENDING_UPDATE_HASH_KEY),
@@ -449,7 +449,7 @@ public class CodePush implements ReactPackage {
                         WritableMap mutableUpdatePackage = CodePushUtils.convertReadableMapToWritableMap(updatePackage);
                         mutableUpdatePackage.putString(BINARY_MODIFIED_TIME_KEY, "" + getBinaryResourcesModifiedTime());
                         codePushPackage.downloadPackage(mutableUpdatePackage, CodePush.this.assetsBundleFileName, new DownloadProgressCallback() {
-                            private boolean nextFrameBusy = false;
+                            private boolean hasScheduledNextFrame = false;
                             private DownloadProgress latestDownloadProgress = null;
 
                             @Override
@@ -459,11 +459,11 @@ public class CodePush implements ReactPackage {
                                 }
 
                                 this.latestDownloadProgress = downloadProgress;
-                                if (nextFrameBusy) {
+                                if (hasScheduledNextFrame) {
                                     return;
                                 }
 
-                                nextFrameBusy = true;
+                                hasScheduledNextFrame = true;
                                 mainActivity.runOnUiThread(new Runnable() {
                                     @Override
                                     public void run() {
@@ -473,7 +473,7 @@ public class CodePush implements ReactPackage {
                                                 getReactApplicationContext()
                                                         .getJSModule(DeviceEventManagerModule.RCTDeviceEventEmitter.class)
                                                         .emit(DOWNLOAD_PROGRESS_EVENT_NAME, latestDownloadProgress.createWritableMap());
-                                                nextFrameBusy = false;
+                                                hasScheduledNextFrame = false;
                                             }
                                         });
                                     }
--- a/android/app/src/main/java/com/microsoft/codepush/react/CodePush.java
+++ b/android/app/src/main/java/com/microsoft/codepush/react/CodePush.java
@@ -458,26 +458,39 @@ public class CodePush implements ReactPackage {
                                     return;
                                 }
 
-                                this.latestDownloadProgress = downloadProgress;
+                                latestDownloadProgress = downloadProgress;
+                                // If the download is completed, synchronously send the last event.
+                                if (latestDownloadProgress.isCompleted()) {
+                                    dispatchDownloadProgressEvent();
+                                    return;
+                                }
+
                                 if (hasScheduledNextFrame) {
                                     return;
                                 }
 
                                 hasScheduledNextFrame = true;
-                                mainActivity.runOnUiThread(new Runnable() {
+                                getReactApplicationContext().runOnUiQueueThread(new Runnable() {
                                     @Override
                                     public void run() {
                                         ReactChoreographer.getInstance().postFrameCallback(ReactChoreographer.CallbackType.TIMERS_EVENTS, new Choreographer.FrameCallback() {
                                             @Override
                                             public void doFrame(long frameTimeNanos) {
-                                                getReactApplicationContext()
-                                                        .getJSModule(DeviceEventManagerModule.RCTDeviceEventEmitter.class)
-                                                        .emit(DOWNLOAD_PROGRESS_EVENT_NAME, latestDownloadProgress.createWritableMap());
+                                                if (!latestDownloadProgress.isCompleted()) {
+                                                    dispatchDownloadProgressEvent();
+                                                }
+
                                                 hasScheduledNextFrame = false;
                                             }
                                         });
                                     }
                                 });
+                            }
+
+                            public void dispatchDownloadProgressEvent() {
+                                getReactApplicationContext()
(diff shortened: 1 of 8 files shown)


ios/CodePush/CodePushDownloadHandler.m
```diff
@@ -22,12 +24,12 @@ -(void)download:(NSString*)url {
     NSURLRequest *request = [NSURLRequest requestWithURL:[NSURL URLWithString:url]
                                              cachePolicy:NSURLRequestUseProtocolCachePolicy
                                          timeoutInterval:60.0];
-    
+    NSOperationQueue *delegateQueue = [NSOperationQueue new];
```
lostintangent: Would it make more sense to move this above the `[connection setDelegateQueue]` call?

android/app/src/main/java/com/microsoft/codepush/react/CodePush.java
```diff
@@ -439,19 +441,43 @@ public void run() {
...
+                                    return;
+                                }
+
+                                this.latestDownloadProgress = downloadProgress;
+                                if (nextFrameBusy) {
+                                    return;
+                                }
+
+                                nextFrameBusy = true;
+                                mainActivity.runOnUiThread(new Runnable() {
```
lostintangent: Does the call to `postFrameCallback` need to run on the UI thread or can this dispatch call be moved to inside the `doFrame` method as needed?

geof90: postFrameCallback needs to run on the UI thread, otherwise it throws this error "Thread does not have a looper"

ios/CodePush/CodePush.m
```diff
@@ -760,4 +760,29 @@ - (void)applicationWillResignActive
     resolve(nil);
 }
 
+#pragma mark - Methods for handling dispatching of download progress events to JS (Private)
```
lostintangent: How about `#pragma mark - RCTFrameUpdateObserver Methods (Private)`? We use this convention in other areas.

geof90: The reason why I didn't do that is because `sendDownloadProgressDuringNextFrame` is not part of the `RCTFrameUpdateObserver` protocol, its a custom method

ios/CodePush/CodePushPackage.m
```diff
@@ -41,6 +41,7 @@ + (void)downloadAndReplaceCurrentBundle:(NSString *)remoteBundleUrl
 
 + (void)downloadPackage:(NSDictionary *)updatePackage
  expectedBundleFileName:(NSString *)expectedBundleFileName
+             usingQueue:(dispatch_queue_t)usingQueue
```
lostintangent: How about calling this `operationQueue`?

ios/CodePush/CodePushDownloadHandler.m
```diff
@@ -6,12 +6,14 @@ @implementation CodePushDownloadHandler {
 }
 
 - (id)init:(NSString *)downloadFilePath
+usingQueue:(dispatch_queue_t)usingQueue
 progressCallback:(void (^)(long long, long long))progressCallback
 doneCallback:(void (^)(BOOL))doneCallback
 failCallback:(void (^)(NSError *err))failCallback {
     self.outputFileStream = [NSOutputStream outputStreamToFileAtPath:downloadFilePath
                                                               append:NO];
     self.receivedContentLength = 0;
```
lostintangent: This isn't related to this PR, but I feel like we could convert all of these properties into ivars and it would clean up the header a little bit.

ios/CodePush/CodePush.m
geof90: If you're fine with ignoring that nitty detail I cna just go ahead and rename that

ios/CodePush/CodePushDownloadHandler.m
```diff
@@ -22,12 +24,12 @@ -(void)download:(NSString*)url {
...
                                          timeoutInterval:60.0];
-    
     NSURLConnection *connection = [[NSURLConnection alloc] initWithRequest:request
                                                                   delegate:self
                                                           startImmediately:NO];
-    [connection scheduleInRunLoop:[NSRunLoop mainRunLoop]
-                          forMode:NSDefaultRunLoopMode];
+    NSOperationQueue *delegateQueue = [NSOperationQueue new];
+    delegateQueue.underlyingQueue = self.usingQueue;
+    [connection setDelegateQueue:delegateQueue];
```
lostintangent: Now that we're doing this, I wonder if we still need to call `dispatch_async` within the `CodePush` class.

ios/CodePush/CodePush.m
```diff
@@ -760,4 +760,29 @@ - (void)applicationWillResignActive
...
+    [self.bridge.eventDispatcher
+     sendDeviceEventWithName:@"CodePushDownloadProgress"
+     body:@{
+            @"totalBytes":[NSNumber numberWithLongLong:latestExpectedContentLength],
+            @"receivedBytes":[NSNumber numberWithLongLong:latestReceivedConentLength]
+           }];
+    _paused = YES;
+}
+
+- (void)sendDownloadProgressDuringNextFrame:(long long)expectedContentLength
```
lostintangent: How about `updateDownloadProgressForNextFrame`? We're not actually sending every download progress event, since they're being throttled.

ios/CodePush/CodePush.m
lostintangent: I'm cool with it.

ios/CodePush/CodePushDownloadHandler.m
geof90: You're right, we don't! Updated.

android/app/src/main/java/com/microsoft/codepush/react/CodePush.java
```diff
@@ -439,19 +441,43 @@ public void run() {
...
-        public void downloadUpdate(final ReadableMap updatePackage, final Promise promise) {
+        public void downloadUpdate(final ReadableMap updatePackage, final boolean notifyProgress, final Promise promise) {
             AsyncTask<Void, Void, Void> asyncTask = new AsyncTask<Void, Void, Void>() {
                 @Override
                 protected Void doInBackground(Void... params) {
                     try {
                         WritableMap mutableUpdatePackage = CodePushUtils.convertReadableMapToWritableMap(updatePackage);
                         mutableUpdatePackage.putString(BINARY_MODIFIED_TIME_KEY, "" + getBinaryResourcesModifiedTime());
                         codePushPackage.downloadPackage(mutableUpdatePackage, CodePush.this.assetsBundleFileName, new DownloadProgressCallback() {
+                            private boolean nextFrameBusy = false;
```
lostintangent: How about calling this `hasScheduledNextFrame`, or `isNextFramePending` or something similar? "Busy" seems a little odd in this context, so we're really talking about an event that we don't need to schedule multiple times, as opposed to an operation that can take too long.

ios/CodePush/CodePush.m
```diff
@@ -488,44 +491,43 @@ - (void)applicationWillResignActive
...
-                    sendDeviceEventWithName:@"CodePushDownloadProgress"
-                    body:@{
-                            @"totalBytes":[NSNumber numberWithLongLong:expectedContentLength],
-                            @"receivedBytes":[NSNumber numberWithLongLong:receivedContentLength]
-                          }];
-            });
+            [self updateDownloadProgressForNextFrame:expectedContentLength
+                               receivedContentLength:receivedContentLength];
+            // If the download is completed, stop observing frame updates and synchronously send the last event.
+            if (expectedContentLength == receivedContentLength) {
```
lostintangent: Would it be cleaner is `updateDownloadProgressForNextFrame` did this check? That way, this block simply requests that the progress be updated and doesn't know or care about how it's been scheduled/throttled.

geof90: But we'll still need to pause the frame observer.

geof90: Unless you're saying `updateDownloadProgressForNextFrame` should handle pausing it too if its completed?

ios/CodePush/CodePush.m
```diff
@@ -171,6 +174,8 @@ + (void)setUsingTestConfiguration:(BOOL)shouldUseTestConfiguration
 
 @synthesize bridge = _bridge;
 @synthesize methodQueue = _methodQueue;
+@synthesize pauseCallback = _pauseCallback;
```
lostintangent: Do we actually need this?

ios/CodePush/CodePush.m
```diff
@@ -7,14 +7,17 @@
...
-@interface CodePush () <RCTBridgeModule>
+@interface CodePush () <RCTBridgeModule, RCTFrameUpdateObserver>
 @end
 
 @implementation CodePush {
     BOOL _hasResumeListener;
     BOOL _isFirstRunAfterUpdate;
     int _minimumBackgroundDuration;
     NSDate *_lastResignedDate;
+    long long latestExpectedContentLength;
```
lostintangent: Should we prefix these names with underscores? That has been the convention we use for ivars.

lostintangent:
Looks beautiful!

Issue #320, opened 2016-04-28
App Store rejection due to downloaded JavaScript bundle


justim:
When we try to release our React Native iOS app in the App Store it's being rejected in the review phase, with the following message:

> **2.23** - Apps must follow the iOS Data Storage Guidelines or they will be rejected
>
> On launch and content download, your app stores 8.14MB on the user's iCloud, which does not comply with the iOS Data Storage Guidelines.

_See complete feedback reply in this [gist](https://gist.github.com/justim/520d104a4d0f9acfe166671f7230a83d)._

The 8.14MB mentioned above is the size of the JavaScript bundle.

I can't seem to find any solution for this problem or even other people mentioning it in the issues/pull requests of the relative projects (the code push ones and the react native one).

Am I missing something here?

Versions used:
- `react-native`: `0.24.0`
- `react-native-code-push`: `1.10.0-beta` (the release notes don't mention a change for this)
- The functionality itself works perfectly in TestFlight

lostintangent:
@justim Interesting, we haven't seen an app be rejected for this purpose before. Is your app silently downloading the update and restarting on app start? If their issue is strictly that the content is synced with iCloud, we can definitely update the plugin to prevent this from happening.

justim:
It's the very basic implementation of Code Push with just an `codePush.sync()` call in the `componentDidMount`, nothing fancy.

It seems from their feedback that it's just about the location of the bundle, which is synced to iCloud.

lostintangent:
@justim Yep that makes sense. It's odd that we haven't seen this rejection before, and I really apologize for the issue. We'll make a change to the plugin to prevent iCloud syncing right now and will get that out ASAP. Stay tuned!

justim:
Great! Thanks.

lostintangent:
@justim This has been resolved via #321. If you want, you can pull the update from GitHub directly. Otherwise, we'll be releasing to NPM shortly. Let us know if you run into any other issues, and once again, apologies for this inconvenience!

lostintangent:
@justim We just release `v1.10.6-beta` to NPM which includes this fix, along with some perf improvements to the update download process. Let us know how that works out for you. Thanks!

justim:
Thanks! I just pushed a new version to iTunes Connect for review with the new version.

justim:
Yay! Our app was successfully released. Thanks for the quick response! 👍

lostintangent:
[...]

PR #321, opened 2016-04-28
Exclude CodePush assets from backups

lostintangent:
Addresses #320 by explicitly marking the CodePush directory as being excluded from any backups when it is first created. By excluding the entire directory, that will transitively exclude all child files and directories. Since we write all updates and metadata files to that directory, this change will result in none of the CodePush assets being synced to end users iCloud storage accounts.

### DIFF
--- a/ios/CodePush/CodePushPackage.m
+++ b/ios/CodePush/CodePushPackage.m
@@ -60,6 +60,11 @@ static NSString *const UnzippedFolderName = @"unzipped";
                                   withIntermediateDirectories:YES
                                                    attributes:nil
                                                         error:&error];
+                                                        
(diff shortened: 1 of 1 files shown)


scottbommarito:
did u verify that before it was syncing CodePush assets to the iCloud and after this it doesn't?

if yes, LGTM

lostintangent:
@scottbommarito I wouldn't have sent out this PR if I hadn't :) Yes, in the app I was testing, I could clearly see ~1mb per CodePush update (we maintain at most 2 updates at any given time) being stored in iCloud before this change, and then after this change, there was no impact on iCloud storage after downloading CodePush updates.

geof90:
LGTM!

PR #326, opened 2016-05-02
Support iOS app extensions

kemcake:
This PR add few methods to get the `bundleURL` from an iOS app extension. Allowing extension targets to use the same JS bundle as the main app.

🙌

### DIFF
--- a/ios/CodePush/CodePush.m
+++ b/ios/CodePush/CodePush.m
@@ -15,7 +15,7 @@
     BOOL _isFirstRunAfterUpdate;
     int _minimumBackgroundDuration;
     NSDate *_lastResignedDate;
-    
+
     // Used to coordinate the dispatching of download progress events to JS.
     long long _latestExpectedContentLength;
     long long _latestReceivedConentLength;
@@ -51,6 +51,7 @@ static NSString *const PackageIsPendingKey = @"isPending";
 static BOOL isRunningBinaryVersion = NO;
 static BOOL needToReportRollback = NO;
 static BOOL testConfigurationFlag = NO;
+static BOOL isAppExtension = NO;
 
 // These values are used to save the bundleURL and extension for the JS bundle
 // in the binary.
@@ -61,7 +62,11 @@ static NSString *bundleResourceName = @"main";
 
 + (NSURL *)binaryBundleURL
 {
-    return [[NSBundle mainBundle] URLForResource:bundleResourceName withExtension:bundleResourceExtension];
+    NSBundle *bundle = [NSBundle mainBundle];
+    if (isAppExtension && [[bundle.bundleURL pathExtension] isEqualToString:@"appex"]) {
+      bundle = [NSBundle bundleWithURL:[[bundle.bundleURL URLByDeletingLastPathComponent] URLByDeletingLastPathComponent]];
+    }
+    return [bundle URLForResource:bundleResourceName withExtension:bundleResourceExtension];
 }
 
 + (NSURL *)bundleURL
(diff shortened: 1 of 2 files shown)


ios/CodePush/CodePush.m
```diff
@@ -61,7 +62,11 @@ @implementation CodePush {
 
 + (NSURL *)binaryBundleURL
 {
-    return [[NSBundle mainBundle] URLForResource:bundleResourceName withExtension:bundleResourceExtension];
+    NSBundle *bundle = [NSBundle mainBundle];
+    if (isAppExtension && [[bundle.bundleURL pathExtension] isEqualToString:@"appex"]) {
```
lostintangent: Do we actually need to maintain the `isAppExtension` variable? It seems like a generally accepted solution to look for the `apex extension in order to detect whether the current executable is an app or an app extension. I'd love to not need to create the three additional `bundleURLForAppExtension` methods unless we absolutely need them.

ios/CodePush/CodePush.m
```diff
@@ -61,7 +62,11 @@ @implementation CodePush {
...
+      bundle = [NSBundle bundleWithURL:[[bundle.bundleURL URLByDeletingLastPathComponent] URLByDeletingLastPathComponent]];
```
lostintangent: Is this line trying to reference the containing app's bundle from within the running extension? If so, could you add a comment explaining that? Presumably this logic could break in future iOS versions, so I'd love to be explicit about what we're trying to achieve if there isn't a better solution.

lostintangent:
@kemcake Thanks for submitting this PR! Could you elaborate a little bit on the scenario this is trying to enable? Are you using React Native for an app extension's UI and you want to use CodePush to load the JS Bundle for it in addition to the main app? 

Does your JS bundle contain two registered app components, and you reference each one individually from either the `AppDelegate.m` or the extension's native code?

While this PR potentially enables the extension to load the containing app's JS bundle from the binary, are there any issues with the extension loading a CodePush update? 

Thanks!

ios/CodePush/CodePush.m
kemcake: I totally agree, but there is the case where we want to have a separate bundle for your app extension...
Maybe we could juste have a `bundle` variable, default to `[NSBundle mainBundle]` and edit this variable in extension's source. Or juste expose isAppExtension...

lostintangent:
@kemcake Any chance you could provide a little more context here? I'd love to get your PR moving forward!

kemcake:
For an iOS app extension using React Native you have 2 options:

- You build an new bundle especially for this extension that you embed in the .appex app
- Or you use the same bundle but specify a different application or parameters while loading the app.

For the first case, you can just act as a normal application, and CodePush will work normally.
For the second case, the idea was to let CodePush decide where to get the bundle. After trying few approach and implement it the app where the problem came from, I finally found that since we can assume CodePush update the bundle at the application root (correct me if I'm wrong), all we need to do is to get that bundle directly **without** CodePush. We definitely don't want to start any CodePush update or job on the bundle while using the extension because it's often a short time experience and the main app could be running on the background.

To conclude, this PR is actually not necessary, I just had an idea that wasn't so good, if someone wants to use the main bundle managed by CodePush in an extension all he has to do is to get the bundle like this:

```
NSBundle* bundle = [NSBundle mainBundle];
bundle = [NSBundle bundleWithURL:[[bundle.bundleURL URLByDeletingLastPathComponent] URLByDeletingLastPathComponent]];
NSURL* bundleURL = [bundle URLForResource:@"main" withExtension:@"jsbundle"];
```
(And yes this may break in future versions so it's better not using it in CodePush directly)

PR #327, opened 2016-05-03
Windows Support

geof90:
This PR attempts to merge in the implementation of CodePush for React Native apps targeting Windows(UWP).

Still TODO:
- Calls to methods for reporting update status (reportStatusDeploy, reportStatusDownload) will currently no-op, users will not be able to see metrics for Windows apps.
- Hashing of the contents stored in the binary is not supported, so updates contents that are exactly identical to the binary contents will still be downloaded
- Recent work to throttle download progress events are not yet ported over to the Windows implementation.


windows/CodePush.cs
```diff
@@ -0,0 +1,668 @@
+﻿using System;
```
lostintangent: Can you sort these?

windows/CodePush.cs
```diff
@@ -0,0 +1,668 @@
...
+    {
+        private static bool needToReportRollback = false;
+        private static bool isRunningBinaryVersion = false;
+        private static bool testConfigurationFlag = false;
+
+        private bool didUpdate = false;
+
+        private string assetsBundleFileName;
+
+        private static readonly string ASSETS_BUNDLE_PREFIX = "ms-appx:///ReactAssets/";
```
lostintangent: Should all of these be marked as const instead of read only? We don't need to change their values in a constructor, so I think that would be the more approrpriate annotation.

windows/CodePush.cs
```diff
@@ -0,0 +1,668 @@
...
+            }
+
+            try
+            {
+                return JArray.Parse(failedUpdatesString);
+            }
+            catch (Exception)
+            {
+                JArray emptyArray = new JArray();
+                settings.Values[FAILED_UPDATES_KEY] = JsonConvert.SerializeObject(emptyArray);
```
lostintangent: What is the intent of this line?

windows/CodePush.cs
```diff
@@ -0,0 +1,668 @@
...
+
+        private class CodePushNativeModule : ReactContextNativeModuleBase
+        {
+            private ILifecycleEventListener lifecycleEventListener = null;
+            private int minimumBackgroundDuration = 0;
+            private ReactContext reactContext;
+
+            private class CodePushResumeListener : ILifecycleEventListener
+            {
+                private DateTime? lastPausedDate = null;
```
lostintangent: We should called this "lastSuspendDate" to follow the terminology of the platform.

windows/CodePush.cs
lostintangent: In general, I think the convention in C# is for constants to be named using pascal casing, so I think we should use the same coding style throughout all of this code.

windows/CodePush.cs
```diff
@@ -0,0 +1,668 @@
...
+
+namespace CodePush.ReactNative
+{
+    public class CodePushModule : IReactPackage
+    {
+        private static bool needToReportRollback = false;
+        private static bool isRunningBinaryVersion = false;
+        private static bool testConfigurationFlag = false;
+
+        private bool didUpdate = false;
```
lostintangent: Like we've been doing in Objective-C, I believe the naming convention for instance variables is to camel casing (like you're doing), but to prefix the names with an underscore.

windows/CodePushInstallMode.cs
```diff
@@ -0,0 +1,6 @@
+﻿enum CodePushInstallMode
```
lostintangent: We should wrap this in our namespace, and I think we should strip the "CodePush" prefix, since it isn't necessary with the namespace.

windows/CodePushUpdateState.cs
```diff
@@ -0,0 +1,6 @@
+﻿enum CodePushUpdateState
```
lostintangent: Same comment as the install mode enum.

windows/FileUtils.cs
```diff
@@ -0,0 +1,23 @@
+﻿using System;
+using System.Threading.Tasks;
+using Windows.Storage;
+
+namespace CodePush.ReactNative
+{
+    class FileUtils
+    {
+        public async static Task MergeDirectories(StorageFolder source, StorageFolder target)
```
lostintangent: I think we should replace "Directories" with "Folders" since that is the common term in UWP.

windows/FileUtils.cs
```diff
@@ -0,0 +1,23 @@
...
+        {
+            foreach (StorageFile sourceFile in await source.GetFilesAsync())
+            {
+                await sourceFile.CopyAndReplaceAsync(await target.CreateFileAsync(sourceFile.Name, CreationCollisionOption.OpenIfExists));
```
lostintangent: Should we be using [CopyAsync](https://msdn.microsoft.com/en-us/library/windows/apps/br227193.aspx)?

windows/CodePushUtils.cs
```diff
@@ -0,0 +1,48 @@
...
+using System.Threading.Tasks;
+using Windows.Storage;
+using Windows.Storage.Streams;
+using Windows.System.Profile;
+
+namespace CodePush.ReactNative
+{
+    class CodePushUtils
+    {
+        public static readonly string REACT_NATIVE_LOG_CATEGORY = "ReactNative";
```
lostintangent: Should this be private?

windows/CodePushUtils.cs
```diff
@@ -0,0 +1,48 @@
...
+            string jsonString = await FileIO.ReadTextAsync(file);
+            if (jsonString.Length == 0)
+            {
+                return new JObject();
+            }
+
+            return JObject.Parse(jsonString);
+        }
+
+        public static void log(string message)
```
lostintangent: Pascal casing?

windows/CodePushUtils.cs
```diff
@@ -0,0 +1,48 @@
...
+            Debug.WriteLine("[CodePush] " + message, REACT_NATIVE_LOG_CATEGORY);
+        }
+        
+        public static void logBundleUrl(string path)
+        {
+            log("Loading JS bundle from \"" + path + "\"");
+        }
+
+        public static string GetDeviceId()
+        {
```
lostintangent: This is a general purpose comment: does it make sense to annotate a method as public when it's contained in an internal class? I feel like we should potentially remove the public modifiers from all internal classes, since I'm not sure it makes a ton of sense.

windows/CodePush.cs
```diff
@@ -0,0 +1,668 @@
...
+            currentInstance = this;
+        }
+
+        private async Task ClearReactDevBundleCache()
+        {
+            StorageFile devBundleCacheFile = null;
+            try
+            {
+                devBundleCacheFile = await ApplicationData.Current.LocalFolder.GetFileAsync(REACT_DEV_BUNDLE_CACHE_FILE_NAME);
+            }
```
lostintangent: Should we use [TryGetItemAsync](https://msdn.microsoft.com/en-us/library/windows/apps/windows.storage.storagefolder.trygetitemasync) instead? That way you don't have to rely on the exception handling logic.

windows/CodePush.cs
```diff
@@ -0,0 +1,668 @@
...
+        {
+            return GetJavaScriptBundleFile(DEFAULT_JS_BUNDLE_NAME);
+        }
+
+        public string GetJavaScriptBundleFile(string assetsBundleFileName)
+        {
+            if (currentInstance == null)
+            {
+                throw new CodePushNotInitializedException("A CodePush instance has not been created yet. Have you added it to your app's list of ReactPackages?");
+            }
```
lostintangent: Do we actually need this custom exception type?

windows/CodePush.cs
```diff
@@ -0,0 +1,668 @@
...
+                return null;
+            }
+        }
+
+        private void InitializeUpdateAfterRestart()
+        {
+            JObject pendingUpdate = GetPendingUpdate();
+            if (pendingUpdate != null)
+            {
+                didUpdate = true;
```
lostintangent: Should we move this to the else block below?

windows/CodePushUnknownException.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
```
lostintangent: I don't think we need this exception type since we don't explicitly catch it anywhere, and we don't store any custom metadata along with the error message either.

windows/CodePushPackage.cs
```diff
@@ -0,0 +1,364 @@
...
+        private readonly string DOWNLOAD_URL_KEY = "downloadUrl";
+        private readonly string PACKAGE_FILE_NAME = "app.json";
+        private readonly string PACKAGE_HASH_KEY = "packageHash";
+        private readonly string PREVIOUS_PACKAGE_KEY = "previousPackage";
+        private readonly string RELATIVE_BUNDLE_PATH_KEY = "bundlePath";
+        private readonly string STATUS_FILE = "codepush.json";
+        private readonly string UNZIPPED_FOLDER_NAME = "unzipped";
+
+        public CodePushPackage()
+        {
```
lostintangent: Do we need to define this? I thought C# auto-gens a parameter less constructor for you.

windows/CodePushNotInitializedException.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
```
lostintangent: I don't think we need this custom error type.

windows/CodePushPackage.cs
```diff
@@ -0,0 +1,364 @@
...
+                return await CodePushUtils.GetJObjectFromFile(packageFile);
+            }
+            catch (IOException)
+            {
+                return null;
+            }
+        }
+
+        public string shortenPackageHash(string packageHash)
+        {
```
lostintangent: Should this be private? Also, the "s" should be uppercase.

windows/CodePushPackage.cs
```diff
@@ -0,0 +1,364 @@
...
+using System.IO.Compression;
+using System.Threading;
+using System.Threading.Tasks;
+using Windows.Storage;
+using Windows.Storage.Streams;
+using Windows.Web.Http;
+
+namespace CodePush.ReactNative
+{
+    class CodePushPackage
```
lostintangent: What do you think about calling this "UpdateManager"? That way we keep the term "Package" out of the new UWP implementation, and I also don't think we need to "CodePush" prefix thanks to the namespace.

windows/CodePushPackage.cs
```diff
@@ -0,0 +1,364 @@
...
+            {
+                return await CodePushUtils.GetJObjectFromFile(statusFile);
+            }
+            catch (Exception e)
+            {
+                throw new CodePushUnknownException("Error getting current package info", e);
+            }
+        }
+
+        public async Task UpdateCurrentPackageInfo(JObject packageInfo)
```
lostintangent: Can we sort the methods in this file? Also, I believe that many of the methods in this class are private. Can we update them as such?

windows/CodePushInvalidUpdateException.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
+
+namespace CodePush.ReactNative
+{
+    class CodePushInvalidUpdateException : Exception
```
lostintangent: I think we should remove the "CodePush" prefix, since the namespace makes this clear already.

windows/CodePush.cs
```diff
@@ -0,0 +1,668 @@
...
+        }
+
+        public IReadOnlyList<Type> CreateJavaScriptModulesConfig()
+        {
+            return new List<Type>();
+        }
+
+        public IReadOnlyList<INativeModule> CreateNativeModules(ReactContext reactContext)
+        {
+            List<INativeModule> nativeModules = new List<INativeModule>();
```
lostintangent: We could use the [collection initializer](https://msdn.microsoft.com/en-us/library/bb384062.aspx#Anchor_2) syntax to clean this up a little bit.

windows/CodePush.cs
```diff
@@ -0,0 +1,668 @@
...
+
+            [ReactMethod]
+            public void getNewStatusReport(IPromise promise)
+            {
+                // TODO implement this
+                promise.Resolve("");
+            }
+
+            [ReactMethod]
+            public void installUpdate(JObject updatePackage, int installMode, int minimumBackgroundDuration, IPromise promise)
```
lostintangent: Does the RN4W bridge not allow accepting custom enums as parameters?

windows/CodePush.cs
```diff
@@ -0,0 +1,668 @@
...
+                    return "CodePush";
+                }
+            }
+
+            public override IReadOnlyDictionary<string, object> Constants
+            {
+                get
+                {
+                    Dictionary<string, object> constants = new Dictionary<string, object>();
+                    constants["codePushInstallModeImmediate"] = CodePushInstallMode.IMMEDIATE;
```
lostintangent: Should we use a [dictionary initializer](https://msdn.microsoft.com/en-us/library/bb531208.aspx) for this?

windows/CodePush.cs
```diff
@@ -0,0 +1,668 @@
...
+            RemoveFailedUpdates();
+        }
+
+        private class CodePushNativeModule : ReactContextNativeModuleBase
+        {
+            private ILifecycleEventListener lifecycleEventListener = null;
+            private int minimumBackgroundDuration = 0;
+            private ReactContext reactContext;
+
+            private class CodePushResumeListener : ILifecycleEventListener
```
lostintangent: I would vote to move this class into its own file if we can.

windows/CodePush.cs
```diff
@@ -0,0 +1,668 @@
...
+        }
+
+        public async Task ClearUpdates()
+        {
+            await codePushPackage.ClearUpdates();
+            RemovePendingUpdate();
+            RemoveFailedUpdates();
+        }
+
+        private class CodePushNativeModule : ReactContextNativeModuleBase
```
lostintangent: Can we move this class into its own file?

windows/CodePush.cs
lostintangent: We should be able to just throw an [InvalidOoerationExceprion](https://msdn.microsoft.com/en-us/library/system.invalidoperationexception(v=vs.110).aspx)

windows/CodePushInvalidUpdateException.cs
lostintangent: Actually I think we can replace this custom exception type and just throw [InvalidOoerationException](https://msdn.microsoft.com/en-us/library/system.invalidoperationexception(v=vs.110).aspx) from the download method.

windows/CodePush.cs
```diff
@@ -0,0 +1,668 @@
...
+                    await ClearUpdates();
+                }
+
+                CodePushUtils.logBundleUrl(binaryJsBundleUrl);
+                isRunningBinaryVersion = true;
+                return binaryJsBundleUrl;
+            }
+        }
+
+        private ApplicationDataContainer GetCodePushSettings()
```
lostintangent: What do you think about moving this and every method that (transitively) calls it into a new "SettingsManager" class?

windows/CodePush.cs
```diff
@@ -0,0 +1,668 @@
...
+
+            if (devBundleCacheFile != null)
+            {
+                await devBundleCacheFile.DeleteAsync();
+            }
+        }
+
+        private async Task<long> GetBinaryResourcesModifiedTime()
+        {
+            StorageFile assetJSBundleFile = await StorageFile.GetFileFromApplicationUriAsync(new Uri(ASSETS_BUNDLE_PREFIX + assetsBundleFileName));
```
lostintangent: I think we should consider using "var" for type inference throughout this code, since it's obvious in many places what the type is.

windows/CodePush.cs
```diff
@@ -0,0 +1,668 @@
...
+using Newtonsoft.Json.Linq;
+using Newtonsoft.Json;
+using System.Reflection;
+using Windows.Web.Http;
+using Windows.Storage.FileProperties;
+using ReactNative;
+
+namespace CodePush.ReactNative
+{
+    public class CodePushModule : IReactPackage
```
lostintangent: We should mark this class as sealed, and I'm actually wondering whether we should call it "CodePushPackage" since that more accurately reflects its purpose to the app developer.

windows/CodePush.cs
```diff
@@ -0,0 +1,668 @@
...
+                reactInstanceManagerType
+                    .GetField("_jsBundleFile", BindingFlags.NonPublic | BindingFlags.Instance)
+                    .SetValue(reactInstanceManager, latestJSBundleFile);
+
+                // #3) Get the context creation method and fire it on the UI thread (which RN enforces)
+                Action recreateReactContextAction = () =>
+                {
+                    reactInstanceManager.RecreateReactContextInBackground();
+                };
+                Context.RunOnDispatcherQueueThread(recreateReactContextAction);
```
lostintangent: Instead of creating the Action instance, can't we simply pass a reference to "reactInstanceManager.RecrateReactContextInBackground" to this method?

windows/CodePush.cs
```diff
@@ -0,0 +1,668 @@
...
+                    return constants;
+                }
+            }
+
+            public override void Initialize()
+            {
+                codePush.InitializeUpdateAfterRestart();
+            }
+
+            private async Task LoadBundle()
```
lostintangent: Does it make sense for this method to be async since it should be synchronously reloading the app upon being called?

README.md
```diff
@@ -7,16 +7,8 @@ This plugin provides client-side integration for the [CodePush service](http://c
 * [Supported Components](#supported-components)
 * [Getting Started](#getting-started)
 * [iOS Setup](#ios-setup)
-    * [Plugin Installation](#plugin-installation-ios)
```
scottbommarito: wait, why are we taking these out if we never touched those sections in the readme?

windows/CodePush.cs
scottbommarito: not super familiar with react native, but if we call it CodePushPackage then that gets confusing with the concept of a "package" with respect to CodePush, aka local and remote packages. maybe CodePushReactPackage to avoid confusion?

windows/CodePush.cs
scottbommarito: we should take a look at our Android code then--the naming convention for member variables in native Android code is to prefix with an m, aka mVariableName

windows/CodePush.cs
scottbommarito: +1 on the change them to const, you're not doing anything special here so just plain old compile-time const is completely fine

windows/CodePush.cs
lostintangent: That's besides the point. We definitely have room for improving our Android implementation, but I'm simply proposing that we try not to introduce non-idiomatic C#, since this is entirely new code. Refactoring existing code is a different discussion.

windows/CodePush.cs
lostintangent: I proposed this name precisely because I am familiar with React Native, and the terminology for what this class represents is a "Package". Just looking at the interface it implements makes that clear :)

Our use of the word "Package" in our API is largely a legacy issue, and something that we've started removing over time. I don't feel super strongly about this comment, but to be entirely accurate, this class doesn't represent the CodePush module, it represents the React Native package object that contributes the module to the runtime.

windows/CodePush.cs
```diff
@@ -0,0 +1,668 @@
...
+        private static bool isRunningBinaryVersion = false;
+        private static bool testConfigurationFlag = false;
+
+        private bool didUpdate = false;
+
+        private string assetsBundleFileName;
+
+        private static readonly string ASSETS_BUNDLE_PREFIX = "ms-appx:///ReactAssets/";
+        private static readonly string BINARY_MODIFIED_TIME_KEY = "binaryModifiedTime";
+        private readonly string CODE_PUSH_PREFERENCES = "CodePush";
```
scottbommarito: why are some of these static and some not?

windows/CodePush.cs
```diff
@@ -0,0 +1,668 @@
...
+{
+    public class CodePushModule : IReactPackage
+    {
+        private static bool needToReportRollback = false;
+        private static bool isRunningBinaryVersion = false;
+        private static bool testConfigurationFlag = false;
+
+        private bool didUpdate = false;
+
+        private string assetsBundleFileName;
```
scottbommarito: this and "didUpdate" should be ordered below the block of static keys right below this, statics/consts/readonlys should all be at the top of the file and then the nonconsts/nonstatics should be next

windows/CodePush.cs
scottbommarito: my understanding of it is "if we can't parse the failedUpdates from the settings, then replace the failedUpdates in the settings with an empty JSON object". makes sense to me if we're sure that the only reason JArray.parse could fail is that failedUpdatesString is filled with invalid data, and that we're not overwriting good data that just failed to parse once due to some other issue

windows/CodePush.cs
scottbommarito: I agree, would look a lot cleaner.

windows/CodePush.cs
```diff
@@ -0,0 +1,668 @@
...
+
+                CodePushUtils.logBundleUrl(binaryJsBundleUrl);
+                isRunningBinaryVersion = true;
+                return binaryJsBundleUrl;
+            }
+        }
+
+        private ApplicationDataContainer GetCodePushSettings()
+        {
+            return ApplicationData.Current.LocalSettings.CreateContainer(CODE_PUSH_PREFERENCES, ApplicationDataCreateDisposition.Always);
```
scottbommarito: is it necessary to keep calling create container every time we need to read or write to the settings? is it possible we could cache this and save time?

windows/CodePush.cs
```diff
@@ -0,0 +1,668 @@
...
+        }
+
+        private JArray GetFailedUpdates()
+        {
+            ApplicationDataContainer settings = GetCodePushSettings();
+            string failedUpdatesString = (string)settings.Values[FAILED_UPDATES_KEY];
+            if (failedUpdatesString == null)
+            {
+                return new JArray();
```
scottbommarito: it looks to me like all of these two get settings functions are very similar (getFailedUpdates/getPendingUpdate). get settings, read from settings and return default if failure, then return parsed object from settings and write default to settings if failure. could we make a helper function for them to use and also so that if we need to expand to additional settings we can do so easily?

windows/CodePush.cs
```diff
@@ -0,0 +1,668 @@
...
+
+        private bool IsPendingUpdate(string packageHash)
+        {
+            JObject pendingUpdate = GetPendingUpdate();
+            return pendingUpdate != null &&
+                    !(bool)pendingUpdate[PENDING_UPDATE_IS_LOADING_KEY] &&
+                    (packageHash == null || ((string)pendingUpdate[PENDING_UPDATE_HASH_KEY]).Equals(packageHash));
+        }
+
+        private void RemoveFailedUpdates()
```
scottbommarito: this and RemovePendingUpdate are nearly identical but with a different key...not sure how helpful it would be to make a helper function but if there were 3 functions here I would definitely do it

windows/CodePush.cs
scottbommarito: may need to do that on the cordova side as well, I think I named the variable there "lastPausedDate" to follow this lol

windows/CodePush.cs
scottbommarito: it's short enough where I don't see a problem with it in the same file as CodePushNativeModule, but I agree that CodePushNativeModule should be moved to its own file

README.md
shishirx34: Probably to clean up unnecessary links, and noise in index. I see windows setup section added below which would need adding these sublinks under that header too in the index. I am fine either ways.

windows/CodePush.cs
```diff
@@ -0,0 +1,668 @@
...
+        private async Task ClearReactDevBundleCache()
+        {
+            StorageFile devBundleCacheFile = null;
+            try
+            {
+                devBundleCacheFile = await ApplicationData.Current.LocalFolder.GetFileAsync(REACT_DEV_BUNDLE_CACHE_FILE_NAME);
+            }
+            catch (FileNotFoundException)
+            {
+            }
```
shishirx34: May be log file not found for clear react?

windows/CodePush.cs
shishirx34: could you add `this.` to the instance variables so that its obvious where the value is coming from?

`this.assetsBundleFileName`

windows/CodePush.cs
```diff
@@ -0,0 +1,668 @@
...
+        {
+            this.assetsBundleFileName = assetsBundleFileName;
+            string binaryJsBundleUrl = ASSETS_BUNDLE_PREFIX + assetsBundleFileName;
+            long binaryResourcesModifiedTime = await GetBinaryResourcesModifiedTime();
+            StorageFile packageFile = await codePushPackage.GetCurrentPackageBundle(this.assetsBundleFileName);
+            if (packageFile == null)
+            {
+                // There has not been any downloaded updates.
+                CodePushUtils.logBundleUrl(binaryJsBundleUrl);
+                isRunningBinaryVersion = true;
```
shishirx34: this.isRunningBinaryVersion

windows/CodePush.cs
```diff
@@ -0,0 +1,668 @@
...
+            {
+                binaryModifiedDateDuringPackageInstall = long.Parse(binaryModifiedDateDuringPackageInstallString);
+            }
+
+            string packageAppVersion = (string)packageMetadata["appVersion"];
+
+            // TODO: test configuration
+            if (binaryModifiedDateDuringPackageInstall != null &&
+                    binaryModifiedDateDuringPackageInstall == binaryResourcesModifiedTime &&
+                    (IsUsingTestConfiguration() || appVersion.Equals(packageAppVersion)))
```
shishirx34: this._appVersion

windows/CodePush.cs
```diff
@@ -0,0 +1,668 @@
...
+            {
+                CodePushUtils.logBundleUrl(packageFile.Path);
+                isRunningBinaryVersion = false;
```
shishirx34: `this._isRunningBinaryVersion` here and everywhere else applicable.

windows/CodePush.cs
```diff
@@ -0,0 +1,668 @@
...
+                };
+
+                Context.RunOnNativeModulesQueueThread(downloadAction);
+            }
+
+            [ReactMethod]
+            public void getConfiguration(IPromise promise)
+            {
+                JObject config = new JObject();
+                config["appVersion"] = codePush.appVersion;
```
scottbommarito: there's gotta be some initializer syntax for this...

windows/CodePushPackage.cs
scottbommarito: Yea this file isn't really a package, it's a collection of methods that deal with packages/updates. I like "UpdateManager"

windows/CodePushPackage.cs
```diff
@@ -0,0 +1,364 @@
...
+using System.Threading.Tasks;
+using Windows.Storage;
+using Windows.Storage.Streams;
+using Windows.Web.Http;
+
+namespace CodePush.ReactNative
+{
+    class CodePushPackage
+    {
+        private readonly string CODE_PUSH_FOLDER_PREFIX = "CodePush";
```
scottbommarito: const instead?

windows/CodePush.cs
```diff
@@ -0,0 +1,668 @@
...
+        private string deploymentKey;
+        private readonly string serverUrl = "https://codepush.azurewebsites.net/";
+
+        private ReactPage mainPage;
+
+        private static CodePushModule currentInstance;
+
+        public CodePushModule(string deploymentKey, ReactPage mainPage)
+        {
+            codePushPackage = new CodePushPackage();
```
scottbommarito: glancing at the CodePushPackage class, it seems to me like the class as a whole just really consists of methods that might as well be static anyway and therefore not require an instance of the class. do you agree? could we get rid of this private variable and just call `CodePushPackage.functionName()` directly?

windows/CodePushPackage.cs
```diff
@@ -0,0 +1,364 @@
...
+            }
+            else
+            {
+                return (string)previousPackageMetadata[PACKAGE_HASH_KEY];
+            }
+        }
+
+        public async Task<JObject> GetCurrentPackage()
+        {
+            string packageHash = await GetCurrentPackageHash();
```
scottbommarito: how about a GetPackageFromHash(string packageHash) function that GetCurrentPackage and GetPendingPackage wrap?

windows/CodePushPackage.cs
```diff
@@ -0,0 +1,364 @@
...
+            if (packageHash == null)
```
scottbommarito: if you move these null checks into GetPackage you can shorten these functions down to "return await GetPackage(await GetCurrentPackageHash());" and then u also don't have to worry about being passed null packageHashes

windows/CodePushPackage.cs
```diff
@@ -0,0 +1,364 @@
...
+            }
+            catch (FileNotFoundException)
+            {
+                return null;
+            }
+        }
+
+        public async Task<string> GetCurrentPackageHash()
+        {
+            JObject info = await GetCurrentPackageInfo();
```
scottbommarito: GetCurrentPackageHash and GetPreviousPackageHash are nearly identical except using a different key...helper function?

windows/CodePushPackage.cs
```diff
@@ -0,0 +1,364 @@
...
+        }
+
+        public string shortenPackageHash(string packageHash)
+        {
+            return packageHash.Substring(0, 8);
+        }
+
+        public async Task DownloadPackage(JObject updatePackage, string expectedBundleFileName, Progress<HttpProgress> downloadProgress)
+        {
+            StorageFolder codePushFolder = await GetCodePushFolder();
```
scottbommarito: I would recommend adding comments above each of those 4 line chunks of code.

e.g. `// Get the folder of the new update using its hash`

windows/CodePush.cs
geof90: It could be the same now, but it might not be the same in the future, therefore I don't want to over-engineer this.

windows/CodePush.cs
geof90: It could be the same now, but it might not be the same in the future, therefore I don't want to over-engineer this.

windows/CodePush.cs
geof90: Yup, basically to help it recover from a bad state.

windows/CodePushPackage.cs
```diff
@@ -0,0 +1,364 @@
...
+                // Downloaded file is not a zip, assume it is a jsbundle
+                await downloadFile.RenameAsync(expectedBundleFileName);
+                await downloadFile.MoveAsync(newUpdateFolder);
+            }
+
+            // Save metadata to the folder
+            await FileIO.WriteTextAsync(newUpdateMetadataFile, JsonConvert.SerializeObject(updatePackage));
+        }
+
+        public async Task InstallPackage(JObject updatePackage, bool removePendingUpdate)
```
scottbommarito: I think this method needs some comments to be a little clearer. From now, what I understand is "if removePendingUpdate, then the current package is removed from the current package info and replaced by the update package and the previous package remains--otherwise, the previous package is removed and replaced by the current package and the update package is put as the current package." The phrase "removePendingUpdate" is not descriptive enough for me because what it's really doing is deciding which package remains in the package info, and while the current package is assumed to be a pending update, that is not inforced or explained in this code anywhere.

windows/CodePush.cs
geof90: I did think about it, but each implementation of this method is 2 lines. Making a helper function would turn it to one line, but add 1 extra hop of method indirection for each, and add one more new method.. the readability and brevity gain from doing this doesn't seem that great.

windows/CodePushPackage.cs
```diff
@@ -0,0 +1,364 @@
...
+            }
+
+            info[CURRENT_PACKAGE_KEY] = info[PREVIOUS_PACKAGE_KEY];
+            info[PREVIOUS_PACKAGE_KEY] = null;
+            await UpdateCurrentPackageInfo(info);
+        }
+
+        public void DownloadAndReplaceCurrentBundle(string remoteBundleUrl, string bundleFileName)
+        {
+            // TODO implement this method (only used in tests)
```
scottbommarito: if this is only used in your ios tests that you made for the react native ios tests, then i'm probably going to remove it so u shouldn't worry about having to implement this

windows/CodePushUpdateUtils.cs
```diff
@@ -0,0 +1,49 @@
...
+using System.IO;
+using System.Threading.Tasks;
+using Windows.Storage;
+
+namespace CodePush.ReactNative
+{
+    class CodePushUpdateUtils
+    {
+        // TODO: Generate binary hash
+        // private static readonly String CODE_PUSH_HASH_FILE_NAME = "CodePushHash.json";
```
scottbommarito: (const?)

windows/CodePushUpdateUtils.cs
```diff
@@ -0,0 +1,49 @@
...
+
+        public async static Task CopyNecessaryFilesFromCurrentPackage(StorageFile diffManifestFile, StorageFolder currentPackageFolder, StorageFolder newPackageFolder)
```
scottbommarito: this does more than just copying...it also deletes the files that were deleted by the update. this is more like `performDiffWithPackage` or something

windows/CodePushUtils.cs
```diff
@@ -0,0 +1,48 @@
...
+
+        public async static Task<JObject> GetJObjectFromFile(StorageFile file)
+        {
+            string jsonString = await FileIO.ReadTextAsync(file);
+            if (jsonString.Length == 0)
+            {
+                return new JObject();
+            }
+
+            return JObject.Parse(jsonString);
```
scottbommarito: you try-catch this in getPendingUpdate...is it safe not to do so here?

windows/CodePushUtils.cs
```diff
@@ -0,0 +1,48 @@
...
+using System;
+using System.Diagnostics;
+using System.Threading.Tasks;
+using Windows.Storage;
+using Windows.Storage.Streams;
+using Windows.System.Profile;
+
+namespace CodePush.ReactNative
+{
+    class CodePushUtils
```
scottbommarito: probably should get rid of the CodePush prefix

windows/CodePushUpdateUtils.cs
```diff
@@ -0,0 +1,49 @@
+﻿using Newtonsoft.Json.Linq;
+using System;
+using System.IO;
+using System.Threading.Tasks;
+using Windows.Storage;
+
+namespace CodePush.ReactNative
+{
+    class CodePushUpdateUtils
```
scottbommarito: probably should get rid of the CodePush prefix

windows/CodePush.cs
geof90: I think @lostintangent is referring to matching the terminology of the platform, i.e.
`lastSuspendDate` -> `OnSuspend()` in Windows
`lastPausedDate` -> 'OnPause()` in Android

Therefore I think keeping what we have on the Cordova side is fine.

windows/FileUtils.cs
scottbommarito: yea, looks like you would have to use `CopyAsync(destinationFolder, newName, nameCollisionOption)` and pass in the existing name as the new name and replaceIfExisting for the nameCollisionOption

windows/CodePush.cs
scottbommarito: I think if we follow the idiom with the underscore prefixes then using "this." shouldn't be necessary

windows/CodePush.cs
scottbommarito: yea I totally agree, I was just making a mental note to fix the idiom in our java code sometime

windows/CodePush.cs
scottbommarito: yea I agree was just mentioning this as something to consider

windows/CodePush.cs
geof90: I think in the future, we want to enable the scenario that an app has multiple `ReactPage`s, and a user might want to CodePush them individually.

windows/CodePush.cs
geof90: It was a hack for the iOS version of the tests, where it was not easy to test changes the info.plist app version in the XCTest framework. I agree this should be removed in the windows implementation.

windows/CodePush.cs
geof90: Specifying ApplicationDataCreateDisposition.Always creates it if it doesn't exist, and opens it if it does.
https://msdn.microsoft.com/library/windows/apps/br241611

windows/CodePush.cs
geof90: But `codePush.GetJavaScriptBundleFileAsync(codePush.assetsBundleFileName);` is a `Task`, so this method needs to be `async` in order to await on that.

windows/CodePushPackage.cs
geof90: How about renaming `removePendingUpdate` to `overwriteCurrentUpdate`?

windows/CodePushUpdateUtils.cs
geof90: Its not performing the full diff though. The full diff needs to include the new files too. This copies the necessary files (omitting files that are not needed) from the previous package. Hence the name `CopyNecessaryFilesFromCurrentPackage`

windows/CodePushUtils.cs
geof90: Because this is an internal method, I think it should be the caller's responsibility to give a file that has valid JSON written in it, and if not, a JSON parsing exception should be thrown. `try... catch`-ing it and returning `null` will make the error harder to diagnose.

geof90:
thanks for all the reviews guys! I have incorporated most of the feedback, it should look way better now, please take another look.

README.md
geof90: Exactly right @shishirx34

windows/CodePushUnknownException.cs
geof90: Its any unexpected exception thrown can be annotated with our error message and that it is indeed coming from CodePush.

geof90: Right... I forgot that Exceptions are not checked in C#.

windows/UpdateManager.cs
```diff
@@ -0,0 +1,269 @@
...
+using System.Threading.Tasks;
+using Windows.Storage;
+using Windows.Storage.Streams;
+using Windows.Web.Http;
+
+namespace CodePush.ReactNative
+{
+    internal class UpdateManager
+    {
+        // Internal methods
```
lostintangent: If we want to achieve tooling support for grouping method definitions, you could use the [#region directive](https://msdn.microsoft.com/en-us/library/9a1ybwek.aspx), which allows expanding/collapsing the entire group. This is somewhat equivalent to our use of #pragma in Obj-C.

windows/SettingsManager.cs
```diff
@@ -0,0 +1,98 @@
...
+using Windows.Storage;
+
+namespace CodePush.ReactNative
+{
+    internal class SettingsManager
+    {
+        private static ApplicationDataContainer GetCodePushSettings()
+        {
+            return ApplicationData.Current.LocalSettings.CreateContainer(CodePushConstants.CodePushPreferences, ApplicationDataCreateDisposition.Always);
+        }
```
lostintangent: I wonder whether it would be nicer to have a static constructor that initialized a private variable, that all other methods in this class use internally. The static constructor won't get called until one of its classes methods are actually called, so you gain the same benefits of deferring this code until it's necessary, but the code that consumes it would arguably be a little cleaner.

windows/FileUtils.cs
```diff
@@ -0,0 +1,23 @@
+﻿using System;
+using System.Threading.Tasks;
+using Windows.Storage;
+
+namespace CodePush.ReactNative
+{
+    internal class FileUtils
+    {
+        internal async static Task MergeDirectories(StorageFolder source, StorageFolder target)
+        {
```
lostintangent: I would still vote for using "Folders" instead of "Directories".

windows/CodePushReactPackage.cs
```diff
@@ -0,0 +1,223 @@
...
+                        ClearReactDevBundleCache().Wait();
+                    }
+                    // Mark that we tried to initialize the new update, so that if it crashes,
+                    // we will know that we need to rollback when the app next starts.
+                    SettingsManager.SavePendingUpdate((string)pendingUpdate[CodePushConstants.PendingUpdateHashKey], /* isLoading */true);
+                }
+            }
+        }
+
+        internal bool IsFailedHash(string packageHash)
```
lostintangent: Should this and isPendingUpdate be moved to the SettingsManager?

windows/CodePushResumeListener.cs
```diff
@@ -0,0 +1,49 @@
+﻿using ReactNative.Bridge;
+using System;
+
+namespace CodePush.ReactNative
+{
+    internal class CodePushResumeListener : ILifecycleEventListener
```
lostintangent: I would vote to rename this to "LifecycleEventListener". We listen to suspend as well, so it isn't just a resume listener, and the CodePush prefix doesn't add any value since the namespace already scopes it down pretty specifically.

windows/CodePushResumeListener.cs
```diff
@@ -0,0 +1,49 @@
...
+    {
+        private DateTime? _lastSuspendDate = null;
+        private CodePushNativeModule _codePushNativeModule;
+
+        internal int MinimumBackgroundDuration { get; set; }
+
+        internal CodePushResumeListener(CodePushNativeModule codePushNativeModule, int minimumBackgroundDuration)
```
lostintangent: What do you think about this constructor simply taking in an Action that it triggers as necessary? The consuming code could then just pass in a reference to the loadBundle method, and this class is completely simpllified to just listening to lifecycle events and throttling them as necessary.

windows/InstallMode.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace CodePush.ReactNative
+{
+    enum InstallMode
+    {
+        IMMEDIATE = 0,
```
lostintangent: These names should be pascal-cased, and you can omit the values, since the compiler will take care of it.

windows/UpdateState.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace CodePush.ReactNative
+{
+    enum UpdateState
```
lostintangent: Same comment as the install mode enum.

windows/CodePushNativeModule.cs
```diff
@@ -0,0 +1,279 @@
...
+
+            Context.RunOnNativeModulesQueueThread(downloadAction);
+        }
+
+        [ReactMethod]
+        public void getConfiguration(IPromise promise)
+        {
+            var config = new JObject
+            {
+                { "appVersion", _codePush.AppVersion },
```
lostintangent: Sort by key?

windows/CodePushNativeModule.cs
```diff
@@ -0,0 +1,279 @@
...
+                        _codePush.AssetsBundleFileName,
+                        new Progress<HttpProgress>(
+                            (HttpProgress progress) =>
+                            {
+                                if (!notifyProgress)
+                                {
+                                    return;
+                                }
+
+                                var downloadProgress = new JObject();
```
lostintangent: Use the dictionary initializer syntax?

windows/CodePushInvalidUpdateException.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
```
lostintangent: Since we're not actually augmenting the error object in any way, I still vote for us using the InvalidOperationException class (or another one as appropriate). I believe that C# favors re-using core exception types as much as possible, and I think we can achieve the behavior we want of rejecting a bad download without needing to subclass our our exception type.

windows/CodePushUtils.cs
```diff
@@ -0,0 +1,53 @@
...
+        }
+
+        internal static void LogBundleUrl(string path)
+        {
+            Log("Loading JS bundle from \"" + path + "\"");
+        }
+
+        internal static string GetDeviceId()
+        {
+            HardwareToken token = HardwareIdentification.GetPackageSpecificToken(null);
```
lostintangent: The var keyword definitely seems appropriate in a few places in this method.

windows/SettingsManager.cs
```diff
@@ -0,0 +1,98 @@
...
+
+            failedUpdates.Add(failedPackage);
+            settings.Values[CodePushConstants.FailedUpdatesKey] = JsonConvert.SerializeObject(failedUpdates);
+        }
+
+        internal static void SavePendingUpdate(string packageHash, bool isLoading)
+        {
+            ApplicationDataContainer settings = GetCodePushSettings();
+            var pendingUpdate = new JObject();
+            pendingUpdate[CodePushConstants.PendingUpdateHashKey] = packageHash;
```
lostintangent: Use dictionary initializer syntax?

windows/CodePushPackage.cs
scottbommarito: i like `overwriteCurrentUpdate` a lot more

windows/CodePushLifecycleEventListener.cs
```diff
@@ -0,0 +1,44 @@
+﻿using ReactNative.Bridge;
+using System;
+
+namespace CodePush.ReactNative
+{
+    internal class CodePushLifecycleEventListener : ILifecycleEventListener
+    {
+        private DateTime? _lastSuspendDate = null;
+        private Action _loadBundleAction;
+
```
lostintangent: I would keep the name of this agnostic to what it actually does, that way it's fully decoupled from the consumer. Maybe just "_resumeAction"?

windows/CodePushNativeModule.cs
```diff
@@ -0,0 +1,291 @@
...
+                var pendingHash = (string)updatePackage[CodePushConstants.PackageHashKey];
+                SettingsManager.SavePendingUpdate(pendingHash, /* isLoading */false);
+                if (installMode == (int)InstallMode.OnNextResume)
+                {
+                    if (_codePushLifecycleEventListener == null)
+                    {
+                        // Ensure we do not add the listener twice.
+                        Action loadBundleAction = () =>
+                        {
+                            Context.RunOnNativeModulesQueueThread(async () =>
```
lostintangent: Does this need to be dispatched to the native modules thread? Since the reload logic just immediately dispatches back to the UI thread, and this operation doesn't need to actually be async, I don't think we need to worry about doing this unless it's required.

windows/CodePushReactPackage.cs
```diff
@@ -0,0 +1,200 @@
...
+        #region Public methods
+        public IReadOnlyList<Type> CreateJavaScriptModulesConfig()
+        {
+            return new List<Type>();
+        }
+
+        public IReadOnlyList<INativeModule> CreateNativeModules(ReactContext reactContext)
+        {
+            _codePushNativeModule = new CodePushNativeModule(reactContext, this);
+            var nativeModules = new List<INativeModule>
```
lostintangent: Just return the new List, since the local var isn't necessary?

windows/CodePushLifecycleEventListener.cs
lostintangent: If you did the above, then you could call this class something like "ThrottledResumeListener" or something more specific, without needing to be tied to CodePush per se. Not at all necessary, but it could help improve the semantics of the component.

windows/CodePushNativeModule.cs
```diff
@@ -0,0 +1,291 @@
...
+using System.IO;
+using System.Reflection;
+using System.Threading.Tasks;
+using Windows.Web.Http;
+
+namespace CodePush.ReactNative
+{
+    internal class CodePushNativeModule : ReactContextNativeModuleBase
+    {
+        private CodePushLifecycleEventListener _codePushLifecycleEventListener = null;
```
lostintangent: Sort these by variable name. Additionally, wouldn't the compiler set this to null by default?

lostintangent:
LGTM!

windows/CodePushNativeModule.cs
geof90: This only works if this action could somehow be made non-async. Unfortunately `LoadBundle()` is async, apart from dispatching back to the UI thread, it also awaits on `GetJavaScriptBundleFileAsync()`. I will try different ways to make it non-async because I agree it will be much cleaner.. but otherwise, do you think this is fine?

lostintangent: Yep I think it's fine. This was just a minor question that I was curious about.

windows/CodePushLifecycleEventListener.cs
geof90: I thought you didn't want the word "resume" in it. How about "MinimumBackgroundListener"? Right now all this listener does is perform an action once the app has been in the background for x number of seconds.

CodePush.js
```diff
@@ -271,13 +271,9 @@ async function syncInternal(options = {}, syncStatusChangeCallback, downloadProg
...
             break;
           case CodePush.SyncStatus.UPDATE_INSTALLED:
-            /*
-             * If the install mode is IMMEDIATE, this will not get returned as the
-             * app will be restarted to a new Javascript context.
-             */
             if (resolvedInstallMode == CodePush.InstallMode.ON_NEXT_RESTART) {
               log("Update is installed and will be run on the next app restart.");
-            } else {
+            } else if (resolvedInstallMode == CodePush.InstallMode.ON_NEXT_RESUME) {
```
geof90: During my testing, I found that for an IMMEDIATE install, even though the JS context gets reloaded immediately, the "resume logging" still shows up in the IDE console, that's why I made this change.

windows/CodePushReactPackage.cs
```diff
@@ -0,0 +1,197 @@
...
+            InitializeUpdateAfterRestart();
+            if (CurrentInstance != null)
+            {
+                CodePushUtils.Log("More than one CodePush instance has been initialized. Please use the instance method codePush.getBundleUrlInternal() to get the correct bundleURL for a particular instance.");
+            }
+
+            CurrentInstance = this;
+        }
+
+        #region Public methods
```
lostintangent: I would vote for adding newlines between the region directives.

windows/MinBackgroundListener.cs
```diff
@@ -0,0 +1,44 @@
...
+namespace CodePush.ReactNative
+{
+    internal class MinBackgroundListener : ILifecycleEventListener
+    {
+        private DateTime? _lastSuspendDate;
+        private Action _resumeAction;
+
+        internal int MinimumBackgroundDuration { get; set; }
+
+        internal MinBackgroundListener(Action resumeAction, int minBackgroundDuration)
```
lostintangent: I think that abbreviations are generally frowned upon in .NET for class names and methods, so the use of "Min" here feels a little odd.

PR #330, opened 2016-05-04
Moving d.ts to typings folder

lostintangent:
This PR simply moves our `*.d.ts` file to a new `typings` folder, which is the convention that libraries are using when distributing their typings file via NPM.


shishirx34:
LGTM

PR #332, opened 2016-05-07
Fix for iOS7

geof90:
This fixes https://github.com/Microsoft/code-push/issues/216.

We are using two Obj-C APIs that are only supported in iOS 8+: `[string containsString]` and `[operationQueue setUnderlyingQueue]`. This fix simply changes our implementation to use the iOS7 workarounds if necessary.


lostintangent:
LGTM!

PR #334, opened 2016-05-10
Retry Report Status if failed

geof90:
This PR improves the reportStatus mechanism in order to make sure that metrics reporting calls are not lost due to bad network conditions. If the call to the server fails/times out, we will save a copy of the statusReport in the Preferences, and retry again the next time the app resumes.

### DIFF
--- a/android/app/src/main/java/com/microsoft/codepush/react/CodePushTelemetryManager.java
+++ b/android/app/src/main/java/com/microsoft/codepush/react/CodePushTelemetryManager.java
@@ -33,14 +33,14 @@ public class CodePushTelemetryManager {
 
     public WritableMap getBinaryUpdateReport(String appVersion) {
         String previousStatusReportIdentifier = this.getPreviousStatusReportIdentifier();
+        WritableNativeMap reportMap = null;
         if (previousStatusReportIdentifier == null) {
             this.clearRetryStatusReport();
-            WritableNativeMap reportMap = new WritableNativeMap();
+            reportMap = new WritableNativeMap();
             reportMap.putString(APP_VERSION_KEY, appVersion);
-            return reportMap;
         } else if (!previousStatusReportIdentifier.equals(appVersion)) {
             this.clearRetryStatusReport();
-            WritableNativeMap reportMap = new WritableNativeMap();
+            reportMap = new WritableNativeMap();
             if (this.isStatusReportIdentifierCodePushLabel(previousStatusReportIdentifier)) {
                 String previousDeploymentKey = this.getDeploymentKeyFromStatusReportIdentifier(previousStatusReportIdentifier);
                 String previousLabel = this.getVersionLabelFromStatusReportIdentifier(previousStatusReportIdentifier);
@@ -52,10 +52,9 @@ public class CodePushTelemetryManager {
                 reportMap.putString(APP_VERSION_KEY, appVersion);
                 reportMap.putString(PREVIOUS_LABEL_OR_APP_VERSION_KEY, previousStatusReportIdentifier);
             }
-            return reportMap;
         }
 
-        return null;
+        return reportMap;
     }
 
     public WritableMap getRetryStatusReport() {
@@ -68,7 +67,6 @@ public class CodePushTelemetryManager {
                 return CodePushUtils.convertJsonObjectToWritable(retryStatusReport);
             } catch (JSONException e) {
                 e.printStackTrace();
-                return null;
             }
         }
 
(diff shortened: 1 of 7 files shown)


android/app/src/main/java/com/microsoft/codepush/react/CodePushTelemetryManager.java
```diff
@@ -24,34 +34,51 @@ public CodePushTelemetryManager(Context applicationContext, String codePushPrefe
...
                 // Previous status report was with a binary app version.
-                reportMap.putString("appVersion", appVersion);
-                reportMap.putString("previousLabelOrAppVersion", previousStatusReportIdentifier);
+                reportMap.putString(APP_VERSION_KEY, appVersion);
+                reportMap.putString(PREVIOUS_LABEL_OR_APP_VERSION_KEY, previousStatusReportIdentifier);
             }
             return reportMap;
         }
 
         return null;
```
shishirx34: initialize the reportMap outside of the above `if..else if..` block and then return it here. Remove intermediate returns in the `if...else if...` blocks

```
WriteableNativeMap reportMap = null;

if (...) {...}
else if (...) {...}

return reportMap;
```

android/app/src/main/java/com/microsoft/codepush/react/CodePushTelemetryManager.java
```diff
@@ -3,18 +3,28 @@
...
 import com.facebook.react.bridge.WritableMap;
 import com.facebook.react.bridge.WritableNativeMap;
 
+import org.json.JSONException;
+import org.json.JSONObject;
+
 public class CodePushTelemetryManager {
 
     private Context applicationContext;
+    private final String APP_VERSION_KEY = "appVersion";
```
scottbommarito: usually you make these things static as well, since they're shared across every CodePushTelemetryManager.

that may require you to add "CodePushTelemetryManager." before every use but I don't think so

android/app/src/main/java/com/microsoft/codepush/react/CodePushTelemetryManager.java
```diff
@@ -3,18 +3,28 @@
...
     private final String CODE_PUSH_PREFERENCES;
     private final String DEPLOYMENT_FAILED_STATUS = "DeploymentFailed";
```
scottbommarito: how about `DEPLOYMENT_STATUS_SUCCEEDED` and `DEPLOYMENT_STATUS_FAILED` ?

ios/CodePush/CodePush.m
```diff
@@ -783,19 +783,35 @@ - (void)applicationWillResignActive
...
 
     resolve(nil);
 }
 
+RCT_EXPORT_METHOD(recordStatusReported:(NSDictionary *)statusReport)
+{
+    [CodePushTelemetryManager recordStatusReported:statusReport];
+}
+
+RCT_EXPORT_METHOD(saveStatusReportForRetry:(NSDictionary *)statusReport)
```
scottbommarito: idk iOS idioms but this method name looks really odd to me...

scottbommarito:
LGTM...we should do this on cordova as well, a lot of the code should be almost directly portable

android/app/src/main/java/com/microsoft/codepush/react/CodePushTelemetryManager.java
geof90: Can I make style changes like this in a separate PR? Unlike my Windows PR, this is not entirely new code so I want to follow the existing convention.

android/app/src/main/java/com/microsoft/codepush/react/CodePushTelemetryManager.java
geof90: Same for this.. Its being used in existing working production code, I don't want to bloat this PR and introduce new risks.

ios/CodePush/CodePush.m
geof90: Do you have a better suggestion? I considered "queue" instead of "save" but it's not really putting anything in a queue..

lostintangent:
[Code reviewing from the plane from Dallas to Tampa]

This change looks really good! I just had a couple of thoughts:

1. Do we need to implement retry logic for the download status report? Presumably, that could fail for some reason, and if we didn't retry it later, then we wouldn't be accurately reflecting the current number of pending installs. Maybe this isn't as big of a deal, but for apps using a restart-based install, it's possible that there could be users with pending installs for hours/days, that we wouldn't be displaying in the CLI until the app is restarted and the successful install report is made. Attempting to retry it on app resume might help the data remain mostly fresh. I'm also not sure what kind of negative impact there would be if the successful install count was greater than the download count for a specific update.

2. Is it sufficient to only save a single status report for retry? Is it possible to get into a situation where multiple status reports fail and we only retry the last one, and therefore, get into a situation where the data is "corrupted"? Since we're trying to improve the reliability of this feature, I just wanted to toss out this scenario and see if it's something we need to consider.

geof90:
I did think about those cases and chose not to handle them..

1. If the install count is greater than the download count, the CLI is clever enough to show "0 pending". I also think it is very, very unlikely that the download status report would fail (since the client just completed a download successfully which means the client access to an internet connection).

2. I think saving a single report is good enough, and here's why: As long as the last report gets sent, the numbers will tally. Let's say the user downloads and installs v2, but the v2 report didn't get sent. With this change, the client will not assume that v2 got sent, so if the client somehow installs v3 without sending the v2 report, the client will report an update from v1 to v3, ignoring v2. The net effect on metrics is that there are 0 users on v1, 0 on v2, and 1 on v3, whether or not we chose to report that v1 -> v2 update.

geof90:
I guess one effect though is that if the v2 report gets lost, v2 will forever have that unresolved "pending" install.. But again, I think its very unlikely because for that to happen, the report would have to fail repeatedly everytime the app resumes. and even fail right before the call to `codePush.sync()` which downloads the v3 update, which is again unlikely because if the network conditions are so poor for that report to be dropped, its unlikely that v3 would even be downloaded successfully

lostintangent:
Sounds good, thanks a lot for confirming this.

ios/CodePush/CodePushTelemetryManager.m
```diff
@@ -50,38 +70,63 @@ + (NSDictionary *)getUpdateReport:(NSDictionary *)currentPackage
...
-                          @"status": DeploymentSucceeded,
-                          @"previousDeploymentKey": previousDeploymentKey,
-                          @"previousLabelOrAppVersion": previousLabel
+                          PackageKey: currentPackage,
+                          StatusKey: DeploymentSucceeded,
+                          PreviousDeploymentKeyKey: previousDeploymentKey,
+                          PreviousLabelOrAppVersionKey: previousLabel
                         };
             } else {
+                [self clearRetryStatusReport];
```
shishirx34: you can move this outside of this `if..else..`

ios/CodePush/CodePushTelemetryManager.m
```diff
@@ -1,46 +1,66 @@
...
+        [self clearRetryStatusReport];
+        return retryStatusReport;
+    } else {
+        return nil;
+    }
+}
+
 + (NSDictionary *)getRollbackReport:(NSDictionary *)lastFailedPackage
 {
+    [self clearRetryStatusReport];
```
shishirx34: why do you want to `clearRetryStatusReport` here inside `getRollbackReport`?

android/app/src/main/java/com/microsoft/codepush/react/CodePushTelemetryManager.java
```diff
@@ -23,71 +33,103 @@ public CodePushTelemetryManager(Context applicationContext, String codePushPrefe
...
+        }
+
+        return reportMap;
+    }
+
+    public WritableMap getRetryStatusReport() {
+        SharedPreferences settings = applicationContext.getSharedPreferences(CODE_PUSH_PREFERENCES, 0);
+        String retryStatusReportString = settings.getString(RETRY_DEPLOYMENT_REPORT_KEY, null);
+        if (retryStatusReportString != null) {
+            clearRetryStatusReport();
```
shishirx34: I am confused.. why do you `clearRetryStatusReport` when you `getRetryStatusReport`, should the status report be cleared either after successful retry or after before the retry and then overwrite the new status for retry after done?

geof90: We clear it so that the same one won't possibly get reported twice. If the retry is unsuccessful, it will get re-saved on failure.

shishirx34: Just wondering how would it get reported twice? Either there could be a failure after we getReport but before we actually report it, in that case we have lost the report, isn't it?

shishirx34:
LGTM!

android/app/src/main/java/com/microsoft/codepush/react/CodePushTelemetryManager.java
geof90: If `codePush.sync()` and `tryReportStatus()` both get called during an app resume, it could happen.. Maybe it's a rare scenario but since it's just one line needed to protect against such scenarios, i thought it wouldn't be much harm doing it.

Issue #335, opened 2016-05-11
More than one CodePush instance Error 


ACCTFORGH:
I was testing CodePush under bad network condition in my app. When the download got stuck, I minimized the app and when I click on the App icon again to open it, I saw the error message below.

**Error Message:**
 D/ReactNative( 1428): [CodePush] More than one CodePush instance has been initialized. Please use the instance method codePush.getBundleUrlInternal() to get the correct bundleURL for a particular instance.
D/ReactNative( 1428): [CodePush] Loading JS bundle from "/data/data/com.golorry/files/CodePush/51febf286bed5defb9314cc4e2c95b91a1a33aa96ca511637940a73d187ef4e7/CodePush/index.android.bundle"

The code snippet in my app is as below.

**Javascript:**
codePush.sync(null,null,this.onCodePushDownloadProgress,this.onCodePushError);

**Download Progress**
I/ReactNativeJS( 1434): [CodePush] Downloading package.
I/ReactNativeJS( 1434): Code Push Downloading 15920 of 401488
I/ReactNativeJS( 1434): Code Push Downloading 32304 of 401488
I/ReactNativeJS( 1434): Code Push Downloading 48688 of 401488
I/ReactNativeJS( 1434): Code Push Downloading 65072 of 401488
I/ReactNativeJS( 1434): Code Push Downloading 81456 of 401488
I/ReactNativeJS( 1434): Code Push Downloading 97840 of 401488

Another instance, stuck in 
I/ReactNativeJS( 1434): [CodePush] Downloading package.

Restarting of the emulator (Genymotion Android 4.4.2) gets around the error.

There is one rolled-back update available locally but I've seen message in previous successful updates that it's been ignored. Code Push has been working fine under good network condition.

Environment
RN 0.24 Android
CodePush ^1.10.6-beta
Tested with RN dev/debug pkg

geof90:
Hi @ACCTFORGH, did you see the app crash at all, or are these just from the debug logs? If the app did not crash, the download could have simply timed out due to poor network conditions, and reverted to the previous version, which is fine and expected behaviour.

[...]

PR #341, opened 2016-05-19
Remove "More than one CodePush instance" message

geof90:
This PR simply removes the "More than one CodePush instance" error because it is confusing as indicated in https://github.com/Microsoft/react-native-code-push/issues/335.

### DIFF
--- a/android/app/src/main/java/com/microsoft/codepush/react/CodePush.java
+++ b/android/app/src/main/java/com/microsoft/codepush/react/CodePush.java
@@ -104,10 +104,6 @@ public class CodePush implements ReactPackage {
             throw new CodePushUnknownException("Unable to get package info for " + applicationContext.getPackageName(), e);
         }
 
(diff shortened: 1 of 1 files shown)


lostintangent:
LGTM!

android/app/src/main/java/com/microsoft/codepush/react/CodePush.java
```diff
@@ -104,10 +104,6 @@ public CodePush(String deploymentKey, Activity mainActivity, boolean isDebugMode
             throw new CodePushUnknownException("Unable to get package info for " + applicationContext.getPackageName(), e);
         }
 
-        if (currentInstance != null) {
```
scottbommarito: i would prefer it if we left a comment describing how currentInstance could not be null but ultimately it doesn't matter

scottbommarito:
`// LGTM`

PR #342, opened 2016-05-19
Fix rollback status report

geof90:
The RN plugin is missing this crucial logic in the Cordova Plugin (https://github.com/Microsoft/cordova-plugin-code-push/blob/master/src/ios/CodePushReportingManager.m#L90) that ensures that we do not erroneously record an upgrade to a new version in the case of a rollback status report.


lostintangent:
LGTM!

scottbommarito:
`// LGTM`

PR #343, opened 2016-05-22
Add ability to delay restarts

dbasedow:
Added a new InstallMode ON_NEXT_RESTART_OPPORTUNITY which works a bit like IMMEDIATE except that you can define when restarts can happen.

CodePush.allowRestart() and CodePush.disallowRestart() can be used to define if restarts are allowed right now or not. If CodePush.disallowRestart() was called and sync() is called with InstallMode ON_NEXT_RESTART_OPPORTUNITY the installation will start immediately but there will be no restart until CodePush.allowRestart() is called.

### DIFF
--- a/CodePush.js
+++ b/CodePush.js
@@ -415,7 +415,7 @@ if (NativeCodePush) {
     InstallMode: {
       IMMEDIATE: NativeCodePush.codePushInstallModeImmediate, // Restart the app immediately
       ON_NEXT_RESTART: NativeCodePush.codePushInstallModeOnNextRestart, // Don't artificially restart the app. Allow the update to be "picked up" on the next app restart
-      ON_NEXT_RESUME: NativeCodePush.codePushInstallModeOnNextResume, // Restart the app the next time it is resumed from the background
+      ON_NEXT_RESUME: NativeCodePush.codePushInstallModeOnNextResume // Restart the app the next time it is resumed from the background
     },
     SyncStatus: {
       CHECKING_FOR_UPDATE: 0,
--- a/CodePush.js
+++ b/CodePush.js
@@ -214,10 +214,6 @@ async function tryReportStatus(resumeListener) {
   }
 }
 
-function restartApp(onlyIfUpdateIsPending = false) {
-  NativeCodePush.restartApp(onlyIfUpdateIsPending);
-}
-
 var testConfig;
 
 // This function is only used for tests. Replaces the default SDK, configuration and native bridge
@@ -410,7 +406,7 @@ if (NativeCodePush) {
     log,
     notifyAppReady: notifyApplicationReady,
     notifyApplicationReady,
-    restartApp,
+    restartApp: RestartManager.restartApp,
     setUpTestDependencies,
     sync,
     disallowRestart: RestartManager.disallow,
@@ -420,7 +416,6 @@ if (NativeCodePush) {
       IMMEDIATE: NativeCodePush.codePushInstallModeImmediate, // Restart the app immediately
       ON_NEXT_RESTART: NativeCodePush.codePushInstallModeOnNextRestart, // Don't artificially restart the app. Allow the update to be "picked up" on the next app restart
       ON_NEXT_RESUME: NativeCodePush.codePushInstallModeOnNextResume, // Restart the app the next time it is resumed from the background
-      ON_NEXT_RESTART_OPPORTUNITY: NativeCodePush.codePushInstallModeOnNextRestartOpportunity
     },
     SyncStatus: {
       CHECKING_FOR_UPDATE: 0,
--- a/CodePush.js
+++ b/CodePush.js
@@ -3,6 +3,7 @@ import { Alert } from "./AlertAdapter";
 import requestFetchAdapter from "./request-fetch-adapter";
 import { AppState, Platform } from "react-native";
 import RestartManager from "./RestartManager";
+import log from './logging';
 
 let NativeCodePush = require("react-native").NativeModules.CodePush;
 const PackageMixins = require("./package-mixins")(NativeCodePush);
@@ -153,11 +154,6 @@ function getPromisifiedSdk(requestFetchAdapter, config) {
   };
 
   return sdk;
-}
-
-/* Logs messages to console with the [CodePush] prefix */
-function log(message) {
-  console.log(`[CodePush] ${message}`)
 }
 
 // This ensures that notifyApplicationReadyInternal is only called once
@@ -411,7 +407,6 @@ if (NativeCodePush) {
     sync,
     disallowRestart: RestartManager.disallow,
     allowRestart: RestartManager.allow,
-    restartAllowed: RestartManager.allowed,
     InstallMode: {
       IMMEDIATE: NativeCodePush.codePushInstallModeImmediate, // Restart the app immediately
       ON_NEXT_RESTART: NativeCodePush.codePushInstallModeOnNextRestart, // Don't artificially restart the app. Allow the update to be "picked up" on the next app restart
--- a/CodePush.js
+++ b/CodePush.js
@@ -2,6 +2,7 @@ import { AcquisitionManager as Sdk } from "code-push/script/acquisition-sdk";
 import { Alert } from "./AlertAdapter";
 import requestFetchAdapter from "./request-fetch-adapter";
 import { AppState, Platform } from "react-native";
+import RestartManager from "./RestartManager";
 
 let NativeCodePush = require("react-native").NativeModules.CodePush;
 const PackageMixins = require("./package-mixins")(NativeCodePush);
@@ -412,10 +413,14 @@ if (NativeCodePush) {
     restartApp,
     setUpTestDependencies,
     sync,
+    disallowRestart: RestartManager.disallow,
+    allowRestart: RestartManager.allow,
+    restartAllowed: RestartManager.allowed,
     InstallMode: {
       IMMEDIATE: NativeCodePush.codePushInstallModeImmediate, // Restart the app immediately
       ON_NEXT_RESTART: NativeCodePush.codePushInstallModeOnNextRestart, // Don't artificially restart the app. Allow the update to be "picked up" on the next app restart
-      ON_NEXT_RESUME: NativeCodePush.codePushInstallModeOnNextResume // Restart the app the next time it is resumed from the background
+      ON_NEXT_RESUME: NativeCodePush.codePushInstallModeOnNextResume, // Restart the app the next time it is resumed from the background
+      ON_NEXT_RESTART_OPPORTUNITY: NativeCodePush.codePushInstallModeOnNextRestartOpportunity
     },
     SyncStatus: {
       CHECKING_FOR_UPDATE: 0,
--- a/RestartManager.js
+++ b/RestartManager.js
@@ -2,22 +2,16 @@ let NativeCodePush = require("react-native").NativeModules.CodePush;
 
 const RestartManager = (() => {
     let _allowed = true;
-    let restartPending = false;
 
-    function tryRestart() {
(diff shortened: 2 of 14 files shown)


RestartManager.js
```diff
@@ -0,0 +1,39 @@
+let NativeCodePush = require("react-native").NativeModules.CodePush;
+
+const RestartManager = (() => {
+    let _allowed = true;
+    let restartPending = false;
+
+    function tryRestart() {
+        if (restartPending && _allowed == true) {
+            NativeCodePush.restartApp(true);
```
lostintangent: By pasing `true` to `restartApp`, you're already preventing the restart from happening unless there is a pending update. Do we need the `restartPending` flag at that point?

CodePush.js
```diff
@@ -412,10 +413,14 @@ if (NativeCodePush) {
...
     sync,
+    disallowRestart: RestartManager.disallow,
+    allowRestart: RestartManager.allow,
+    restartAllowed: RestartManager.allowed,
     InstallMode: {
       IMMEDIATE: NativeCodePush.codePushInstallModeImmediate, // Restart the app immediately
       ON_NEXT_RESTART: NativeCodePush.codePushInstallModeOnNextRestart, // Don't artificially restart the app. Allow the update to be "picked up" on the next app restart
-      ON_NEXT_RESUME: NativeCodePush.codePushInstallModeOnNextResume // Restart the app the next time it is resumed from the background
+      ON_NEXT_RESUME: NativeCodePush.codePushInstallModeOnNextResume, // Restart the app the next time it is resumed from the background
+      ON_NEXT_RESTART_OPPORTUNITY: NativeCodePush.codePushInstallModeOnNextRestartOpportunity
```
lostintangent: Do we actually need the new install mode for this? The install mode represents when the update should be attempted to be installed automatically, but it may happen sooner (e.g. if the app calls `restartApp`), and it could also happen later (e.g. if you specified a `minimumBackgroundDurarion`). The `ON_NEXT_RESTART` install mode basically disables automatically forcing a restart, so the name of this new install mode is also potentially a little confusing, since it's behavior is more closely aligned with `IMMEDIATE`.

I would vote to remove this install mode, since I think this new behavior is conceptually similar to what `minimumBackgroundDuration` already does for resume-based installs, and therefore, it seems fine to have a way to delay immediate installs without requiring a new install mode.

As an aside, should `restartApp` even be allowed to be called if you've explicitly disallowed restarts? It feels like `restartApp` is really the "primitive" that allows restarting the app, and if we're adding the ability to disallow it, then maybe that's where the delay behavior should go. Then, anywhere that calls `restartApp` would simply be allowed or not, but they wouldn't need to worry about interacting with the `RestartManager`.

RestartManager.js
dbasedow: You're right it's not needed.

CodePush.js
dbasedow: I agree that the new install mode is very closely related to IMMEDIATE. I initially wanted to name it ASAP but that seemed even more confusing. At least for my use case a time based delay will not be sufficient, I want to be able to only have restarts happen at certain points.

Moving the decision to restartApp makes complete sense, but the state still has to be kept somewhere with functions for mutating it. Where would you put it?

lostintangent: Maybe move `restartApp` into the `RestartManager`? That way, that module handles the policy of when a restart is allowed and there behavior of a restart itself. Then everywhere that currently internally calls the native bridge's `restartApp` directly could just start calling the version within the `RestartManager`.

dbasedow: I'll implement that.

If I remove the new install mode there is another potentially confusing issue: if I disable restarts and then receive a mandatory update it will also be delayed. I'm not sure if this is actually an issue but I could imagine the case that I want normal updates as soon as a restart is allowed but mandatory updates should still force an immediate restart. For my project it's not a problem, I'm not sure what other users need.

lostintangent: I think this behavior is fine. If the need ever arises, I suppose we could allow `disallowRestart` to specify whether mandatory updates are allowed to "force" a restart that would otherwise be disallowed.

RestartManager.js
```diff
@@ -0,0 +1,33 @@
...
+const RestartManager = (() => {
+    let _allowed = true;
+
+    function restartApp(onlyIfUpdateIsPending = false) {
+        if (_allowed) {
+            NativeCodePush.restartApp(onlyIfUpdateIsPending);
+        }
+    }
+
+    function allow() {
```
lostintangent: Now that I think about it, you were right to track the state of whether a restart was requested. If an app calls this method, but an update was installed with a restart-based install mode, I'm not sure if it makes sense to trigger a restart. It seems like it makes sense for this method to toggle the flag and only call restart if an update is pending that wanted to programmatically restart the app in the first place, but couldn't because it was previously disallowed.

lostintangent:
Thanks BTW for this awesome and thorough PR :) Before commentinf any more, I want to spend a little time tomorrow playing around with this implementation to see how it "feels". 

Could you also share a little bit of code from the app you're working on, to help me understand exactly how you're using this? Are you calling disallowRestart at the beginning of an "onboarding" flow and then calling allowRestart when the end user finishes it? Where is your call to sync at? In a different component?

Thanks so much!

dbasedow:
I have currently two points in the app where I disable restarts. The on-boarding is very simple, I use a container component that wraps the NavigationCardStack for all the on-boarding screens. The container has a componentDidMount and componentWillUnmount for allowing/disallowing restarts. The sync call is in the app root's componentDidMount (I want to experiment with more frequent calls).  For the first start this works really well. If an update is available it gets downloaded during the on-boarding flow and once the on-boarding is finished the app reloads and you see exactly what you're supposed to see.

The other case is specific to our app. The app I am working on does workout tracking. You can imagine that people will be extremely annoyed if their workout "crashes" (that's what it looks like). Right now I am also using componentDidMount and componentWillUnmount but on a single screen in the flow. So let's say I have a Tracking screen and after that a Result screen. If I put the componentDidMount and componentWillUnmount in the Tracking screen, the restarts will stay disabled until the Tracking screen is popped from the navigation stack. This is definitely not a good way to do it. I think for cases like this there should always be a container component.

I can also imagine cases that are not bound to components, like file uploads or other background operations. My initial implementation allowed nesting allowRestart/disallowRestart for cases like this. I removed it because I think it could be confusing and if needed is trivial to implement.

lostintangent:
@dbasedow Makes sense, thanks a lot for providing that detail. How are you currently using `restartAllowed`?

RestartManager.js
```diff
@@ -0,0 +1,33 @@
+let NativeCodePush = require("react-native").NativeModules.CodePush;
+
+const RestartManager = (() => {
+    let _allowed = true;
+
+    function restartApp(onlyIfUpdateIsPending = false) {
```
lostintangent: If the restart isn't currently allowed, I wonder if we should indicate that to the caller somehow. Maybe it isn't too important for now, but it's something to keep in mind.

dbasedow:
restartAllowed is only used in the example app to make a toggle button. I'm not sure if anyone will need it for anything other than debugging.

lostintangent:
I wonder whether we can add the appropriate logging to the `restartApp`, `disallowRestart` and `allowRestart` methods, and ensure that `restartApp` resolves/rejects it's returned promise/etc. to assist in debugging, and remove the `restartAllowed` method.

CodePush.js
```diff
@@ -409,13 +406,16 @@ if (NativeCodePush) {
...
     setUpTestDependencies,
     sync,
+    disallowRestart: RestartManager.disallow,
+    allowRestart: RestartManager.allow,
+    restartAllowed: RestartManager.allowed,
     InstallMode: {
       IMMEDIATE: NativeCodePush.codePushInstallModeImmediate, // Restart the app immediately
       ON_NEXT_RESTART: NativeCodePush.codePushInstallModeOnNextRestart, // Don't artificially restart the app. Allow the update to be "picked up" on the next app restart
-      ON_NEXT_RESUME: NativeCodePush.codePushInstallModeOnNextResume // Restart the app the next time it is resumed from the background
+      ON_NEXT_RESUME: NativeCodePush.codePushInstallModeOnNextResume, // Restart the app the next time it is resumed from the background
```
scottbommarito: don't forget to remove this comma given that we're no longer doing ON_NEXT_RESTART_OPPORTUNITY

RestartManager.js
scottbommarito: definitely this. `return _allowed`?

RestartManager.js
```diff
@@ -0,0 +1,33 @@
...
+            NativeCodePush.restartApp(onlyIfUpdateIsPending);
+        }
+    }
+
+    function allow() {
+        _allowed = true;
+        restartApp(true);
+    }
+
+    function allowed() {
```
scottbommarito: i think `isAllowed` would be a better name for this

lostintangent: We're discussing removing this method altogether, so we can discuss the name if we decide to leave it in.

typings/react-native-code-push.d.ts
```diff
@@ -217,6 +217,21 @@ declare namespace CodePush {
...
+
+    /**
+     * Forbid CodePush to restart the app.
+     */
+    function disallowRestart(): void;
+
+    /**
+     * Check if CodePush is allowed to restart the app.
+     */
+    function restartAllowed(): boolean;
```
scottbommarito: `isRestartAllowed`?

RestartManager.js
dbasedow: Since I am now exposing the onlyIfUpdateIsPending a simple flag no longer suffices. Is there really a case where onlyIfUpdateIsPending is set to false? installUpdate always calls savePendingUpdate regardless of InstallMode. The only use of restartApp(false) I could find was in local package's install(). But that happens after installUpdate.

scottbommarito: i don't think it's necessary to track whether or not a restart was requested. i think what needs to happen is if an update is pending and its restart mode is IMMEDIATE then it needs to restart the app here, and otherwise shouldn't. tracking whether a restart was requested gets close to that behavior, but i'd be worried that it may lead to issues with situations like the following:
```
- restarts are disallowed
- IMMEDIATE update gets downloaded, wants to restart but is rejected
- ON_NEXT_RESUME update gets downloaded, doesn't try to restart
- restarts are allowed again
- app restarts because a restart was rejected while it was disallowed
```
this shouldn't happen because the pending update is a ON_NEXT_RESUME update

typings/react-native-code-push.d.ts
lostintangent: Like I mentioned above, I'd prefer to remove this method entirely if we can.

RestartManager.js
lostintangent: @scottbommarito I agree we would need to accommodate that scenario, but that doesn't mean we shouldn't account for resume-based installs as well. If we're going to have a feature for disallowing synthetic restarts, it doesn't make sense for it to be specific to updates installed via `IMMEDIATE`, since that is just one way of triggering it.

The `RestartManager` could easily have a private method called `clearPendingRestart` (or whatever) that is called anytime the `LocalPackage.install` method is called with a non-immediate mode.

lostintangent:
LGTM! Thanks a lot for this contribution. I really really like this feature :)

geof90:
LGTM!

lostintangent:
@dbasedow Even though it would be pretty trivial for folks to create themselves, I wonder whether it wouldn't be cool to create a standard container component that handles the mount/dismount logic of disallowing/allowing restarts in a declarative way. That way, you wouldn't even need to use these new APIs, you just wrap your component in a `<codePush.RestartBoundary>` (crappy name for example only) and let it handle the details for you.

dbasedow:
That's a good idea. I will try to come up with something.

Issue #344, opened 2016-05-23
The telemetry manager re-reports a running binary version every time the app is restarted


lostintangent:
**UPDATE 2:** The real bug here is that binary version reports are never marked as having been reported, so every time you restart the app when running on a binary version, the CodePush plugin will keep reporting it.

<img width="630" alt="screen shot 2016-05-23 at 2 04 46 pm" src="https://cloud.githubusercontent.com/assets/116461/15484599/9e48f026-20ef-11e6-84c3-4c28531f0f2d.png">

**UPDATE:** I confirmed that this issue doesn't repro when using the latest NPM release, so I'm assuming we regressed this scenario with the fix in master that address rollback reporting. The reason appears to be because the deployment success report no longer sends the **previousLabelOrAppVersion** field when migrating from a binary release to a CodePush update.

**Repro steps:**

1. Ensure that your app is running the latest bits from GitHub not NPM (I believe this issue is a regression in a recent commit)
1. Clear your app deployment via `code-push deployment clear` so that all releases and install metrics are flushed (this makes the repro easier)
1. Uninstall the app from the device/emulator to ensure that it will re-report the new user/installation
1. Deploy the app (e.g. `react-native run-android`) and run it to ensure that it reports itself to the server
1. Make a JS change and release a CodePush update
1. Sync the app with the server, but don't trigger a restart (e.g. make sure you're using `InstallMode.ON_NEXT_RESTART`)
1. View the release history and see the expected single pending install

    <img width="644" alt="screen shot 2016-05-23 at 9 49 28 am" src="https://cloud.githubusercontent.com/assets/116461/15477833/c2847812-20cb-11e6-9686-8bea95d8121c.png">

1. Restart the app to apply the update
1. View the release history

    <img width="621" alt="screen shot 2016-05-23 at 9 50 57 am" src="https://cloud.githubusercontent.com/assets/116461/15477879/0807d0fa-20cc-11e6-9493-49cfa4ff8cec.png">

**Expected:** To see `Active: 100% (1 of 1)` in the `Install Metrics` column of the `v1` release
**Actual:** It displays `Active: 50% (1 of 2)`

Not sure if this is helpful at all, but here is the log from the successful install report:

[...]

PR #346, opened 2016-05-23
Metrics fix

lostintangent:
Addresses #344. This issue occurres because binary version reports don't include a status field, and therefore, aren't being properly recorded as having been reported.

This is a regression that was caused by [this commit](https://github.com/Microsoft/react-native-code-push/commit/66a36071e537cf8b3b6c0c8ec5c275c952fc2370), which sought to prevent recording rollback reports, but in the process, also prevented recording binary version reports as well.

### DIFF
--- a/android/app/src/main/java/com/microsoft/codepush/react/CodePushTelemetryManager.java
+++ b/android/app/src/main/java/com/microsoft/codepush/react/CodePushTelemetryManager.java
@@ -114,8 +114,9 @@ public class CodePushTelemetryManager {
 
     public void recordStatusReported(ReadableMap statusReport) {
         // We don't need to record rollback reports, so exit early if that's what was specified.
-        if (statusReport.hasKey(STATUS_KEY) && DEPLOYMENT_FAILED_STATUS.equals(statusReport.getString(STATUS_KEY)))
+        if (statusReport.hasKey(STATUS_KEY) && DEPLOYMENT_FAILED_STATUS.equals(statusReport.getString(STATUS_KEY))) {
             return;
+        }
         
         if (statusReport.hasKey(APP_VERSION_KEY)) {
             saveStatusReportedForIdentifier(statusReport.getString(APP_VERSION_KEY));
--- a/android/app/src/main/java/com/microsoft/codepush/react/CodePushTelemetryManager.java
+++ b/android/app/src/main/java/com/microsoft/codepush/react/CodePushTelemetryManager.java
@@ -113,13 +113,15 @@ public class CodePushTelemetryManager {
     }
 
     public void recordStatusReported(ReadableMap statusReport) {
-        if (statusReport.hasKey(STATUS_KEY) && DEPLOYMENT_SUCCEEDED_STATUS.equals(statusReport.getString(STATUS_KEY))) {
-            if (statusReport.hasKey(APP_VERSION_KEY)) {
-                saveStatusReportedForIdentifier(statusReport.getString(APP_VERSION_KEY));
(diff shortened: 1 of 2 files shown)


android/app/src/main/java/com/microsoft/codepush/react/CodePushTelemetryManager.java
```diff
@@ -113,13 +113,15 @@ public WritableMap getUpdateReport(WritableMap currentPackage) {
...
-        if (statusReport.hasKey(STATUS_KEY) && DEPLOYMENT_SUCCEEDED_STATUS.equals(statusReport.getString(STATUS_KEY))) {
-            if (statusReport.hasKey(APP_VERSION_KEY)) {
-                saveStatusReportedForIdentifier(statusReport.getString(APP_VERSION_KEY));
-            } else if (statusReport.hasKey(PACKAGE_KEY)) {
-                String packageIdentifier = getPackageStatusReportIdentifier(statusReport.getMap(PACKAGE_KEY));
-                saveStatusReportedForIdentifier(packageIdentifier);
-            }
+        // We don't need to record rollback reports, so exit early if that's what was specified.
+        if (statusReport.hasKey(STATUS_KEY) && DEPLOYMENT_FAILED_STATUS.equals(statusReport.getString(STATUS_KEY)))
+            return;
```
scottbommarito: i generally avoid multi-line if statements without brackets, because
```
if (a)
    b;
```
becomes
```
if (a)
    b;
    c;
```
without much thought, which leads to issues. i would just append the `return;` on the end of the single line.

android/app/src/main/java/com/microsoft/codepush/react/CodePushTelemetryManager.java
```diff
@@ -113,13 +113,15 @@ public WritableMap getUpdateReport(WritableMap currentPackage) {
...
     public void recordStatusReported(ReadableMap statusReport) {
-        if (statusReport.hasKey(STATUS_KEY) && DEPLOYMENT_SUCCEEDED_STATUS.equals(statusReport.getString(STATUS_KEY))) {
-            if (statusReport.hasKey(APP_VERSION_KEY)) {
-                saveStatusReportedForIdentifier(statusReport.getString(APP_VERSION_KEY));
-            } else if (statusReport.hasKey(PACKAGE_KEY)) {
-                String packageIdentifier = getPackageStatusReportIdentifier(statusReport.getMap(PACKAGE_KEY));
-                saveStatusReportedForIdentifier(packageIdentifier);
-            }
+        // We don't need to record rollback reports, so exit early if that's what was specified.
+        if (statusReport.hasKey(STATUS_KEY) && DEPLOYMENT_FAILED_STATUS.equals(statusReport.getString(STATUS_KEY)))
```
scottbommarito: should we add this logic to the cordova plugin as well?

scottbommarito:
`// LGTM`

android/app/src/main/java/com/microsoft/codepush/react/CodePushTelemetryManager.java
lostintangent: Fine with me :)

android/app/src/main/java/com/microsoft/codepush/react/CodePushTelemetryManager.java
lostintangent: I'm not sure if the original bug exists in Cordova. If so, then yeah, we should add it there as well.

geof90:
LGTM!

Issue #337, opened 2016-05-17
Working with ZipArchive


wrlqwe:
CodePush and ZipArchive are used in my project with CocoaPods, they both import the "aes" and "minizip" libs which made conflicts, so how could I do to make them work well?

geof90:
@wrlqwe I believe https://github.com/Microsoft/react-native-code-push/issues/241 is a similar issue, let me know if the solution steps in that issue work for you.

wrlqwe:
@geof90 sorry for the late reply, I haven't use this solution because each developers in our team have to configure this, I fix this by wrote an script to remove the minizip and aes libs refs in CodePush's podspec and link it to npm's post-install script, everyone's 'pod install' will install CodePush without these 2 libs, so everyone would be happy. 😁

lostintangent:
@wrlqwe Would it be helpful if we updated our podspec to include a subspec that exclude these? That way, you wouldn't need the post-install script. You would simply need to update your pod file to depend on the "no-SSZipArchive" version of CodePush.

lostintangent:
@wrlqwe Regardless if you keep your private script, I wanted to take this opportunity to improve this situation for future users. Could you take a quick look at #348 and see if that looks reasonable? Thanks!

wrlqwe:
@lostintangent may be you could use spec.dependency to manage SSZipArchive. but this problem is conflicts between ZipArchive and SSZipArchive after all, because they both using minizip and aes by containing them in their pods, not CodePush's problem...

may be our team should not use mattconnolly/ZipArchive any more, it's inactive for a long time.

lostintangent:
@wrlqwe The reason we aren't using CocoaPods to manage our `SSZipArchive` dependency is so that React Native users who aren't using CocoaPods don't need to run `pod install`. I like that CodePush is fully self-contained for most of our users. That said, if someone is directly using CocoaPods, then it definitely makes sense that we allow linking with CodePush without `SSZipArchive` included.

lostintangent:
I'm going to close this via #348. It sounds like you're unblocked via your own scripts, but let us know if you run into any other problems or if the pod spec change doesn't suit your needs. Thanks!

chirag04:
@lostintangent can we do something for those who are not using CocoaPods ?

sidferreira:
It should be reopened...

PR #348, opened 2016-05-24
Adding nozip option to podspec

lostintangent:
This addresses issue #337 by adding a new subspace for CocoaPods users to easily exclude the `SSZipArchive` dependency.

### DIFF
--- a/CodePush.podspec
+++ b/CodePush.podspec
@@ -12,10 +12,19 @@ Pod::Spec.new do |s|
   s.homepage            = 'http://microsoft.github.io/code-push/'
   s.source              = { :git => 'https://github.com/Microsoft/react-native-code-push.git', :tag => "v#{s.version}-beta" }
   s.platform            = :ios, '7.0'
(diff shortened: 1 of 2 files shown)


wrlqwe:
This is an great idea, It's helpful!!

I think use SSZipArchive by spec.dependency is useful. can it be an default setting?
 Let pod manage the dependency is better than linking files directly.

lostintangent:
@wrlqwe The primary reason we haven't used `spec.dependency` up to this point is so that the CodePush NPM package is entirely self-contained and doesn't require CocoaPods to resolve any dependencies. I'd prefer to leave it this way for now, since most of our users aren't using CocoaPods are part of their dev workflow.

wrlqwe:
@lostintangent got it, thanks for your great work~

lostintangent:
@wrlqwe OK cool :) Would this fix allow you and your team to remove your post-install NPM scripts and use CodePush?

PR #350, opened 2016-05-25
Adding docs for dis/allowRestart

lostintangent:
This PR documents the `allowRestart` and `disallowRestart` methods that were contributed by @dbasedow, so that we can release them later this week.

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,7 +124,8 @@ We hope to eventually remove the need for steps #2-4, but in the meantime, RNPM 
     pod 'CodePush', :path => './node_modules/react-native-code-push', :subspecs => ['NoZip']
     ```
     
-    *NOTE: The above paths needs to be relative to your app's `Podfile`, so adjust it as neccessary.*
+    *NOTE: The above paths needs to be relative to your app's `Podfile`, so adjust it as nec
+    cessary.*
     
 2. Run `pod install`
 
@@ -409,7 +410,7 @@ code-push release-react MyApp ios
 code-push release-react MyApp-Android android
 ```
 
-The `release-react` command enables such a simple workflow because it provides many sensible defaults (e.g. generating a release bundle, assuming your app's entry file on iOS is either `index.ios.js` or `index.js`). However, all of these defaults can be customized to allow incremental flexibility as neccessary, which makes it a good fit for most scenarios. 
+The `release-react` command enables such a simple workflow because it provides many sensible defaults (e.g. generating a release bundle, assuming your app's entry file on iOS is either `index.ios.js` or `index.js`). However, all of these defaults can be customized to allow incremental flexibility as necessary, which makes it a good fit for most scenarios. 
 
 ```shell
 # Release a mandatory update with a changelog
@@ -450,11 +451,11 @@ The following sections describe the shape and behavior of these APIs in detail:
 
 When you require `react-native-code-push`, the module object provides the following top-level methods:
 
-* [allowRestart](#codepushallowrestart): Re-allows programmatic restarts to occur as a result of an update being installed, and optionally, immediately restarts the app if a pending update had attempted to restart the app while restarts were disallowed. This is an advanced API and is only neccessary if your app explicitly disallowed restarts via the `disallowRestart` method.
+* [allowRestart](#codepushallowrestart): Re-allows programmatic restarts to occur as a result of an update being installed, and optionally, immediately restarts the app if a pending update had attempted to restart the app while restarts were disallowed. This is an advanced API and is only necessary if your app explicitly disallowed restarts via the `disallowRestart` method.
 
 * [checkForUpdate](#codepushcheckforupdate): Asks the CodePush service whether the configured app deployment has an update available. 
 
-* [disallowRestart](#codepushdisallowrestart): Temporarily disallows any programmatic restarts to occur as a result of a CodePush update being installed. This is an advanced API, and is useful when a component within your app (e.g. an onboarding wizard) needs to ensure that no end-user interruptions can occur during its lifetime.
+* [disallowRestart](#codepushdisallowrestart): Temporarily disallows any programmatic restarts to occur as a result of a CodePush update being installed. This is an advanced API, and is useful when a component within your app (e.g. an onboarding process) needs to ensure that no end-user interruptions can occur during its lifetime.
 
 * [getCurrentPackage](#codepushgetcurrentpackage): Retrieves the metadata about the currently installed update (e.g. description, installation time, size). *NOTE: As of `v1.10.3-beta` of the CodePush module, this method is deprecated in favor of [`getUpdateMetadata`](#codepushgetupdatemetadata)*.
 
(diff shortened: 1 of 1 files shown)


shishirx34:
This feature is pretty cool. LGTM! :shipit:

README.md
```diff
@@ -450,8 +450,12 @@ The following sections describe the shape and behavior of these APIs in detail:
 
 When you require `react-native-code-push`, the module object provides the following top-level methods:
 
+* [allowRestart](#codepushallowrestart): Re-allows programmatic restarts to occur as a result of an update being installed, and optionally, immediately restarts the app if a pending update had attempted to restart the app while restarts were disallowed. This is an advanced API and is only neccessary if your app explicitly disallowed restarts via the `disallowRestart` method.
```
scottbommarito: "necessary"?

README.md
```diff
@@ -450,8 +450,12 @@ The following sections describe the shape and behavior of these APIs in detail:
...
+
 * [checkForUpdate](#codepushcheckforupdate): Asks the CodePush service whether the configured app deployment has an update available. 
 
+* [disallowRestart](#codepushdisallowrestart): Temporarily disallows any programmatic restarts to occur as a result of a CodePush update being installed. This is an advanced API, and is useful when a component within your app (e.g. an onboarding wizard) needs to ensure that no end-user interruptions can occur during its lifetime.
```
scottbommarito: i think "onboarding wizard" is a little bit of a confusing term to use here, maybe "login process" or something more colloquial?

README.md
```diff
@@ -462,6 +466,30 @@ When you require `react-native-code-push`, the module object provides the follow
 
 * [sync](#codepushsync): Allows checking for an update, downloading it and installing it, all with a single call. Unless you need custom UI and/or behavior, we recommend most developers to use this method when integrating CodePush into their apps
 
+#### codePush.allowRestart
+
+```javascript
+codePush.allowRestart(): void;
+```
+
+Re-allows programmatic restarts to occur, that would have otherwise been rejected due to a previous call to `disallowRestart`. If `disallowRestart` was never called in the first place, then calling this method will simply result in a no-op.
```
scottbommarito: it's weird to me that in these docs, `allowRestart` is first and `disallowRestart` is second, when you'd always want to call them in the opposite order. i think it's fine given that they're alphabetized but it's something to consider

README.md
```diff
@@ -462,6 +466,30 @@ When you require `react-native-code-push`, the module object provides the follow
...
+
+If a CodePush update is currently pending, which attempted to restart the app (e.g. it used `InstallMode.IMMEDIATE`), but was blocked due to `disallowRestart` having been called, then calling `allowRestart` will result in an immediate restart. This allows the update to be applied as soon as possible, without interrupting the end user during critical workflows (e.g. an onboarding wizard).
```
scottbommarito: "login process" or something more colloquial than "onboarding wizard"?

README.md
```diff
@@ -462,6 +466,30 @@ When you require `react-native-code-push`, the module object provides the follow
...
+
+For example, calling `allowRestart` would trigger an immediate restart if either of the three scenarios mentioned in the [`disallowRestart` docs](#codepushdisallowrestart) occured after `disallowRestart` was called. However, calling `allowRestart` wouldn't trigger a restart if the following were true:
+
+1. No CodePush updates were installed since the last time `disallowRestart` was called, and therefore, there isn't any need to restart anyways.
+
+2. There is currently a pending CodePush update, but it was installed via `InstallMode.ON_NEXT_RESTART`, and therefore, doesn't require a programmatic restart.
+
+3. There is currently a pending CodePush update, but it was installed via `InstallMode.ON_NEXT_RESUME` and the app hasn't been put into the background yet, and therefore, there isn't a need to programmatically restart yet.
+
+3. No calls to `restartApp` were made since the last time `disallowRestart` was called.
```
scottbommarito: you have two "3"s LOL

README.md
shishirx34: "Onboarding wizard" is colloquial in software development. Look it up. Its the thing that fires up first time an app is started, and it shows you the steps to follow to get started. Just because you don't know about it doesn't mean its the wrong thing.

README.md
```diff
@@ -493,6 +521,49 @@ codePush.checkForUpdate()
...
 
+#### codePush.disallowRestart
+
+```javascript
+codePush.disallowRestart(): void;
+```
+
+Temporarily disallows programmatic restarts to occur as a result of either of following scenarios:
+
+1. A CodePush update is installed using `InstallMode.IMMEDIATE`
```
scottbommarito: why is this numbered list not separated by new lines while the one in `allowRestart` is?

README.md
scottbommarito: i did.

https://www.google.com/search?q=onboarding+wizard&ie=utf-8&oe=utf-8

unlike "watermark", i don't see any wikipedia or similar page describing what it means in a dictionary definition, just a bunch of articles where a couple people seem to describe their process as an "onboarding wizard" arbitrarily. if somebody, like me, who has never heard the term and doesn't know what it means, tries to look it up and its not immediately obvious to them what it is, then we should use a different term lol

README.md
```diff
@@ -493,6 +521,49 @@ codePush.checkForUpdate()
...
+
+This is an advanced API, and is primarily useful when individual components within your app (e.g. an onboarding wizard) need to ensure that no end-user interruptions can occur during their lifetime, while continuing to allow the app to keep syncing with the CodePush server at its own pace and using whatever install modes are appropriate. This has the benefit of allowing the app to discover and download available updates as soon as possible, while also preventing any disruptions during key end-user experiences.
+
+As an alternative, you could also choose to simply use `InstallMode.ON_NEXT_RESTART` whenever calling `sync` (which will never attempt to programmatically restart the app), and then explicity calling `restartApp` at points in your app that you know it is "safe" to do so. `disallowRestart` provides an alternative approach to this when the code that sychronizes with the CodePush server is seperate from the code/comopents that want to enforce a no-restart policy.
```
scottbommarito: "synchronizes", "separate", and "components"

README.md
```diff
@@ -493,6 +521,49 @@ codePush.checkForUpdate()
...
+
+Temporarily disallows programmatic restarts to occur as a result of either of following scenarios:
+
+1. A CodePush update is installed using `InstallMode.IMMEDIATE`
+2. A CodePush update is installed using `InstallMode.ON_NEXT_RESUME` and the app is resumed from the background (optionally being throttled by the `minimumBackgroundDuration` property)
+3. The `restartApp` method was called
+
+*NOTE: #1 and #2 effectively work by calling `restartApp` for you, so you can think of `disallowRestart` as blocking any call to `restartApp`, regardless if your app calls it directly or indirectly.*
+
+After calling this method, any calls to `sync` would still be allowed to check for an update, download it and install it, but an attempt to restart the app would be queued until `reallowRestart` is called. This way, the restart request is captured and can be "flushed" whenever you want to allow it to occur.
```
scottbommarito: i think i would mention here that if a second update is installed while one is waiting to restart, the old update gets ignored in favor of the new update, and if that new update doesn't require a restart, no restart will ahppen.

README.md
```diff
@@ -493,6 +521,49 @@ codePush.checkForUpdate()
...
+
+This is an advanced API, and is primarily useful when individual components within your app (e.g. an onboarding wizard) need to ensure that no end-user interruptions can occur during their lifetime, while continuing to allow the app to keep syncing with the CodePush server at its own pace and using whatever install modes are appropriate. This has the benefit of allowing the app to discover and download available updates as soon as possible, while also preventing any disruptions during key end-user experiences.
```
scottbommarito: (another "onboarding wizard" here in case decide to change to a different term)

README.md
lostintangent: As you mentioned, this is ordered alphabetically, not based on the sequence you would call them in. I don't see the reason why we would order things based on sequence in an API reference document. Additionally, as I call out, this is already an advanced API, so I don't think we need to optimize our doc flow specifically for this.

README.md
lostintangent: That's an edge case scenario that I decided not to call out explicitly, since this doc is already fairly long as it is. If we get a question about it, then I'll update it.

README.md
lostintangent: I'm fine with "process" over "wizard" (though I don't think it makes much of a difference), but "login" definitely doesn't capture the intent of the example, and the exact scenario that I've seen people implement this feature for in the past was for an on boarding process, not anything to do with login.

scottbommarito: yea i prefer "onboarding process" over login

PR #357, opened 2016-05-28
ADD AUTOMATED TESTING

scottbommarito:
Used the PluginTestingFramework to create tests for React-Native.

Also needed to create a gulpfile to build the tasks and compile the typescript test files.

A large portion of the PR is typings files that can be ignored.

TO DO: add plugintestingframework to npm and add as a dependency, (maybe) add some RN-specific tests, (eventually) improve test run time


CodePush.js
```diff
@@ -418,15 +418,15 @@ if (NativeCodePush) {
...
-      CHECKING_FOR_UPDATE: 0,
-      AWAITING_USER_ACTION: 1,
-      DOWNLOADING_PACKAGE: 2,
-      INSTALLING_UPDATE: 3,
-      UP_TO_DATE: 4, // The running app is up-to-date
-      UPDATE_IGNORED: 5, // The app had an optional update and the end-user chose to ignore it
-      UPDATE_INSTALLED: 6, // The app had an optional/mandatory update that was successfully downloaded and is about to be installed.
-      SYNC_IN_PROGRESS: 7, // There is an ongoing "sync" operation in progress.
-      UNKNOWN_ERROR: -1
+      UP_TO_DATE: 0, // The running app is up-to-date
```
scottbommarito: This was done so that the Cordova plugin and this plugin have identical numberings.

android/app/src/main/java/com/microsoft/codepush/react/CodePush.java
```diff
@@ -510,6 +516,10 @@ public void dispatchDownloadProgressEvent() {
                         e.printStackTrace();
                         saveFailedUpdate(updatePackage);
                         promise.reject(e);
+                    } catch (CodePushMalformedDataException e) {
```
scottbommarito: this was to fix a bug discovered during testing that threw this error but was never caught and crashed the app on startup

typings/assert.d.ts
```diff
@@ -0,0 +1,63 @@
+// Type definitions for assert and power-assert
```
scottbommarito: after this point, the rest of the PR is typings files

scottbommarito:
note: i have checked in the entirety of the code-push-plugin-testing-framework so that we can start testing the app while we wait for OSS approval

scottbommarito:
succeeds on VSTS!

nevir:
This PR is sacrilege! 👍

PR #363, opened 2016-06-05
Fix bug in podspec caused by spelling

tonyxiao:
This causes cocoapod based installation to fail.


lostintangent:
Thanks for the fix!

PR #364, opened 2016-06-06
installLocation property

lostintangent:
This PR adds a new property to the `LocalPackage` object called `installLocation`, which simply exposes the absolute path of the directory that the update's contents are installed within. This allows JS code to construct absolute URLs to files, which are relative to the current (or pending) update, and unblocks the use of APIs that require an absolute path, while still allowing updating these assets via CodePush.

For example, the following snippet shows how to use the [react-native-sound](https://github.com/zmxv/react-native-sound) plugin to play an MP3, in a way that also enables you to ship updates to the file via CodePush.

```javascript
const update = await codePush.getUpdateMetadata();
const sound = new Sound("sound.mp3", update.installLocation);
sound.play();
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -948,7 +948,7 @@ Contains details about an update that has been downloaded locally or already ins
 
 - __failedInstall__: Indicates whether this update has been previously installed but was rolled back. The `sync` method will automatically ignore updates which have previously failed, so you only need to worry about this property if using `checkForUpdate`. *(Boolean)*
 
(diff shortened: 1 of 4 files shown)


shishirx34:
LGTM

geof90:
LGTM!

lostintangent:
I'm going to close this PR for now since I think the better solution here is to send a PR to `react-native-sound` and other components to ensure they support the RN assets system, as opposed to allowing CodePush to "virtualize" assets on its own.

PR #367, opened 2016-06-07
Fix downloadProgressCallback (RCTFrameUpdateObserver bug)

mikedouglas:
Currently, downloadProgressCallback isn't called while CodePush is downloading an update. This is because the RCTFrameUpdateObserver isn't called, due to a bug with how the _paused ivar is used. After _paused is modified, React requires that the _pauseCallback function be fired. This pull request ensures that modifying the paused property leads to a _pauseCallback call.

### DIFF
--- a/ios/CodePush/CodePush.m
+++ b/ios/CodePush/CodePush.m
@@ -179,6 +179,16 @@ static NSString *bundleResourceName = @"main";
 @synthesize pauseCallback = _pauseCallback;
 @synthesize paused = _paused;
 
+- (void)setPaused:(BOOL)paused
+{
+    if (_paused != paused) {
+        _paused = paused;
+        if (_pauseCallback) {
+            _pauseCallback();
+        }
+    }
+}
+
 /*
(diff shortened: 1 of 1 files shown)


lostintangent:
Interesting. I didn't realize we had an issue here :) Thanks for submitting this fix. It's look great, let me just do some smoke testing and we can get it merged in.

mikedouglas:
We just started getting reports after upgrading from RN 0.24.1 to 0.26. Either the global pausing in RCTDisplayLink is a newer optimization, or there was another module that was previously subscribed to the frame updates.

lostintangent:
LGTM! I'll wait for @geof90 to take a look and merge it in. Thanks again for this fix.

geof90:
LGTM!

lostintangent:
@mikedouglas We just released this fix to NPM. Thanks so much for the contribution.

mikedouglas:
Thanks for the quick review, merge, and release!

Issue #362, opened 2016-06-01
DOWNLOADING_PACKAGE forever...


cnjsstong:
We've been consistently having issues with deploying new bundles. Existing users who have ever had downloaded a new push will see DOWNLOADING_PACKAGE forever. First-time users are fine. We are using 1.10.4-beta, both native and js parts.

lostintangent:
That is really bizarre! Are you seeing this on both Android and iOS?

lostintangent:
@cnjsstong Are you still hitting this issue? I'd love to make sure you're unblocked, and would love to get a little more details so we can try to repro the problem. Thanks!

lostintangent:
@cnjsstong We sent out a PR to prevent the call to `sync` from hanging like this, however, it doesn't address the root cause of why you're hitting this issue in the first place.

After the fix in #368 is merged, you would see the `DOWNLOADING_PACKAGE` status, and then the `Promise` returned by `sync` would resolve with `UNKNOWN_ERROR`, because the call was hanging due to an exception being thrown when the update is downloading. That said, you want the `sync` call to actually succeed, so we need to understand what's going on.

If you could let us know if you're experiencing this problem on iOS or Android, what version of React Native you're using, and what other RN plugins your app has installed, that would be hugely helpful to try to get a repro. Thanks!

lostintangent:
I'm going to close this for now since we can't unfortunately repro it, and we resolved the hang. If you experienced any instances where calling sync results in an exception, please let us know, and we can investigate the issue further. Thanks!

PR #368, opened 2016-06-08
Propagate errors thrown in update dialog button press handler

geof90:
This ensures errors thrown in the `onPress` handler of the update dialog will get propagated and result in the rejection of the `sync` promise, instead of resulting in an unhandled promise rejection and leaving the main `sync` promise hanging. The problem that this fixes might be a root cause for https://github.com/Microsoft/react-native-code-push/issues/362.

### DIFF
--- a/CodePush.js
+++ b/CodePush.js
@@ -354,7 +354,11 @@ async function syncInternal(options = {}, syncStatusChangeCallback, downloadProg
         const dialogButtons = [{
           text: null,
           onPress: async () => {
-            resolve(await doDownloadAndInstall());
+            try {
+              resolve(await doDownloadAndInstall());
+            } catch (downloadError) {
+              reject(downloadError);
+            }
           }
         }];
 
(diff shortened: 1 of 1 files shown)


CodePush.js
```diff
@@ -354,7 +354,11 @@ async function syncInternal(options = {}, syncStatusChangeCallback, downloadProg
         const dialogButtons = [{
           text: null,
           onPress: async () => {
-            resolve(await doDownloadAndInstall());
```
lostintangent: Would it be simpler to just do `doDownloadAndInstall().then(resolve, reject)`? The `Alert` module simply calls the `onPress` function, without regard to the fact that it could return a `Promise`, so it feels a little odd to make the arrow function `async` simply to be able to `await` the download call and propagate a rejection using `try/catch`.

lostintangent:
LGTM!

scottbommarito:
`// LGTM`

PR #369, opened 2016-06-08
Explicitly mark SSZipArchive headers as private

nevir:
This allows the pod to be built as a framework (the SSZipArchive headers were previously being included in the umbrella header for the framework by cocoapods, which causes compile errors "non modular header").  E.g. when the dependent's `Podfile` includes `use_frameworks!`

Also, moved SSZipArchive into its own subspec for clarity.

FYI I tested the few configurations I can where I'm _not_ using `use_frameworks!` - but you should verify this does't bust that

### DIFF
--- a/CodePush.podspec
+++ b/CodePush.podspec
@@ -17,13 +17,6 @@ Pod::Spec.new do |s|
 
   s.dependency 'React'
 
-  s.default_subspec = 'Full'
-
-  s.subspec 'Full' do |ss|
-    ss.dependency 'CodePush/Core'
-    ss.dependency 'CodePush/SSZipArchive'
-  end
-
   s.subspec 'Core' do |ss|
     ss.source_files = 'ios/CodePush/*.{h,m}'
     ss.public_header_files = ['ios/CodePush/CodePush.h']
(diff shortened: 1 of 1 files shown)


lostintangent:
@nevir This looks great! I really like your change to move the `SSZipArchive` bits to its own subspec.

Now that I see your change, I wonder whether we shouldn't rename the `NoZip` sub-spec to `Core` and then remove the `Full` subspec as well as the `default_subspec` setting. That way, depending on just `CodePush` gets you our bits + `SSZipArchive`, and if you didn't need/want `SSZipArchive`, you just depend on `CodePush/Core`, which actually seems cleaner than `CodePush/NoZip` and allows us to remove code from the `Podspec`.

nevir:
👍

lostintangent:
@nevir If we don't specify a `default_subspec`, is the default behavior to include all subspecs? If so, I wonder if we could remove that setting as well as the `Full` sub spec entirely, since they may be redundant.

nevir:
Oh, sorry missed that in your initial comment.  I'm not sure how it behaves! lemme test

nevir:
Works for framework builds!

lostintangent:
Beautiful! I love this change. Thanks so much dude :)

lostintangent:
@nevir Just released this fix to NPM. Thanks again for the contribution!

nevir:
awesome, thanks!

Issue #371, opened 2016-06-10
 The callback showDialog() exists in module CodePushDialog, but only one callback may be registered to a function in a native module.


oney:
I got ` The callback showDialog() exists in module CodePushDialog, but only one callback may be registered to a function in a native module.` crash of my Android app only.
I am using `react-native` `0.26.2` and `react-native-code-push` `1.12.0-beta`
The entire crash log from Crashlytics is
```
Fatal Exception: com.facebook.react.modules.core.JavascriptException: The callback showDialog() exists in module CodePushDialog, but only one callback may be registered to a function in a native module., stack:
value@62:4072
<unknown>@62:1996
g@62:451
value@62:1966

       at com.facebook.react.modules.core.ExceptionsManagerModule.showOrThrowError(ExceptionsManagerModule.java:90)
       at com.facebook.react.modules.core.ExceptionsManagerModule.reportFatalException(ExceptionsManagerModule.java:78)
       at java.lang.reflect.Method.invoke(Method.java)
       at com.facebook.react.bridge.BaseJavaModule$JavaMethod.invoke(BaseJavaModule.java:316)
       at com.facebook.react.bridge.NativeModuleRegistry$ModuleDefinition.call(NativeModuleRegistry.java:187)
       at com.facebook.react.bridge.NativeModuleRegistry.call(NativeModuleRegistry.java:62)
       at com.facebook.react.bridge.CatalystInstanceImpl$NativeModulesReactCallback.call(CatalystInstanceImpl.java:432)
       at com.facebook.react.bridge.queue.NativeRunnableDeprecated.run(NativeRunnableDeprecated.java)
       at android.os.Handler.handleCallback(Handler.java:739)
       at android.os.Handler.dispatchMessage(Handler.java:95)
       at com.facebook.react.bridge.queue.MessageQueueThreadHandler.dispatchMessage(MessageQueueThreadHandler.java:31)
       at android.os.Looper.loop(Looper.java:168)
       at com.facebook.react.bridge.queue.MessageQueueThreadImpl$3.run(MessageQueueThreadImpl.java:187)
       at java.lang.Thread.run(Thread.java:818)
```

geof90:
Seems like a race condition (user presses the button more than once, even though technically the dialog should get dismissed after the first button press). I will try to reproduce this issue and submit a fix if I can. Thanks for reporting!

oney:
Thanks!

PR #373, opened 2016-06-10
Prevent dialog from invoking callback twice

geof90:
This might fix https://github.com/Microsoft/react-native-code-push/issues/371, although I had trouble repro-ing it. For reference, the [React Native dialog module does something similar](https://github.com/facebook/react-native/blob/master/ReactAndroid/src/main/java/com/facebook/react/modules/dialog/DialogModule.java#L157).


lostintangent:
LGTM!

PR #379, opened 2016-06-15
Expose status reports via notifyAppReady()

nevir:
Handy for tracking whether a deployment succeeded/failed!  We're using this to emit analytics (Mixpanel) events on subsequent launches (success or rollback)

---

Note that I removed the `resumeListener` references, as it was completely unused; lemme know if you want me to preserve it instead!  Ditto the whitespace changes

### DIFF
--- a/CodePush.js
+++ b/CodePush.js
@@ -429,6 +429,10 @@ if (NativeCodePush) {
       PENDING: NativeCodePush.codePushUpdateStatePending,
       LATEST: NativeCodePush.codePushUpdateStateLatest
     },
+    DeploymentStatus: {
+      FAILED: "DeploymentFailed",
+      SUCCEEDED: "DeploymentSucceeded",
+    },
     DEFAULT_UPDATE_DIALOG: {
       appendReleaseDescription: false,
       descriptionPrefix: " Description: ",
--- a/CodePush.js
+++ b/CodePush.js
@@ -178,9 +178,7 @@ async function notifyApplicationReadyInternal() {
   return statusReport;
 }
 
-async function tryReportStatus(statusReport) {
-  if (!statusReport) return;
-
+async function tryReportStatus(statusReport, resumeListener) {
   const config = await getConfiguration();
   const previousLabelOrAppVersion = statusReport.previousLabelOrAppVersion;
   const previousDeploymentKey = statusReport.previousDeploymentKey || config.deploymentKey;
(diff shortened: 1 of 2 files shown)


scottbommarito:
Looks pretty good to me! Thanks for these changes!

lostintangent:
This looks great to me as well. We really appreciate all the recent contributions @nevir!

nevir:
👍

CodePush.js
```diff
@@ -172,50 +172,41 @@ const notifyApplicationReady = (() => {
...
-      NativeCodePush.recordStatusReported(statusReport);
-      resumeListener && AppState.removeEventListener("change", resumeListener);
-    } catch (e) {
-      log(`Report status failed: ${JSON.stringify(statusReport)}`);
-      NativeCodePush.saveStatusReportForRetry(statusReport);
-      // Try again when the app resumes
-      if (!resumeListener) {
-        resumeListener = (newState) => {
-          newState === "active" && tryReportStatus(resumeListener);
-        };
```
geof90: @nevir This `resumeListener` is our mechanism to retry status reports when the app resumes if the status report call to the server fails. Could we put it back?

nevir: Oho, yup!  Totally missed this bit of recursion

nevir: Alright, it's back via 04e84b5 - PTAL!

nevir: Would it be better if I _don't_ refresh the status (and just let the original report propagate through the status changes?)

geof90: No, I think refreshing it better mimics the original behaviour. Also, between the time it fails and the next time the app resumes, some other process might have already flushed the queue and reported it. So, what you have now is fine. :)

nevir: :+

typings/react-native-code-push.d.ts
```diff
@@ -146,77 +146,99 @@ export interface SyncOptions {
...
+
+    /**
+     * Deployment key used when deploying the previous package.
+     */
+    previousDeploymentKey?: string;
+
+    /**
+     * The label (v#) of the package that was upgraded from.
+     */
+    previousLabelOrAppVersion?: string;
```
geof90: There is also an `appVersion` key in the report, we also report binary updates. For more info: https://github.com/Microsoft/react-native-code-push/blob/master/ios/CodePush/CodePushTelemetryManager.m#L29

geof90: Also, the `package` key is optional

geof90:
It would be great if you could make the `.d.ts` changes as well, but as it is now it looks good to me!

nevir:
Just pushed the `.d.ts` tweaks

geof90:
Thanks!

lostintangent:
@nevir This has now been shipped to NPM via the `1.13.0-beta` release of this plugin. Thanks again for your contribution!

nevir:
sweet :D

PR #382, opened 2016-06-17
Fix incorrect links in README.md

arnif:
multi-deployment-releases does not exist it seems to be called multi-deployment-testing now, but the links were incorrect.


lostintangent:
Thanks so much for this fix!

Issue #381, opened 2016-06-16
Always error: main.jsbundle could not be found


jiyinyiyong:
```
Jun 16 17:37:04 yiyongs-iPod-touch dark_rn[410] <Notice>: [CodePush] Downloading package.
Jun 16 17:37:04 yiyongs-iPod-touch securityd[84] <Error>:  SecOCSPSingleResponseCreate OCSPResponse: single response has extension(s).
Jun 16 17:37:04 yiyongs-iPod-touch securityd[84] <Error>:  SecOCSPSingleResponseCreate OCSPResponse: single response has extension(s).
Jun 16 17:37:04 yiyongs-iPod-touch securityd[84] <Error>:  SecOCSPSingleResponseCreate OCSPResponse: single response has extension(s).
Jun 16 17:37:06 yiyongs-iPod-touch dark_rn[410] <Notice>: [CodePush] An unknown error occurred.
Jun 16 17:37:06 yiyongs-iPod-touch dark_rn[410] <Notice>: [CodePush] Update is invalid - A JS bundle file named "main.jsbundle" could not be found within the downloaded contents. Please check that you are releasing your CodePush updates using the exact same JS bundle file name that was shipped with your app's binary.
Jun 16 17:37:10 yiyongs-iPod-touch SpringBoard[54] <Warning>: [MPUSystemMediaControls] Disabling lock screen media controls updates for screen turning off.
```

Talked a bit on Discord and checked deployment key and `jsCodeLocation`, found no clue why it is happening.

jiyinyiyong:
I found the bug now in my code. Unexpected. There are **2** deployment keys in the package, the one in `info.plist` is correct, but the one in `CodePush.sync({deploymentKey: ...})` is wrong since it's common code for both platforms.

The error message is really misleading.

lostintangent:
Ah that would explain it. This is why I was asking on Discord if you could check if your key was for your Android app, since this error indicates that the update it's syncing doesn't appear to be a "valid" iOS application. I agree we could improve the message. For the most part, it is attempting to shed light on the fact that you released a JS bundle with a different name than your binary did, which we don't allow. Up until now, that is the primary error case, and to be fair, your scenario is a unique spin on that same issue.

[...]

PR #385, opened 2016-06-19
Improving "update is invalid" error message

lostintangent:
This addresses #381 by improving our existing "Update is invalid" message to also mention the need to verify that the configured deployment is correct.


geof90:
LGTM!

PR #387, opened 2016-06-20
fix double restart issue

scottbommarito:
fixes the issue on discord where `restartApp` was being called twice and crashing the app.

i verified that without my new changes (in `RestartManager`) the test case crashes, and with the change the test passes.


scottbommarito:
@geof90 @lostintangent

RestartManager.js
```diff
@@ -2,6 +2,7 @@ const log = require("./logging");
 const NativeCodePush = require("react-native").NativeModules.CodePush;
 
 const RestartManager = (() => {
+    let _in_progress = false;
```
lostintangent: I'd prefer to call this _restartInProgress, but either way is fine. But, we should use camel casing for the name to be consistent with others, and sort it alphabetically.

RestartManager.js
```diff
@@ -25,6 +26,11 @@ const RestartManager = (() => {
     }
 
     function restartApp(onlyIfUpdateIsPending = false) {
+        if (_in_progress) {
+            log("A restart request is already in progress or queued");
+            return;
+        }
+        _in_progress = true;
```
lostintangent: If there isn't a pending update, and an app calls this twice specifying `true`, for the `onlyIfUpdateIsPending` parameter, then this wouldn't be reset to `false` since the app wouldn't actually restart. I don't think that's the end of the world, but since we're logging messages when a restart is "blocked" due to one being in progress, I'd prefer to address this edge case.

scottbommarito:
updated this again
@geof90 @lostintangent

RestartManager.js
```diff
@@ -26,6 +28,12 @@ const RestartManager = (() => {
 
     function restartApp(onlyIfUpdateIsPending = false) {
         if (_allowed) {
+            if (_inProgress) {
+                log("A restart request is already in progress or queued");
+                return;
+            }
+            // The restart won't execute if `onlyIfUpdateIsPending === true` and there is no pending update.
+            _inProgress = !onlyIfUpdateIsPending || !!(NativeCodePush.getUpdateMetadata(CodePush.UpdateState.PENDING));
```
lostintangent: Isn't this duplicating logic between this JS method and its corresponding native implementation? It might be better to have the native `restartApp` method return a `Promise` (or something similar) which allows the JS-side to determine whether the restart actually happened, at which point, you can reset the `_inProgress` variable to `false`.

scottbommarito: I've been trying all day to get it work like that to, unfortunately, no avail. The general issue is that receiving information from the plugin requires using a promise. Because the promise is not synchronous, the `_inProgress` field takes a non-negligible amount of time to update, and calling `restartApp` twice in a row causes it to crash still because `_inProgress` has not been updated yet. In retrospect, I think this line is incorrect as well because it always sets `_inProgress` to true because `NativeCodePush.getUpdateMetadata` returns a promise (which passes the `!!` check) and not an actual boolean. I wrote a new test case so we now have coverage of a) restarts in progress block other restarts and b) restarts that are ignored do not block additional restarts but unfortunately this seems to be a harder problem than i had thought initially.

lostintangent: Couldn't you set `__inProgress` to `true` and then set it to `false` when a `Promise` returned by `NativeCodePush.restartApp` resolves? That way, you set the flag to `true` immediately, which blocks subsequent calls to `restartApp` in the same flush cycle of the JS->native call queue. If the app actually restarts, then it doesn't matter, but at least you can guarantee that calls to `restartApp` can succeed after the native call which doesn't result in a restart has completed, and returned back to JS.

scottbommarito: yea, talked with @geof90 about it offline, seems like the sync in progress approach is the best way to go

scottbommarito:
@geof90 @lostintangent 
how about now?

RestartManager.js
```diff
@@ -3,41 +3,54 @@ const NativeCodePush = require("react-native").NativeModules.CodePush;
 
 const RestartManager = (() => {
     let _allowed = true;
-    let _restartPending = false;
+    let _restartInProgress = false;
+    let _restartQueue = [];
```
silhouettes: For my own education - is there a reason we'd want to restart multiple times back-to-back, as opposed to squashing them into a single restart?

RestartManager.js
```diff
@@ -3,41 +3,54 @@ const NativeCodePush = require("react-native").NativeModules.CodePush;
...
         log("Disallowing restarts");
         _allowed = false;
     }
 
-    function restartApp(onlyIfUpdateIsPending = false) {
-        if (_allowed) {
-            NativeCodePush.restartApp(onlyIfUpdateIsPending);
-            log("Restarting app");
-        } else {
+    async function restartApp(onlyIfUpdateIsPending = false) {
```
silhouettes: (For my education) In what scenario does the value of `onlyIfUpdateIsPending` matter (assuming that we check whether an update is pending before we call restartApp)? If I understand correctly, you'd have to call `restartApp()`, but by the time the restart happens, the update is no longer pending?

RestartManager.js
geof90: When you say "squash" them, I assume you mean drop the second one if the first one is in progress.

The problem is with `onlyIfUpdateIsPending`. Lets say we have this scenario:

- The first time, `onlyIfUpdateIsPending` is `true`.
- The second time, `onlyIfUpdateIsPending` is `false`. 
- There is no update pending.

If we allowed them to run in parallel, the second one would succeed. But if we dropped the second one because the first one is in progress, the customer would see the restart not actually happen despite setting `onlyIfUpdateIsPending` to `false`, which is not the right behaviour based on our documentation.

RestartManager.js
geof90: If the customer wants to use this API to restart the app regardless of whether an update is pending or not, he would call this with `false`. If he only wants to restart if there is an actual update downloaded, he would call this with `true`.

silhouettes: Oh! Didn't realize this was exposed directly to the customer. So it's a convenience thing, gotcha!

silhouettes:
Looks good!

RestartManager.js
silhouettes: Okay, so this makes sense. Basically, we process these requests one by one, with the idea that we may eventually find one where a restart actually happens. If a restart happens at any point, the context gets destroyed and the queue gets wiped (so the restart would only happen once).

Would it make sense to rename these two variables to `_processingRestartRequest` and `_restartRequests`, to make it clear that they're not necessarily restarts?

RestartManager.js
```diff
@@ -3,41 +3,54 @@ const NativeCodePush = require("react-native").NativeModules.CodePush;
...
+            log("Restart request queued until the current restart is completed");
+            _restartQueue.push(onlyIfUpdateIsPending);
+        } else if (!_allowed) {
             log("Restart request queued until restarts are re-allowed");
-            _restartPending = true;
-            return true;
+            _restartQueue.push(onlyIfUpdateIsPending);
+        } else {
+            _restartInProgress = true;
+            log("Restarting app");
```
lostintangent: Should we move this log within the if statement below it, that way we only actuallynlog that the app is being restarted if in fact it is?

silhouettes: Does the interior of that if block actually get executed if the app is restarted?

geof90: Honestly, I'm not sure. I will test it, but I'm guessing the answer is no, so in fact, the entire `if` switch might not be needed now that we are awaiting.

lostintangent:
LGTM!

RestartManager.js
geof90: So I was wrong, it does indeed get executed, which is kinda weird, but i guess it means that this implementation is ok as-is.

PR #397, opened 2016-06-23
Bug fix for #379 where status report can be undefined on notifyApplicationReady

nevir:
I introduced a JS error on subsequent launches after an update is applied, where `getNewStatusReport()` returns nothing.  This fixes it!

### DIFF
--- a/CodePush.js
+++ b/CodePush.js
@@ -173,7 +173,7 @@ const notifyApplicationReady = (() => {
 async function notifyApplicationReadyInternal() {
   await NativeCodePush.notifyApplicationReady();
   const statusReport = await NativeCodePush.getNewStatusReport();
-  tryReportStatus(statusReport); // Don't wait for this to complete.
+  statusReport && tryReportStatus(statusReport); // Don't wait for this to complete.
 
   return statusReport;
 }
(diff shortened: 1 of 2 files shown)


lostintangent:
Good catch! This LGTM, thanks for the contribution :)

CodePush.js
```diff
@@ -173,7 +173,7 @@ const notifyApplicationReady = (() => {
 async function notifyApplicationReadyInternal() {
   await NativeCodePush.notifyApplicationReady();
   const statusReport = await NativeCodePush.getNewStatusReport();
-  tryReportStatus(statusReport); // Don't wait for this to complete.
+  statusReport && tryReportStatus(statusReport); // Don't wait for this to complete.
```
scottbommarito: do you think it might be better to put this check inside `tryReportStatus` instead?

scottbommarito:
small question, `// LGTM` regardless

CodePush.js
nevir: It seems to complicate it a fair bit, primarily due to [the retry logic](https://github.com/convoyinc/react-native-code-push/blob/2b628007e0298909e0063e5906cb890cd6ae40aa/CodePush.js#L209-L220)

Making that code behave the same, without checking the status report up front, is hard to follow; and it seems weird to only treat it as optional from `notifyApplicationReadyInternal` but not when recursing

lostintangent: I agree with the current implementation. Thanks again for the fix dude!

PR #398, opened 2016-06-24
0.29 compat + refactoring android

geof90:
This PR:

- Adds compatibility to breaking changes in the RN 0.29 API, where config will be done in the ReactApplication class instead of ReactActivity.

- Refactors some of the Android code - Moves settings access code to a SettingsManager class, moves CodePushNativeModule into its own file, and moves all our constants declaration into a CodePushConstants class.


android/app/build.gradle
```diff
@@ -22,5 +22,5 @@ android {
 dependencies {
     compile fileTree(dir: "libs", include: ["*.jar"])
     compile "com.android.support:appcompat-v7:23.0.1"
-    compile "com.facebook.react:react-native:0.19.+"
+    compile 'com.facebook.react:react-native:+'
```
shishirx34: nit: are single quotes on purpose?

README.md
```diff
@@ -255,8 +255,47 @@ And that's it for installation using RNPM! Continue below to the [Plugin Configu
...
+public class MainApplication extends Application implements ReactApplication {
+
+    private final ReactNativeHost mReactNativeHost = new ReactNativeHost(this) {
+        ...
+        // 2. Override the getJSBundleFile method in order to let
+        // the CodePush runtime determine where to get the JS
+        // bundle location from on each app start
+        @Override
+        protected String getJSBundleFile() {
+            return CodePush.getBundleUrl();
```
lostintangent: Should we finally alias this method to be `CodePush.getJSBundleFile`? That way its named just like the "outer" method? This is what we did for UWP. It's a small change, but I feel like it cleans up the API a bit, and we're already making a ton of changes for RN 0.29 anyways.

android/app/build.gradle
```diff
@@ -22,5 +22,5 @@ android {
 dependencies {
     compile fileTree(dir: "libs", include: ["*.jar"])
     compile "com.android.support:appcompat-v7:23.0.1"
-    compile "com.facebook.react:react-native:0.19.+"
+    compile "com.facebook.react:react-native:+"
```
lostintangent: Why this change? I don't believe our plugin would support 0.18, so having a min version still seems to make sense.

geof90: I think "0.19.+" means "minor version 19, any patch version". Before I made this change, I was getting compiler errors.

android/app/src/main/java/com/microsoft/codepush/react/CodePush.java
```diff
@@ -110,28 +79,46 @@ public CodePush(String deploymentKey, Activity mainActivity, boolean isDebugMode
         initializeUpdateAfterRestart();
     }
 
-    // USED FOR TESTING SO THAT IT CAN CONNECT TO DEBUG SERVER
-    public CodePush(String deploymentKey, Activity mainActivity, boolean isDebugMode, String serverUrl) {
```
scottbommarito: I don't see a new version of this function anywhere, it's necessary for tests.

geof90: I've added it back.

Examples/CodePushDemoApp/android/app/src/main/java/com/microsoft/codepushdemoapp/MainActivity.java
```diff
@@ -12,7 +12,7 @@
 
     @Override
     protected String getJSBundleFile() {
-        return CodePush.getBundleUrl("index.android.bundle");
+        return CodePush.getJSBundleFile("index.android.bundle");
```
lostintangent: We can probably simplify the demo app by removing the explicit bundle name parameter.

android/app/src/main/java/com/microsoft/codepush/react/CodePush.java
```diff
@@ -1,51 +1,31 @@
...
-import java.util.HashMap;
 import java.util.List;
-import java.util.Map;
 import java.util.zip.ZipEntry;
 import java.util.zip.ZipFile;
 
 public class CodePush implements ReactPackage {
+
     private static boolean needToReportRollback = false;
     private static boolean isRunningBinaryVersion = false;
```
lostintangent: Sort these?

android/app/src/main/java/com/microsoft/codepush/react/CodePush.java
```diff
@@ -147,20 +139,36 @@ private long getBinaryResourcesModifiedTime() {
...
     public static String getBundleUrl() {
-        return getBundleUrl(DEFAULT_JS_BUNDLE_NAME);
+        return getJSBundleFile();
     }
 
     public static String getBundleUrl(String assetsBundleFileName) {
+        return getJSBundleFile(assetsBundleFileName);
+    }
+
+    public String getBundleUrlInternal(String assetsBundleFileName) {
```
lostintangent: Do we need this one method anymore?

android/app/src/main/java/com/microsoft/codepush/react/CodePush.java
```diff
@@ -147,20 +139,36 @@ private long getBinaryResourcesModifiedTime() {
     }
 
     public static String getBundleUrl() {
```
lostintangent: Should we mark these as being @Deprecated? That way the code and usage is clearer that we expected you to just the newer methods?

android/app/src/main/java/com/microsoft/codepush/react/CodePushConstants.java
```diff
@@ -0,0 +1,28 @@
+package com.microsoft.codepush.react;
+
+public class CodePushConstants {
+    static final String ASSETS_BUNDLE_PREFIX = "assets://";
```
lostintangent: Should we explicitly mark all of these are public? I believe we chose to be explicit in the Windows implementation, and RN Java code seems to do the same.

android/app/src/main/java/com/microsoft/codepush/react/CodePushDialog.java
```diff
@@ -11,30 +11,23 @@
 
 public class CodePushDialog extends ReactContextBaseJavaModule{
 
-    private Activity mainActivity;
+    private CodePushNativeModule codePushNativeModule;
```
lostintangent: Should we start using the "m" prefix for private class members? That seems to be the equivalent of the _ in the Java world, and its what RN uses.

android/app/src/main/java/com/microsoft/codepush/react/SettingsManager.java
```diff
@@ -0,0 +1,131 @@
...
+public class SettingsManager {
+
+    private Context applicationContext;
+
+    public SettingsManager(Context applicationContext) {
+        this.applicationContext = applicationContext;
+    }
+
+    JSONArray getFailedUpdates() {
+        SharedPreferences settings = applicationContext.getSharedPreferences(CodePushConstants.CODE_PUSH_PREFERENCES, 0);
```
lostintangent: Should we initialize this object in the constructor? That way each method didn't need to construct it themselves?

android/app/src/main/java/com/microsoft/codepush/react/CodePushNativeModule.java
```diff
@@ -0,0 +1,482 @@
...
+        Activity currentActivity = getCurrentActivity();
+
+        if (!ReactActivity.class.isInstance(currentActivity)) {
+            // Our preferred reload logic relies on the user's Activity inheriting
+            // from the core ReactActivity class, so if it doesn't, we fallback
+            // early to our legacy behavior.
+            loadBundleLegacy();
+        } else {
+            try {
+                ReactActivity reactActivity = (ReactActivity)getCurrentActivity();
```
lostintangent: Can't you cast "currentActivity" instead of needing to retrieve the current activity again?

android/app/src/main/java/com/microsoft/codepush/react/CodePushPackage.java
```diff
@@ -18,18 +18,6 @@
 import java.nio.ByteBuffer;
 
 public class CodePushPackage {
```
lostintangent: In the UWP implementation, I believe we renamed this to "UpdateManager". I'd be in favor of changing this to CodePushUpdateManager as well.

android/app/src/main/java/com/microsoft/codepush/react/CodePushConstants.java
geof90: Leaving out the access modifier makes it "package private", which I thought more closely resembles the behaviour in the Windows implementation.

android/app/src/main/java/com/microsoft/codepush/react/CodePush.java
```diff
@@ -146,119 +126,111 @@ private long getBinaryResourcesModifiedTime() {
         }
     }
 
+    public int getBuildVersion() {
```
lostintangent: Do we actually use this anywhere? I wonder whether we can't remove it.

android/app/src/main/java/com/microsoft/codepush/react/CodePush.java
```diff
@@ -1,137 +1,117 @@
...
+    private boolean mDidUpdate = false;
+
+    private String mAssetsBundleFileName;
 
     // Helper classes.
-    private CodePushNativeModule codePushNativeModule;
-    private CodePushPackage codePushPackage;
-    private CodePushTelemetryManager codePushTelemetryManager;
+    private CodePushNativeModule mCodePushNativeModule;
+    private CodePushUpdateManager mCodePushUpdateManager;
```
lostintangent: Do we need the "CodePush" prefix in these? "mNativeModule", etc. might be sufficient in the context of the CodePush class.

android/app/src/main/java/com/microsoft/codepush/react/CodePushDialog.java
```diff
@@ -11,30 +11,23 @@
...
         super(reactContext);
-        this.mainActivity = mainActivity;
+        mCodePushNativeModule = codePushNativeModule;
     }
 
     @ReactMethod
     public void showDialog(String title, String message, String button1Text, String button2Text,
                       final Callback successCallback, Callback errorCallback) {
-        AlertDialog.Builder builder = new AlertDialog.Builder(mainActivity);
+        AlertDialog.Builder builder = new AlertDialog.Builder(mCodePushNativeModule.getCurrentActivityPublic());
```
lostintangent: Since this is a native module as well, can't you just call getCurrentActivity to get the Activity?

android/app/src/main/java/com/microsoft/codepush/react/CodePushNativeModule.java
```diff
@@ -0,0 +1,482 @@
...
+            // binaryHash will be null if the React Native assets were not bundled into the APK
+            // (e.g. in Debug builds)
+            configMap.putString(CodePushConstants.PACKAGE_HASH_KEY, binaryHash);
+        }
+
+        promise.resolve(configMap);
+    }
+
+    public Activity getCurrentActivityPublic() {
+        return getCurrentActivity();
```
lostintangent: Per my comment below, do we actually need this?

android/app/src/main/java/com/microsoft/codepush/react/CodePush.java
```diff
@@ -1,137 +1,117 @@
...
-    private static final String DEFAULT_JS_BUNDLE_NAME = "index.android.bundle";
-    private final String DOWNLOAD_PROGRESS_EVENT_NAME = "CodePushDownloadProgress";
-    private final String FAILED_UPDATES_KEY = "CODE_PUSH_FAILED_UPDATES";
-    private final String PACKAGE_HASH_KEY = "packageHash";
-    private final String PENDING_UPDATE_HASH_KEY = "hash";
-    private final String PENDING_UPDATE_IS_LOADING_KEY = "isLoading";
-    private final String PENDING_UPDATE_KEY = "CODE_PUSH_PENDING_UPDATE";
-    private final String RESOURCES_BUNDLE = "resources.arsc";
+
+    private static boolean mIsRunningBinaryVersion = false;
```
lostintangent: The convention for non-final static members seems to be to prefix them with "s" instead of "m".

lostintangent:
LGTM! Thanks for making this change, it cleans things up a lot.

PR #400, opened 2016-06-24
Binary hash (iOS)

lostintangent:
This fixes the binary hashing feature for apps using React Native 0.29.0+ by accounting for the new `*.meta` file which is generated as part of the bundle. I verified that it works perfectly with this fix, and the fix that the RN team made to 0.29.0-RC.


geof90:
LGTM!

PR #404, opened 2016-06-27
Android binary hashing when there are no asset files

geof90:
The script used to generate the binary hash would no-op in the case of a project that has no assets at all (images, videos, sound etc). Unfortunately, this was a bug that existed since the time this feature was first released...


lostintangent:
LGTM!

PR #405, opened 2016-06-28
Prepend iOS native logs with "\n[CodePush] "

geof90:
This PR creates a `CPLog` C function which is used to prepend all the logs from native with `\n[CodePush] `. The line break is needed so that the `code-push debug` command picks them up, plus it looks nicer in the debug console anyway:

Without line break:
```
2016-06-27 17:44:11.708 CodePushDemoApp[10313:75249] [CodePush] Loading JS bundle from file:///***/main.jsbundle
```

With line break:
```
2016-06-27 17:44:11.708 CodePushDemoApp[10313:75249]
[CodePush] Loading JS bundle from file:///***/main.jsbundle
```

The `CPLog` function is implemented as a C function for convenience and API similarity with `NSLog`.


lostintangent:
LGTM!

Issue #378, opened 2016-06-15
Can get stuck in an infinite update loop


nevir:
(advanced-mode only)

If you call `LocalPackage#install()`  before calling `codePush.notifyAppReady()`, your application will restart forever, even if the package being loaded is the one that was asked to install.

It appears a bit difficult to detect, at least in Obj-C land, as the update isn't marked with "IsLoading" until some time later (next full launch).  Should that flag be set whenever reloading the bundle, too?

---

rough repro steps are just to have a bundle that does:

```
const remote = await codePush.checkForUpdate();
const pending = await remote.download();
await pending.install(codePush.InstallMode.IMMEDIATELY);
```

nevir:
So, this is actually a bit of an issue for us:

* We'd like to be able to check for updates _before_ the app fully loads
  * Intention: ensure that first launch after downloading a new native app has the most recent bundle (transparently to the user)
  * We'll also time it out after a few seconds and just update in the background (in case of bad connectivity)
* But at the same time, we don't want to mark the app as ready before that (since we're not likely to encounter errors on boot until after checking for updates on boot)

lostintangent:
We can look into this ASAP to make sure you're unblocked. Just to make sure we can repro this, can you share where in your app you're calling the code snippet you previously posted, and where you're calling notifyApplicationReady? Or does it matter?

nevir:
Thanks!  Really appreciate how response you guys are

Doesn't seem to matter (since the currently-applying package stays pending until we call `notifyAppReady`) - but we do call `notifyAppReady` asynchronously later (if there is no update to install), often 250ms+ later

Here's a slightly cleaned up copy of what we've got atm:

```
function handleAppLaunch() {
  return new Promise((resolve, reject) => {
    let completed = false;

    setTimeout(() => {
      if (completed) return;
      completed = true;
      resolve();
    }, 5000);

    codePush.sync({installMode: codePush.InstallMode.ON_NEXT_RESUME})
      .then(() => {
        // We didn't make it in time; nothing more to do.
        if (completed) return;
        completed = true; // Don't allow the timer to fire.
        const pendingPackage = await codePush.getUpdateMetadata(codePush.UpdateState.PENDING);
        pendingPackage.install(codePush.InstallMode.IMMEDIATE).then(resolve, reject);
      })
      .catch(error => {
        if (completed) {
          console.error(`handleAppLaunch failed after timeout:`, error);
        } else {
          completed = true;
          reject(error);
        }
[...]

Issue #376, opened 2016-06-15
[obj-c] nil object error


nevir:
I haven't been able to reproduce this one, but I just got a crash on a launch after a failed update (in the iOS simulator, code push 1.12.2, RN 0.27.1):

```
Jun 14 18:18:03 NevBook-2.local Driver[84012]: Update did not finish loading the last time, rolling back to a previous version.
Jun 14 18:18:03 NevBook-2.local Driver[84012]: *** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '*** -[__NSArrayM insertObject:atIndex:]: object cannot be nil'
	*** First throw call stack:
	(
		0   CoreFoundation                      0x00d3f946 __exceptionPreprocess + 182
		1   libobjc.A.dylib                     0x009caa97 objc_exception_throw + 44
		2   CoreFoundation                      0x00bf6c01 -[__NSArrayM insertObject:atIndex:] + 881
		3   CoreFoundation                      0x00bf6861 -[__NSArrayM addObject:] + 65
		4   CodePush                            0x0012469b -[CodePush saveFailedUpdate:] + 283
		5   CodePush                            0x001244fb -[CodePush rollbackPackage] + 139
		6   CodePush                            0x00123a55 -[CodePush initializeUpdateAfterRestart] + 341
		7   CodePush                            0x001238cd -[CodePush init] + 125
		8   libobjc.A.dylib                     0x009e097f +[NSObject new] + 98
		9   React                               0x003698c1 -[RCTModuleData setUpInstanceAndBridge] + 993
		10  React                               0x0036adec __25-[RCTModuleData instance]_block_invoke + 44
		11  React                               0x003fe7e4 RCTExecuteOnMainThread + 100
		12  React                               0x0036aba1 -[RCTModuleData instance] + 657
		13  React                               0x002f455b __49-[RCTBatchedBridge initModulesWithDispatchGroup:]_block_invoke_2 + 139
		14  libdispatch.dylib                   0x023aca2a _dispatch_call_block_and_release + 15
		15  libdispatch.dylib                   0x023c903f _dispatch_client_callout + 14
		16  libdispatch.dylib                   0x023b2764 _dispatch_main_queue_callback_4CF + 470
		17  CoreFoundation                      0x00c9995e __CFRUNLOOP_IS_SERVICING_THE_MAIN_DISPATCH_QUEUE__ + 14
		18  CoreFoundation                      0x00c58760 __CFRunLoopRun + 2256
[...]

PR #407, opened 2016-06-29
No-op in CodePushPackage.m / CodePushUpdateManager.java if the pending update is already installed

geof90:
This PR fixes https://github.com/Microsoft/react-native-code-push/issues/378. When the same update is installed twice, we would previously delete the current package's contents erroneously.

### DIFF
--- a/android/app/src/main/java/com/microsoft/codepush/react/CodePushUpdateManager.java
+++ b/android/app/src/main/java/com/microsoft/codepush/react/CodePushUpdateManager.java
@@ -275,6 +275,13 @@ public class CodePushUpdateManager {
     public void installPackage(ReadableMap updatePackage, boolean removePendingUpdate) {
         String packageHash = CodePushUtils.tryGetString(updatePackage, CodePushConstants.PACKAGE_HASH_KEY);
         WritableMap info = getCurrentPackageInfo();
+
+        String currentPackageHash = CodePushUtils.tryGetString(info, CodePushConstants.CURRENT_PACKAGE_KEY);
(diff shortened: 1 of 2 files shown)


lostintangent:
Could this be handled entirely in JS? That way the UWP implementation would benefit as well? If we added a new "isInstalled" property, that was set to false for remote packages, and true for local packages (similarly to how isPending works currently), then presumably we could simply check if a package was installed and no-op a call to install, without ever needing to transition to the native side.

This change looks great to me, I just wanted to throw the thought out there :)

geof90:
The reason why I didnt go with that is that we still have to add the resume listener / restart the app depending on the installMode of the second call to install. Since all of that logic is currently implemented in native, i wanted to reuse that instead of reimplementing everything in JS.

lostintangent:
Ah OK. I guess there's a bit of a design question of whether it would be better to completely no-op subsequent installs of a package, as opposed to short-circuiting some of the behavior. I had originally thought we were doing the former, based on the fact that I dint think we wanted to prescribe the use of the install method in order to "overwrite" the install mode of a pending update. The restartApp method would presumably solve this scenario, and if we found that people wanted to attach a resume listener explicitly, there may be another solution for that.

Either way, this change looks good, I just had the wrong expectations of the behavior.

lostintangent:
BTW, could this possibly fix #376 as well? The root issue there was that the current package was missing when performing a rollback, so I wonder whether the double `install` bug deleted the package at the right time, and things got screwed up.

PR #410, opened 2016-07-01
Improve legacy restart logic

lostintangent:
This PR improves our legacy restart logic in two ways:

1. It re-uses the same `Activity` instance that was captured in the `loadBundle` method, which prevents it from failing victim to the `getCurrentActivity` method returning `null` by the time it is called. It's rare, but possible, so in general, it seems to make sense to capture a strong reference to the `Activity` once, and then re-use that throughout the lifecycle of the reload operation.

2. It uses the `Activity.recreate` method, instead of "manually" finishing the current `Activity` and then attempting to restart is via an `Intent`. This is the "official" way of doing this post-Honeycomb, and since we don't need to support Android versions older than that (or even as old as that!), we mine as well use it. Additionally, this safeguards us from potential issues where `Activity.getIntent()` returns `null` due to the `Activity` having been destroyed, and we are therefore not actually able to restart the `Activity` again.

As a side-effect, the user experience is actually much better, and the `Activity` doesn't "slide out" and then come back. The `recreate` method performs a "flash" update very similarly to what our reflection-based logic achieves, so in general, this seems to be much better then the existing solution for scenarios that we need to fall back to it.

### DIFF
--- a/android/app/src/main/java/com/microsoft/codepush/react/CodePushNativeModule.java
+++ b/android/app/src/main/java/com/microsoft/codepush/react/CodePushNativeModule.java
@@ -79,13 +79,13 @@ public class CodePushNativeModule extends ReactContextBaseJavaModule {
         return false;
     }
 
-    private void loadBundleLegacy(final Activity activity) {
+    private void loadBundleLegacy(final Activity currentActivity) {
         mCodePush.invalidateCurrentInstance();
 
-        activity.runOnUiThread(new Runnable() {
+        currentActivity.runOnUiThread(new Runnable() {
             @Override
             public void run() {
-                activity.recreate();
+                currentActivity.recreate();
             }
         });
     }
--- a/android/app/src/main/java/com/microsoft/codepush/react/CodePushNativeModule.java
+++ b/android/app/src/main/java/com/microsoft/codepush/react/CodePushNativeModule.java
@@ -79,24 +79,26 @@ public class CodePushNativeModule extends ReactContextBaseJavaModule {
         return false;
     }
 
-    private void loadBundleLegacy() {
-        Activity currentActivity = getCurrentActivity();
-        Intent intent = currentActivity.getIntent();
-        currentActivity.finish();
-        currentActivity.startActivity(intent);
-
+    private void loadBundleLegacy(final Activity activity) {
         mCodePush.invalidateCurrentInstance();
+
+        activity.runOnUiThread(new Runnable() {
+            @Override
+            public void run() {
+                activity.recreate();
+            }
+        });
     }
 
     private void loadBundle() {
         mCodePush.clearDebugCacheIfNeeded();
-        Activity currentActivity = getCurrentActivity();
+        final Activity currentActivity = getCurrentActivity();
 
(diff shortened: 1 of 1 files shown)


android/app/src/main/java/com/microsoft/codepush/react/CodePushNativeModule.java
```diff
@@ -79,24 +79,26 @@ private boolean isReactApplication(Context context) {
...
-    private void loadBundleLegacy() {
-        Activity currentActivity = getCurrentActivity();
-        Intent intent = currentActivity.getIntent();
-        currentActivity.finish();
-        currentActivity.startActivity(intent);
-
+    private void loadBundleLegacy(final Activity activity) {
         mCodePush.invalidateCurrentInstance();
+
+        activity.runOnUiThread(new Runnable() {
```
geof90: Should we also null check the activity?

android/app/src/main/java/com/microsoft/codepush/react/CodePushNativeModule.java
```diff
@@ -79,24 +79,26 @@ private boolean isReactApplication(Context context) {
...
-    private void loadBundleLegacy() {
-        Activity currentActivity = getCurrentActivity();
-        Intent intent = currentActivity.getIntent();
-        currentActivity.finish();
-        currentActivity.startActivity(intent);
-
+    private void loadBundleLegacy(final Activity currentActivity) {
         mCodePush.invalidateCurrentInstance();
+
+        currentActivity.runOnUiThread(new Runnable() {
```
geof90: Do you think we should also null check the activity?

geof90:
LGTM!

android/app/src/main/java/com/microsoft/codepush/react/CodePushNativeModule.java
lostintangent: I don't think it will actually help safe-guard us here, since the `loadBundle` method doesn't check against `null`, and therefore, would crash far before we hit this part, and the code which recreates the RN context couldn't `null` out the `Activity` reference we have, or force it to be GC'd, since we'd still be keeping it alive. The issue here was that we allowed RN to control the lifetime of the `Activity` reference, by delegating to `getCurrentActivity`, as opposed to capturing our own reference for the span of time that we need it.

geof90: Almost all of loadBundle() is inside a `try ...catch` which would catch NPEs arising from a null currentActivity, and then fall back to this legacy behaviour. I'm just wondering whether there is a chance that at the top of the `loadBundle` method where we try to get the current activity, it is already null, and we would be handing over this `null` instance to `loadBundleLegacy`. I thought maybe we should do this check just because most other RN native modules also do the same check. But that said, there is no evidence we have ever seen this happen in practice, so I am fine with you merging this as is.

lostintangent: Yeah I think doing this check might makes sense, but it probably should be within `loadBundle`, as do the error check up front before propagating it further.

PR #411, opened 2016-07-01
Remove dependency on react-native-button from sample app

geof90:
Use the in-built TouchableOpacity instead. @silhouettes


silhouettes:
Looks good!

PR #413, opened 2016-07-05
Clean up gradle file

lostintangent:
This PR primarily ensures that any lint errors within our plugin don't break the build of a hosting app. We had a user-reported issue related to this over the weekend, and other React Native plugins disable this (e.g. react-native-vector-icons), so it seems like a good change to make in general.

While making this change, I also cleaned up some of the stuff that isn't necessary as a first step towards "converting" our Android implementation into a library project, instead of the app structure that it currently has (e.g. we can delete our `settings.gradle` and local grade copy, etc.).

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -1,4 +1,4 @@
-apply plugin: 'com.android.library'
+apply plugin: "com.android.library"
 
 android {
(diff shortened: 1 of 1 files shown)


geof90:
LGTM!

PR #415, opened 2016-07-06
[Android] Removing native module reference from CodePush class

lostintangent:
This PR simply removes the unnecessary `CodePushNativeModule` reference that the `CodePush` class was holding on to, but not actually using for anything (after the recent refactoring). Since the `CodePush` class instance persists across app restarts, it shouldn't be holding on to instances of the `CodePushNativeModule` object, which wouldn't survive a restart, and therefore, could be getting accidentally leaked.


geof90:
LGTM!

Issue #414, opened 2016-07-06
[Android] Crash in getConfiguration


lostintangent:
The following crash report was provided by a user, and was hit in an app running React Native 0.28.0 on Android, and using `v1.12.1` of the CodePush plugin. It seems like for some reason, `getCurrentActivity` is returning `null` in a call to `getConfiguration`, which I have no clue when/why that would happen.

```
Fatal Exception: java.lang.NullPointerException: Attempt to invoke virtual method 'android.content.ContentResolver android.app.Activity.getContentResolver()' on a null object reference
       at com.microsoft.codepush.react.CodePushNativeModule.getConfiguration(CodePushNativeModule.java:249)
```

This is tied to #408 in the sense that it is caused by our change to use `getCurrentActivity` within our native module, however, this exception is hit within the call to `getConfiguration`, as opposed to `restartApp`. That kind of tells me that we may need to re-examine how we're using `getCurrentActivity` to determine whether there are other issues we might run into.

At the very least, maybe we should safeguard apps from potential NPE by checking if `getCurrentActivity` returns `null`, and if so, simply short-circuit the extra features it enables: binary hashing and user identification for install metrics. Omitting both would lead to other issues, but at least they wouldn't cause an app crash.

lostintangent:
Fixed via #416.

PR #416, opened 2016-07-06
[Android] Guarding against NPEs in getConfiguration

lostintangent:
This PR addresses the tactical element of #414 by safeguarding the `getConfiguration` native module method against NPEs. It's still unclear when and why `getCurrentActivity` could return `null` (while the app is still in the foreground), but in general, we should strive to prevent any and all crashes, and we know from customer reports that `getCurrentActivity` can sometimes return `null`, so we should enable "progressive enhancement" in our plugin whenever we can access the `Activity`, and not fail completely when we can't.

Since the device ID is really important (i.e. it enables install metrics and rollout), and cheap to compute, I moved it to be captured as part of the native module being created, at which point, we shouldn't ever have an issue. However, since generating the binary hash is a little more expensive, I left that to happen lazily as part of `getConfiguration`, and therefore, if `getCurrentActivity` returns `null`, the worst case scenario is that binary hashing will stop working.

As pointed out by @geof90, React Native core modules also do `null` checks when calling `getCurrentActivity`, so in general, this seems like a reasonable change to make.

### DIFF
--- a/android/app/src/main/java/com/microsoft/codepush/react/CodePushNativeModule.java
+++ b/android/app/src/main/java/com/microsoft/codepush/react/CodePushNativeModule.java
@@ -32,6 +32,7 @@ import java.util.HashMap;
 import java.util.Map;
 
 public class CodePushNativeModule extends ReactContextBaseJavaModule {
+    private String mBinaryContentsHash = null;
     private String mClientUniqueId = null;    
     private LifecycleEventListener mLifecycleEventListener = null;
     private int mMinimumBackgroundDuration = 0;
@@ -52,6 +53,8 @@ public class CodePushNativeModule extends ReactContextBaseJavaModule {
         mTelemetryManager = codePushTelemetryManager;
         mUpdateManager = codePushUpdateManager;
 
+        // Initialize module state while we have a reference to the current context.
+        mBinaryContentsHash = CodePushUpdateUtils.getHashForBinaryContents(reactContext, mCodePush.isDebugMode());
         mClientUniqueId = Settings.Secure.getString(reactContext.getContentResolver(), Settings.Secure.ANDROID_ID);
     }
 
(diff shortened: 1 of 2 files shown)


android/app/src/main/java/com/microsoft/codepush/react/CodePushNativeModule.java
```diff
@@ -240,19 +245,20 @@ public void dispatchDownloadProgressEvent() {
...
-                        android.provider.Settings.Secure.ANDROID_ID));
-        String binaryHash = CodePushUpdateUtils.getHashForBinaryContents(currentActivity, mCodePush.isDebugMode());
-        if (binaryHash != null) {
-            // binaryHash will be null if the React Native assets were not bundled into the APK
-            // (e.g. in Debug builds)
-            configMap.putString(CodePushConstants.PACKAGE_HASH_KEY, binaryHash);
+
+        Activity currentActivity = getCurrentActivity();
+        if (currentActivity != null) {
+            String binaryHash = CodePushUpdateUtils.getHashForBinaryContents(currentActivity, mCodePush.isDebugMode());
```
geof90: I think we should update `getHashForBinaryContents` to use a Context instead of an Activity. Really, all its doing is looking in the assets folder to retrieve the hash which we computed during compile time. So it doesn't need an Activity, we could use the `reactContext`.

In addition, since the binary hash is precomputed during compile time and is rather cheap to just retrieve it, maybe it would be okay to retrieve it during app start like what you are doing for the clientUniqueId. But I can see your apprehension to adding additional steps on app start, so maybe this is less important.

lostintangent: OK cool, I'll move this to app start and update the until method to take a Context instead of an Activity. Then our only dependency on getCurrentActivity will be the restart logic, which I don't see a way around unfortunately.

geof90:
LGTM!

PR #418, opened 2016-07-06
fixed DeviceEventEmitter deprecation from react native 0.27

dorongutman:
see https://github.com/facebook/react-native/releases/tag/v0.27.0

### DIFF
--- a/package-mixins.js
+++ b/package-mixins.js
@@ -1,5 +1,5 @@
 import { AcquisitionManager as Sdk } from "code-push/script/acquisition-sdk";
-import { DeviceEventEmitter } from "react-native";
+import { Keyboard } from "react-native";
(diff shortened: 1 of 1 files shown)


geof90:
Hi @dorongutman, the `CodePushDownloadProgress` event is not a keyboard event. Are you sure this would work? And even if it does, it looks strange that we would be registering custom events like this on the Keyboard.

Is there any official instruction from RN to plugin authors to no longer use the `DeviceEventEmitter`? I definitely still see the `DeviceEventEmitter` working in RN 0.29.

dorongutman:
Sorry, this is NOT true - the deprecation is only for keyboard related events.closing.

geof90:
Cool, thanks for being on the ball anyway!

PR #421, opened 2016-07-09
Fix 'unchecked call' warnings on classes obtained through reflection

silhouettes:
The types of these classes cannot be imported, presumably because they're inaccessible before RN 0.29 (correct me if I'm wrong!), so explicitly turn off type checking.


geof90:
LGTM!

Issue #423, opened 2016-07-11
[Android 1.13.2-beta]: NPE on loadBundle


overminder:
### Context

- `react-native@0.29.0`
- `react-native-code-push@1.13.2-beta`
- `Android 4.4 and 5.1, armv7a` (Shouldn't matter..)
- `CodePush` created with `isDebugMode` off.

### Stacktrace

```
java.lang.NullPointerException
at CodePushNativeModule.loadBundleLegacy(CodePushNativeModule.java:86)
at CodePushNativeModule.loadBundle(CodePushNativeModule.java:103)
at CodePushNativeModule.restartApp(CodePushNativeModule.java:460)
...
```

[Or here on master.](https://github.com/Microsoft/react-native-code-push/blob/b665afdeb5e83fb390d9ae5bfc10507c1f6d50e5/android/app/src/main/java/com/microsoft/codepush/react/CodePushNativeModule.java#L93)

### Reproduction

0. Ensure there is a new CodePush release in the cloud.
1. Launch the React Native based app.
2. (In the JS side) Call `CodePush.sync({ installMode: IMMEDIATE })`.
3. (In the JS side, at the same time) Launch a native activity (In our case it is a Google AdMob Interstitial).
4. Keep the native activity open. Wait until the CodePush bundle is downloaded. Once the installation finishes (and `CodePushNativeModule#restartApp` is called), the app crashes.

### Analysis

The faulty null pointer came from `ReactContextBaseJavaModule#getCurrentActivity`, which returns `ReactContext#mCurrentActivity#get`. This field will be set to null in `ReactContext#onHostPause`, which in this issue will be triggered by launching another native activity.

It's quite possible that this can't be solved without changes to the react-native repo.. Maybe we can propose to add something like `Promise<Activity> getCurrentActivityEventually()` so that we can restart the activity when it comes back.

geof90:
@overminder are you using https://www.npmjs.com/package/react-native-admob?

overminder:
@geof90 

Thanks for your fix! That's pretty quick..

Yes, I'm using react-native-admob. But as you might have already figured out, any kind of native activity should be able to trigger this.

geof90:
@overminder cool! Could you do us a favour and pull the latest bits from GitHub now that I have merged it, and verify that the fix works?

geof90:
The fix was relatively quick because you provided a ton of useful information.

overminder:
[...]

PR #424, opened 2016-07-11
Add resume listener for immediate installs

geof90:
This fixes https://github.com/Microsoft/react-native-code-push/issues/423.

- No-op `loadBundle()` if `getCurrentActivity()` returns `null`, preventing a `NullPointerException`.
- Add the resume listener even for `InstallMode = IMMEDIATE` installs.

**There can be two situations when `getCurrentActivity` returns `null`:** 

1. The current activity gets backgrounded (e.g. by starting a new native Activity), so [React Native sets it to `null` in the `onHostPause` listener](https://github.com/facebook/react-native/blob/e3b594839262ffecb4e90a075d46df52c4d33350/ReactAndroid/src/main/java/com/facebook/react/bridge/ReactContext.java#L181)

2. The current activity is killed (either by the OS, or explicitly).

**Why I think this is the right fix:**

- If 1. happens, when the app goes back to the current activity after it was backgrounded, the resume listener will kick in and automatically restart the app, picking up the new update. This works because [when the activity comes back to the foreground, React Native resets the `currentActivity`](https://github.com/facebook/react-native/blob/e3b594839262ffecb4e90a075d46df52c4d33350/ReactAndroid/src/main/java/com/facebook/react/bridge/ReactContext.java#L159).

- If 2. happens, we really don't have to do anything beyond letting `loadBundle` no-op, because when the activity next gets started "naturally", React Native's initialization will happen and it will load the latest JS bundle for us for free.

- If neither happens, `restartApp` gets called immediately after the update is downloaded, which is the happy path :) The resume listener that we set would no longer take effect because the React context is torn down and restarted.

### DIFF
--- a/android/app/src/main/java/com/microsoft/codepush/react/CodePushNativeModule.java
+++ b/android/app/src/main/java/com/microsoft/codepush/react/CodePushNativeModule.java
@@ -388,7 +388,7 @@ public class CodePushNativeModule extends ReactContextBaseJavaModule {
                     // if the current activity is backgrounded, we want to reload the bundle when
                     // it comes back into the foreground.
                     installMode == CodePushInstallMode.IMMEDIATE.getValue()) {
-                    
+
                     // Store the minimum duration on the native module as an instance
                     // variable instead of relying on a closure below, so that any
                     // subsequent resume-based installs could override it.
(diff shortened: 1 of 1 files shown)


silhouettes:
Aside from the backgroundDuration comment, LGTM!

Issue #427, opened 2016-07-13
CodePush does not support a bundle resource not at the root of the main bundle


joshuafeldman:
On iOS code push currently only supports `bundleURLForResource:withExtension:` which prevents the ability to contain all of the react resources inside of a parent folder at the root of the main bundle. `NSBundle` supports `URLForResource:withExtension:subdirectory:` which would allow us to specify the directory in the bundle the react resources live in and the directory code push should download to. If I have some time I will try and issue a pull request but it would be great to know if there is a plan to add this and if it was added will it be supported and function correctly. Thanks!

joshuafeldman:
Pull request https://github.com/Microsoft/react-native-code-push/pull/428

geof90:
Closed via #428

joshuafeldman:
@geof90 after using it for a while it looks like this doesn't actually work and is causing a rollback on second launch. Looks like some of the other code push classes make assumptions about the directory structure. In `CodePushPackage.m` there is a method called `-getBinaryAssetsPath` which is assuming that `assets` exists at the root of the bundle. I believe this assumption is breaking this. I advise we fix these assumptions (not sure if there are more yet) or rollback the change in your master branch since it will break for people. This ticket should be reopened.

joshuafeldman:
I see this in the console
```
[info][tid:com.facebook.react.JavaScript] [CodePush] Checking for update.
[info][tid:com.facebook.react.JavaScript] [CodePush] An unknown error occurred.
[info][tid:com.facebook.react.JavaScript] [CodePush] g.getCurrentPackage is not a function. (In 'g.getCurrentPackage()', 'g.getCurrentPackage' is undefined)
```

PR #428, opened 2016-07-13
Issue #427 - Adding subdirectory support for iOS bundle

joshuafeldman:

PR #429, opened 2016-07-14
Prepend subdirectory to assets path

geof90:
Fixes https://github.com/Microsoft/react-native-code-push/issues/427

https://github.com/Microsoft/react-native-code-push/pull/428 added support for obtaining the URL to the binary packaged JS bundle within a subdirectory. However, if anyone used that feature, it would prevents diff updates from being installed correctly because we did not also prepend the subdirectory to the path to the assets folder.

@joshuafeldman

### DIFF
--- a/ios/CodePush/CodePush.h
+++ b/ios/CodePush/CodePush.h
@@ -27,6 +27,8 @@
 
 + (NSString *)getApplicationSupportDirectory;
 
++ (NSString *)bundleAssetsPath;
+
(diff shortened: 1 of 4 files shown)


joshuafeldman:
@geof90 I updated to this branch and I can confirm is does fix the rollback after first launch. Any chance we can get this merged in and the version updated so we can go back to using the master branch?

Thanks!!!

PR #431, opened 2016-07-16
Add @codePush decorator

geof90:
This PR allows you wrap your app's root component with the @codePush decorator! For example:

```javascript
import codePush from 'react-native-code-push';

@codePush({
  checkFrequency: codePush.CheckFrequency.ON_APP_RESUME,
  updateDialog: true,
  installMode: codePush.InstallMode.IMMEDIATE
})
class GetCurrentActivityCrash extends Component {
  codePushStatusDidChange(status) {
    console.log("Received CodePush status: " + status);
  }

  codePushDownloadDidProgress(progress) {
    console.log(progress.receivedBytes  + " of " + progress.totalBytes + " received.");
  }

  render() {
    ...
  }
}
```

The benefit is that you can now declaratively specify the behaviour of how/when you want CodePush to check for updates, whether to show a dialog, when to apply the updates, and our plugin will ensure that we inject the calls appropriately for you. This may help to prevent intermittent rollbacks as a result of misconfiguring our API, such as forgetting to call `codePush.notifyAppReady()` immediately after the update is applied.

Going forward, this also lets us configure adding more capabilities (like only calling `sync` when connected to WiFi, etc).

I have also updated the README docs for this feature.

### DIFF
--- a/CodePush.js
+++ b/CodePush.js
@@ -412,13 +412,14 @@ function codePushify(options = {}) {
       if (!React) {
         throw new Error("Unable to find the 'React' module.");
       }
-      if (!React.Component) {
-        throw new Error(
-`Unable to find the 'Component' class, please either:
+    }
+
+    if (!React.Component) {
+      throw new Error(
+`Unable to find the "Component" class, please either:
 1. Upgrade to a newer version of React Native that supports it, or
-2. Call the codePush.sync API in your component instead of using the @CodePushify decorator`
-        );
-      }
+2. Call the codePush.sync API in your component instead of using the @codePush decorator`
+      );
     }
 
     return class CodePushComponent extends React.Component {
--- a/CodePush.js
+++ b/CodePush.js
@@ -430,7 +430,7 @@ function codePushify(options = {}) {
           let rootComponentInstance = this.refs.rootComponent;
           let syncStatusCallback = rootComponentInstance && rootComponentInstance.codePushStatusDidChange;
           let downloadProgressCallback = rootComponentInstance && rootComponentInstance.codePushDownloadDidProgress;
-          CodePush.sync(options, syncStatusCallback, downloadProgressCallback)
+          CodePush.sync(options, syncStatusCallback, downloadProgressCallback);
           if (options.checkFrequency === CodePush.CheckFrequency.ON_APP_RESUME) {
             ReactNative.AppState.addEventListener("change", (newState) => {
               newState === "active" && CodePush.sync(options, syncStatusCallback, downloadProgressCallback);
--- a/CodePush.js
+++ b/CodePush.js
@@ -410,6 +410,9 @@ function codePushify(options = {}) {
     if (!React) {
       try { React = ReactNative.React; } catch (e) { }
       if (!React) {
+        throw new Error("Unable to find the 'React' module.");
+      }
+      if (!React.Component) {
         throw new Error(
 `Unable to find the 'Component' class, please either:
 1. Upgrade to a newer version of React Native that supports it, or
@@ -423,8 +426,7 @@ function codePushify(options = {}) {
         let rootComponentInstance = this.refs.rootComponent;
         let syncStatusCallback = rootComponentInstance && rootComponentInstance.codePushStatusDidChange;
         let downloadProgressCallback = rootComponentInstance && rootComponentInstance.codePushDownloadDidProgress;
-
-        CodePush.sync(options, syncStatusCallback, downloadProgressCallback);
+        CodePush.sync(options, syncStatusCallback, downloadProgressCallback)
         if (options.syncMode === CodePush.SyncMode.ON_APP_RESUME) {
           ReactNative.AppState.addEventListener("change", (newState) => {
             newState === "active" && CodePush.sync(options, syncStatusCallback, downloadProgressCallback);
--- a/CodePush.js
+++ b/CodePush.js
@@ -455,7 +455,6 @@ if (NativeCodePush) {
   Object.assign(CodePush, {
     AcquisitionSdk: Sdk,
     checkForUpdate,
-    codePushify,
     getConfiguration,
     getCurrentPackage,
     getUpdateMetadata,
--- a/CodePush.js
+++ b/CodePush.js
@@ -423,14 +423,18 @@ function codePushify(options = {}) {
 
     return class CodePushComponent extends React.Component {
       componentDidMount() {
-        let rootComponentInstance = this.refs.rootComponent;
-        let syncStatusCallback = rootComponentInstance && rootComponentInstance.codePushStatusDidChange;
-        let downloadProgressCallback = rootComponentInstance && rootComponentInstance.codePushDownloadDidProgress;
-        CodePush.sync(options, syncStatusCallback, downloadProgressCallback)
-        if (options.syncMode === CodePush.SyncMode.ON_APP_RESUME) {
-          ReactNative.AppState.addEventListener("change", (newState) => {
-            newState === "active" && CodePush.sync(options, syncStatusCallback, downloadProgressCallback);
-          });
+        if (options.checkFrequency === CodePush.CheckFrequency.MANUAL) {
+          CodePush.notifyAppReady();
+        } else {
+          let rootComponentInstance = this.refs.rootComponent;
+          let syncStatusCallback = rootComponentInstance && rootComponentInstance.codePushStatusDidChange;
+          let downloadProgressCallback = rootComponentInstance && rootComponentInstance.codePushDownloadDidProgress;
+          CodePush.sync(options, syncStatusCallback, downloadProgressCallback)
+          if (options.checkFrequency === CodePush.CheckFrequency.ON_APP_RESUME) {
+            ReactNative.AppState.addEventListener("change", (newState) => {
+              newState === "active" && CodePush.sync(options, syncStatusCallback, downloadProgressCallback);
+            });
+          }
         }
       }
 
@@ -446,7 +450,8 @@ function codePushify(options = {}) {
 // and therefore, it doesn't make sense initializing
 // the JS interface when it wouldn't work anyways.
 if (NativeCodePush) {
-  CodePush = {
+  CodePush = codePushify;
+  Object.assign(CodePush, {
     AcquisitionSdk: Sdk,
     checkForUpdate,
     codePushify,
@@ -477,9 +482,10 @@ if (NativeCodePush) {
       DOWNLOADING_PACKAGE: 7,
       INSTALLING_UPDATE: 8
     },
-    SyncMode: {
+    CheckFrequency: {
       ON_APP_START: 0,
       ON_APP_RESUME: 1,
+      MANUAL: 2
     },
     UpdateState: {
       RUNNING: NativeCodePush.codePushUpdateStateRunning,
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -404,33 +404,87 @@ With the CodePush plugin downloaded and linked, and your app asking CodePush whe
...
+let codePushOptions;
 
-    ```javascript
-    codePush.sync();
+@codePushify(codePushOptions)
+class MyApp extends Component {
+}
+```
+
+*NOTE: Decorators are not yet supported in Babel 6.x pending proposal update.* You may need to enable it by doing the following:
```
silhouettes: Are decorators not supported by default? If so, then perhaps it would be worth just making option 2 the default? We could maybe then also just remove this option and leave it as a footnote to option 2, only for those people who already know how to use decorators and have their project configured for it.

README.md
```diff
@@ -717,6 +773,105 @@ codePush.checkForUpdate()
...
+    }
+    ```
+
+##### CodePushOptions
+
+The `codePushify` decorator accepts an "options" object that allows you to customize numerous aspects of the default behavior mentioned above:
+
+* __syncMode__ *(codePush.SyncMode)* - Specifies when you would like to check for updates. Defaults to `codePush.SyncMode.ON_APP_START`. Refer to the [`SyncMode`](#syncmode) enum reference for a description of the available options and what they do.
+
+* __deploymentKey__ *(String)* - Refer to [`SyncOptions`](#syncoptions).
```
silhouettes: If we're intending for the decorator to be the standard prescription, maybe it would be better to describe the options here, and link to this from syncOptions rather than the other way around

README.md
```diff
@@ -404,33 +404,87 @@ With the CodePush plugin downloaded and linked, and your app asking CodePush whe
...
+The simplest way to do this is to "CodePush-ify" your app's root component. To do so, you can choose one of the following three options:
 
-1. Import the JavaScript module for CodePush:
+**Option 1: Use the [ES7 decorator](https://github.com/wycats/javascript-decorators) syntax:**
 
-    ```javascript
-    import codePush from "react-native-code-push";
-    ```
+```javascript
+import codePush, { codePushify } from "react-native-code-push";
```
silhouettes: Just a thought - would it be worth just making codePushify the default export?

README.md
```diff
@@ -717,6 +773,105 @@ codePush.checkForUpdate()
...
+
+```javascript
+codePush.codePushify(options: CodePushOptions)(rootComponent: React.Component): React.Component;
+```
+
+Used as a decorator to wrap a React component inside a "higher order" React component that knows how to synchronize your app's JavaScript bundle and image assets with the latest release to the configured deployment when it is mounted. Internally, the wrapper component calls [`sync`](#codepushsync) inside its `componentDidMount` lifecycle handle, which in turns performs an update check, downloads the update if it exists and installs the update for you.
+
+This decorator provides support for letting you customize its behaviour to easily enable apps with different requirements. Below are some examples of ways you can use it (you can pick one or even use a combination):
+
+1. **Silent sync on app start** *(the simplest, default behavior)*. Your app will automatically download available updates, and apply them the next time the app restarts (e.g. the OS or end user killed it, or the device was restarted). This way, the entire update experience is "silent" to the end user, since they don't see any update prompt and/or "synthetic" app restarts.
```
silhouettes: Seeing as we now have the option of re-defining the default behavior without breaking people, would it be worth considering making this default now to be on resume, after a specific time interval?

README.md
```diff
@@ -404,33 +404,87 @@ With the CodePush plugin downloaded and linked, and your app asking CodePush whe
...
+let codePushOptions;
+
+class MyApp extends Component {
+}
+
+MyApp = codePushify(codePushOptions)(MyApp);
+```
 
-If you would like your app to discover updates more quickly, you can also choose to call `sync` every time the app resumes from the background, by adding the following code (or something equivalent) as part of your app's startup behavior (e.g. your root component's `componentDidMount` method). You can call `sync` as frequently as you would like, so when and where you call it just depends on your personal preference.
+**Option 3: Manually call the `sync` method from within the `componentDidMount` lifecycle event, to initiate a background update on each app start:**
```
silhouettes: This is probably something we can discuss - we want everyone new to use the decorator going forward right? Instead of muddying the waters by documenting `sync()` here, would it be worth just adding a link to the 'legacy' version of the README at the top of this one, for people who were already using it?

Then we can just treat `sync()` as an advanced API that you can find if you scroll down and look at the API reference, similarly to `checkForUpdate()` etc.

CodePush.js
```diff
@@ -436,6 +477,10 @@ if (NativeCodePush) {
       DOWNLOADING_PACKAGE: 7,
       INSTALLING_UPDATE: 8
     },
+    SyncMode: {
```
silhouettes: Would `SyncFrequency` be a more descriptive name here?

CodePush.js
```diff
@@ -436,6 +477,10 @@ if (NativeCodePush) {
...
+      ON_APP_START: 0,
+      ON_APP_RESUME: 1,
```
silhouettes: This is a random thought, I'd be interested in hearing what you think. Would it be worth adding a `MANUAL` option here? This would basically just result in `notifyApplicationReady()` being injected into `componentDidMount()` and nothing else (for now). Our docs could just document the `MANUAL` option as an alternative way of using CodePush, where you call `sync()` wherever you need it (doesn't matter if it's on app start or not). Then `sync()` is not technically an alternative to `codePushify()`, it's just something you can use in `MANUAL` mode.

For example, if someone wants to do a codePush update on button press (who knows why), then they wouldn't need to know about `notifyApplicationReady()`, they can just call `sync()` on the button press and it works as they expect.

CodePush.js
```diff
@@ -401,6 +401,46 @@ async function syncInternal(options = {}, syncStatusChangeCallback, downloadProg
...
+    let React;
+    let ReactNative = require("react-native");
+
+    try { React = require("react"); } catch (e) { }
+    if (!React) {
+      try { React = ReactNative.React; } catch (e) { }
+      if (!React) {
+        throw new Error("Unable to find the 'React' module.");
+      }
+      if (!React.Component) {
```
silhouettes: Should this `if` statement be outside of the `if (!React)` block, so that we can check it even if `React` was successfully found the first time?

Also, I'm not sure we need (2) - do we expect this to happen often?

CodePush.js
```diff
@@ -401,6 +401,46 @@ async function syncInternal(options = {}, syncStatusChangeCallback, downloadProg
...
+        );
+      }
+    }
+
+    return class CodePushComponent extends React.Component {
+      componentDidMount() {
+        let rootComponentInstance = this.refs.rootComponent;
+        let syncStatusCallback = rootComponentInstance && rootComponentInstance.codePushStatusDidChange;
+        let downloadProgressCallback = rootComponentInstance && rootComponentInstance.codePushDownloadDidProgress;
+        CodePush.sync(options, syncStatusCallback, downloadProgressCallback)
```
silhouettes: ;

silhouettes:
I love this, the implementation is very clean! Just left a bunch of comments/ideas on documentation and API shape, probably of varying degrees of quality, that would be cool to hear your thoughts on.

README.md
geof90: Do you have a way to do it that wouldn't break backcompat?

CodePush.js
geof90: Yep, I like this suggestion. 👍

README.md
geof90: They aren't.. I'm ok with making Option 2 default, although it is kinda weird looking.

README.md
geof90: I guess if we go with the `MANUAL` mode you suggested below, we can replace this with the manual instructions (which would look similar)?

README.md
geof90: 👍

README.md
geof90: While I see the value of making resume based syncing a default, I feel that keeping the mapping similar to the `sync` API may help users more easily migrate to this new API (and is just cleaner)?

CodePush.js
geof90: How about `SyncTrigger`? The understanding is we are triggering `sync` during an event, `ON_APP_START`, `ON_APP_RESUME` or `MANUAL` (when the user calls `sync` manually).

README.md
```diff
@@ -406,85 +406,84 @@ With the CodePush plugin downloaded and linked, and your app asking CodePush whe
 
 The simplest way to do this is to "CodePush-ify" your app's root component. To do so, you can choose one of the following three options:
 
-**Option 1: Use the [ES7 decorator](https://github.com/wycats/javascript-decorators) syntax:**
+**Option 1: Call `codePush` on your root component and assign the result to itself after the component declaration:**
```
silhouettes: Could we simplify the phrasing here to something like: `Wrap your root component`?

README.md
```diff
@@ -406,85 +406,84 @@ With the CodePush plugin downloaded and linked, and your app asking CodePush whe
...
 
 class MyApp extends Component {
 }
 
-MyApp = codePushify(codePushOptions)(MyApp);
+MyApp = codePush(CodePushOptions)(MyApp);
 ```
 
-**Option 3: Manually call the `sync` method from within the `componentDidMount` lifecycle event, to initiate a background update on each app start:**
+Alternatively, if you don't want CodePush to trigger checking for updates automatically for you, but want to trigger it manually in app code in response to some event (e.g. a button press), you can specify a `MANUAL` check frequency and do the following:
```
silhouettes: `and call CodePush.sync() as desired:` ?

silhouettes: Is `sync()` going to inherit options from the root decorator?

typings/react-native-code-push.d.ts
```diff
@@ -242,13 +249,6 @@ declare namespace CodePush {
     function checkForUpdate(deploymentKey?: string): Promise<RemotePackage>;
 
     /**
-     * Decorates a React Component configuring it to sync for updates with the CodePush server.
-     *
-     * @param options Options used to configure the end-user sync and update experience (e.g. when to check for updates?, show an prompt?, install the update immediately?).
-     */
-    function codePushify(options?: CodePushOptions): Function;
-
-    /**
```
silhouettes: `codePushify` is still a property on CodePush in the implementation, we should probably make it consistent one way or another

README.md
```diff
@@ -404,33 +404,89 @@ With the CodePush plugin downloaded and linked, and your app asking CodePush whe
 
 2. When an update is available, how to present it to the end user?
 
-The simplest way to do this is to perform the following in your app's root component:
+The simplest way to do this is to "CodePush-ify" your app's root component. To do so, you can choose one of the following three options:
```
silhouettes: two options

README.md
```diff
@@ -404,33 +404,89 @@ With the CodePush plugin downloaded and linked, and your app asking CodePush whe
...
+
+let codePushOptions;
+
+@codePush(codePushOptions)
+class MyApp extends Component {
+}
+```
+
+*NOTE: Decorators are not yet supported in Babel 6.x pending proposal update.* You may need to enable it by doing the following:
+- Install the [`babel-preset-react-native-stage-0` package](https://github.com/skevy/babel-preset-react-native-stage-0)
```
silhouettes: To avoid cluttering the getting started, could we take these babel instructions and move them to a separate section or README and just link to them here?

Issue #440, opened 2016-07-28
[Android] Crash on sync


rilyu:
I press Home key to sent app to background when codePush.sync running, the app will crash sometimes, error message in logcat:

```
java.lang.NullPointerException
   at android.app.AlertDialog.resolveDialogTheme(AlertDialog.java:143)
   at android.app.AlertDialog$Builder.<init>(AlertDialog.java:360)
   at com.microsoft.codepush.react.CodePushDialog.showDialog(CodePushDialog.java:21)
   at java.lang.reflect.Method.invokeNative(Native Method)
   at java.lang.reflect.Method.invoke(Method.java:515)
   at com.facebook.react.bridge.BaseJavaModule$JavaMethod.invoke(BaseJavaModule.java:319)
   at com.facebook.react.bridge.NativeModuleRegistry$ModuleDefinition.call(NativeModuleRegistry.java:187)
   at com.facebook.react.bridge.NativeModuleRegistry.call(NativeModuleRegistry.java:62)
   at com.facebook.react.bridge.CatalystInstanceImpl$NativeModulesReactCallback.call(CatalystInstanceImpl.java:432)
   at com.facebook.react.bridge.queue.NativeRunnableDeprecated.run(Native Method)
   at android.os.Handler.handleCallback(Handler.java:733)
   at android.os.Handler.dispatchMessage(Handler.java:95)
   at com.facebook.react.bridge.queue.MessageQueueThreadHandler.dispatchMessage(MessageQueueThreadHandler.java:31)
   at android.os.Looper.loop(Looper.java:136)
   at com.facebook.react.bridge.queue.MessageQueueThreadImpl$3.run(MessageQueueThreadImpl.java:193)
   at java.lang.Thread.run(Thread.java:841)
```

source code:
`codePush.sync({installMode: codePush.InstallMode.IMMEDIATE, updateDialog: true});
`

geof90:
This should be caused by `getCurrentActivity()` returning null when the app is backgrounded. Should be a simple fix, I will get this out by today.

PR #441, opened 2016-07-28
Fix NPE caused by showing dialog while app is backgrounded

geof90:
Fixes https://github.com/Microsoft/react-native-code-push/issues/440

### DIFF
--- a/android/app/src/main/java/com/microsoft/codepush/react/CodePushDialog.java
+++ b/android/app/src/main/java/com/microsoft/codepush/react/CodePushDialog.java
@@ -5,6 +5,7 @@ import android.app.AlertDialog;
 import android.content.DialogInterface;
 
 import com.facebook.react.bridge.Callback;
+import com.facebook.react.bridge.LifecycleEventListener;
 import com.facebook.react.bridge.ReactApplicationContext;
 import com.facebook.react.bridge.ReactContextBaseJavaModule;
 import com.facebook.react.bridge.ReactMethod;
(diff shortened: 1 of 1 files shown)


android/app/src/main/java/com/microsoft/codepush/react/CodePushDialog.java
```diff
@@ -16,9 +17,43 @@ public CodePushDialog(ReactApplicationContext reactContext) {
...
     @ReactMethod
-    public void showDialog(String title, String message, String button1Text, String button2Text,
-                      final Callback successCallback, Callback errorCallback) {
-        AlertDialog.Builder builder = new AlertDialog.Builder(getCurrentActivity());
+    public void showDialog(final String title, final String message, final String button1Text,
+                           final String button2Text, final Callback successCallback, Callback errorCallback) {
+        Activity currentActivity = getCurrentActivity();
+        if (currentActivity == null) {
+            // If getCurrentActivity is null, it could be because the app is backgrounded,
+            // so we show the dialog when the app resumes)
```
silhouettes: Is this the only case in which it is null?

android/app/src/main/java/com/microsoft/codepush/react/CodePushDialog.java
```diff
@@ -16,9 +17,43 @@ public CodePushDialog(ReactApplicationContext reactContext) {
...
+            getReactApplicationContext().addLifecycleEventListener(new LifecycleEventListener() {
+                private boolean shown = false;
+
+                @Override
+                public void onHostResume() {
+                    Activity currentActivity = getCurrentActivity();
+                    if (!shown && currentActivity != null) {
+                        shown = true;
```
silhouettes: Is there a way to just remove the event listener here?

silhouettes: I don't think we need the `shown` variable here - this listener will be removed as soon as it's shown

android/app/src/main/java/com/microsoft/codepush/react/CodePushDialog.java
geof90: If the activity is killed by the OS, it could be null too. But in that case, there's nothing much we can do, anyway when the app re-launches, it will call sync again, then we will show the dialog again.

android/app/src/main/java/com/microsoft/codepush/react/CodePushDialog.java
geof90: Yep, implementation artifact. Will remove

silhouettes:
Looks good!

PR #443, opened 2016-07-30
Fix ENOTEMPTY warning

silhouettes:
A customer reported seeing a warning of the format `W System.err: remove failed: ENOTEMPTY (Directory not empty)`. We thought it was caused by line 89 of the original `FileUtils.java`, which may attempt to delete a directory before it is empty. I'm removing it, because line 96 will ensure that it's deleted anyway.

Now that I think about it more though, I'm not sure there was actually any problem with the original code, as `file.delete()` will not raise any kind of exception, it'll simply return false if it did not delete.

As a side note, I've refactored out the `deleteDirectory()` function, as it effectively does the same thing as `deleteFileOrFolderSilently()` (aside from the logging if it fails). There is probably more refactoring I can do here, but I'd like to minimize the scope of the changes.


geof90:
LGTM!

PR #446, opened 2016-08-01
Update README with 0.30 compat

geof90:


silhouettes:
Looks good! Thanks for taking care of this :)

Issue #433, opened 2016-07-18
Allow app version to be configured manually


jkrems:
Since `CodePush` uses the app version to mean "has a native interface compatible with the js bundle", maintaining the  "ABI version" independent from the app store version might be desirable. This is especially important for apps that use react-native in just a small part and wouldn't want to bump their (major) version constantly because of it.

I'd be happy to create a PR for this if that sounds good. The default behavior would be to use the app version if no explicit "ABI version" was provided.

silhouettes:
Hey @jkrems, that's a cool idea, although I just want to understand a little bit more about the use case you have in mind. When you mention needing to bump your major version constantly, are you doing this so that older updates are not applied to newer binaries? Or are you referring generally to the problem of having a hybrid native/React-Native app, where the native part churns a lot independently of the React-Native part of the app?

I'm assuming you know that we can release to a [range of semvers](https://github.com/Microsoft/code-push/tree/master/cli#target-binary-version-parameter), and that doesn't quite address your needs?

jkrems:
The problem comes up when only a small part of an app is developed using react native and the app already has a versioning policy. During the development of one app version there might be multiple (incompatible) iterations of the JavaScript side because of changes in native/JS interfaces. But it might not be feasible to bump the app store version "randomly" because of it when other, non-react native-related things are already depending with that value.

In other words: The app store version may already have an established meaning and it might not be "semver ABI version for react-native JS/native code interactions".

silhouettes:
Gotcha. This is a good idea, and it can possibly be linked to another issue we have with beta distribution where the app store version is not typically updated. Is this urgent or blocking for you? I'd like to discuss it with the team next Tuesday, but can expedite things if needed.

jkrems:
It's not blocking me right now. I mostly wanted to get a feeling if this is something you'd consider adding to the SDK.

silhouettes:
@jkrems - Okay, so let me know if you think this is the cleanest way to do it. What we're currently thinking is just an extra optional argument to the native constructor (which sets the ABI globally). There wouldn't be any option to dynamically change it during the running of your app (although that could be added in future as options to `sync()` and `checkForUpdate()`). Does that make sense to you?

jkrems:
That sounds good & matches what I had in mind when creating the issue! Given that it's supposed to be a version for the native interface, changing it at runtime (at least on iOS, not sure what Android would allow) seems less important.

PR #447, opened 2016-08-02
Allow an app's binary version to be manually specified

silhouettes:
As per requests on Discord and on #433, this change allows the binary version of an app to be manually specified. The static, native `overrideAppVersion()` function should ideally be called once before the bundle is loaded.

Note that it needs to be native, because if it were specified in JS, it would be pushed out over CodePush, which can cause unintended issues. We originally considered adding it as a constructor argument, but CodePush is a static instance in iOS, and in Android we didn't want to overload the constructor too much. We chose to implement it as a function rather than a parameter in a plist or gradle file, to allow for dynamic transformation of the app store version if desired. Any feedback welcome!

@jkrems @joshuafeldman if you're interested!

### DIFF
--- a/android/app/src/main/java/com/microsoft/codepush/react/CodePush.java
+++ b/android/app/src/main/java/com/microsoft/codepush/react/CodePush.java
@@ -29,7 +29,7 @@ public class CodePush implements ReactPackage {
     private static boolean sIsRunningBinaryVersion = false;
     private static boolean sNeedToReportRollback = false;
     private static boolean sTestConfigurationFlag = false;
-    private static String sAppVersionOverride = null;
+    private static String sAppVersion = null;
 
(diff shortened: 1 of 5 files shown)


android/app/src/main/java/com/microsoft/codepush/react/CodePush.java
```diff
@@ -40,7 +41,7 @@
     private SettingsManager mSettingsManager;
 
     // Config properties.
-    private String mAppVersion;
+    private String mPListAppVersion;
```
geof90: There's no PList in android.. maybe mBinaryVersion? or mManifestAppVersion?

geof90:
LGTM!

jkrems:
Thanks!

PR #448, opened 2016-08-02
Update rnpm instructions

geof90:
As of 0.27 of React Native, `rnpm link` is already merged into the React Native CLI. In addition, the RNPM team doesn't seem to be updating the `rnpm` package on NPM anymore. Because of that, as of 0.29 of React Native, our `rnpm link` setup for Android no longer works, because of the introduction of the `ReactNativeHost`. https://github.com/facebook/react-native/pull/8612 addresses this, however the change was made the official React Native CLI and not RNPM. 

This change updates our README to reference the `react-native link` command instead of `rnpm link`, since it seems that the latter is no longer being maintained, and the former is the only command that works for our RN 0.29+ Android users.

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,7 +98,7 @@ In order to accommodate as many developer preferences as possible, the CodePush 
 
 #### Plugin Installation (iOS - RNPM)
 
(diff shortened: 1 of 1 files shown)


silhouettes:
Looks good!

PR #450, opened 2016-08-02
Add postlink scripts

geof90:
This PR adds postlink scripts that RNPM will automatically call when executing `react-native link react-native-code-push` or `rnpm link react-native-code-push` so as to simplify setup.

Once https://github.com/facebook/react-native/pull/9157 goes through, we can also add `postunlink` scripts as well to simplify removal of our plugin.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,9 @@
 # React Native Module for CodePush
 
+*Note: This README is only relevant to the latest version of our plugin in source. If you are using an older version, please switch to the relevant tag to view the docs for that particular version.*
+
+![Switching tags](https://cloud.githubusercontent.com/assets/8598682/17350832/ce0dec40-58de-11e6-9c8c-906bb114c34f.png)
+
 This plugin provides client-side integration for the [CodePush service](http://codepush.tools), allowing you to easily add a dynamic update experience to your React Native app(s).
 
 * [How does it work?](#how-does-it-work)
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -17,7 +17,9 @@
     "url": "https://github.com/Microsoft/react-native-code-push"
   },
   "dependencies": {
-    "code-push": "1.8.0-beta"
+    "code-push": "1.8.0-beta",
+    "inquirer": "^1.1.2",
+    "plist": "^1.2.0"
```
silhouettes: Should we pin these dependencies too?

README.md
```diff
@@ -110,6 +114,8 @@ In order to accommodate as many developer preferences as possible, the CodePush
 
     *Note: If you don't already have RNPM installed, you can do so by simply running `npm i -g rnpm` and then executing the above command. If you already have RNPM installed, make sure you have v1.9.0+ in order to benefit from this one step install.*
 
+2. You will be prompted for the deployment key you'd like to use. If you don't already have it, you can retreive this value by running `code-push deployment ls <appName> -k`, or you can choose to ignore it (by simply hitting `<ENTER>`) and add it in later. To get started, we would recommend just using your `Staging` deployment key, so that you can test out the CodePush end-to-end.
```
silhouettes: retrieve

silhouettes:
Would it be worth doing something to check that the app version is semver and/or allow them to change it?

geof90:
> Would it be worth doing something to check that the app version is semver and/or allow them to change it?

I'm hesitant.. mainly a) I don't see that as part of linking and b) not everyone uses RNPM/react-native link so its not going to help everyone. If anything, I feel like we could display a warning when they use `release-react`.

Issue #453, opened 2016-08-03
Interface CodePush.h method signature +overrideAppVersion: does not match implementation


joshuafeldman:
The interface uses `+ (void)overrideAppVersion:(NSString *)deploymentKey` whereas the implementation is `+ (void)overrideAppVersion:(NSString *)appVersion`.

silhouettes:
Nice catch - I'll tidy that up!

PR #454, opened 2016-08-03
Name the 'appVersion' parameter consistently

silhouettes:
This parameter is misnamed in the interface, addressing it as per #453.


geof90:
LGTM!

PR #461, opened 2016-08-09
Remove SoLoader and fix 0.31 compat

geof90:
[SoLoader](https://github.com/facebook/SoLoader) is a native code library for Android by Facebook that React Native uses. Specifically, a call to `SoLoader.init()` is needed before we can construct instances of `WritableNativeMap`, which is a data structure implemented in native C++ used by React Native libraries send JSON maps to/from JavaScript to native.

[`SoLoader.init()` is called for us](https://github.com/facebook/react-native/blob/3c4fd42749e0079b0cf54e1106bf81c993f9d29a/ReactAndroid/src/main/java/com/facebook/react/XReactInstanceManagerImpl.java#L289) by React Native when the `ReactInstanceManager` is constructed. Hence, normally React Native plugins wouldn't have to call it themselves. Our case is a little special because we require the user to override `getJSBundleFile` in their `MainApplication.java`. `CodePush.getJSBundleFile` calls `getCurrentPackage`, which returns a `WritableNativeMap`. Because in the initialization lifecycle, [the JSBundleLoader is created before the ReactInstanceManager](https://github.com/facebook/react-native/blob/3c4fd42749e0079b0cf54e1106bf81c993f9d29a/ReactAndroid/src/main/java/com/facebook/react/ReactInstanceManager.java#L392), that ends up causing an exception because we are constructing a `WritableNativeMap` before calling `SoLoader.init()`.

Calling `SoLoader.init()` ourselves works perfectly fine for us in mitigating the above issue pre RN 0.31. In RN 0.31, [SoLoader was made an external dependency](https://github.com/facebook/react-native/commit/dd06b74157efa0ebd4980eb79d92b3f0e4be35d6). That has resulted in a `java.lang.NoClassDefFoundError: com.facebook.soloader.SoLoader` reported by some customers during CodePush initialization when our plugin tries to use it directly - somehow, this transitive dependency is not accessible to us.

The fix that this PR is proposing is to remove the need to call `SoLoader.init` entirely: really, the `WritableNativeMap` is only needed to generate an argument object that we can send back to JS. Hence, during initialization, we can simply read and use the current package metadata as a normal `org.json.JSONObject`. We only need convert that to a `WritableNativeMap` when JS is calling into a native method, by which point, the `ReactInstanceManager` would already have been constructed.

In addition, `WritableNativeMap` is an internal data structure which subclasses `WritableMap` that React Native uses, they already provided us a utility [`Arguments.createMap`](https://facebook.github.io/react-native/docs/native-modules-android.html#promises) which is a documented API, so I changed our plugin to use that instead.

### DIFF
--- a/android/app/src/main/java/com/microsoft/codepush/react/CodePush.java
+++ b/android/app/src/main/java/com/microsoft/codepush/react/CodePush.java
@@ -4,10 +4,7 @@ import com.facebook.react.ReactPackage;
 import com.facebook.react.bridge.JavaScriptModule;
 import com.facebook.react.bridge.NativeModule;
 import com.facebook.react.bridge.ReactApplicationContext;
-import com.facebook.react.bridge.ReadableMap;
-import com.facebook.react.bridge.WritableMap;
 import com.facebook.react.uimanager.ViewManager;
-import com.facebook.soloader.SoLoader;
 
 import android.content.Context;
 import android.content.pm.ApplicationInfo;
@@ -54,7 +51,6 @@ public class CodePush implements ReactPackage {
     }
 
     public CodePush(String deploymentKey, Context context, boolean isDebugMode) {
-        SoLoader.init(context, false);
         mContext = context.getApplicationContext();
 
         mUpdateManager = new CodePushUpdateManager(context.getFilesDir().getAbsolutePath());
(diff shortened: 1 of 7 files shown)


silhouettes:
Looks good!

PR #464, opened 2016-08-12
Fix android patch

geof90:
Our auto linking patch for android doesn't work Post RN 0.29, because `this` is an instance of `ReactNativeHost`, not a proper Android `Context`.

And unfortunately, this fix alone doesn't fix linking, we are currently blocked on this too: https://github.com/facebook/react-native/pull/9381

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
   },
   "rnpm": {
     "android": {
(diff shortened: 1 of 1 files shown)


silhouettes:
Looks good!

Issue #465, opened 2016-08-15
[Android] custom ReactInstanceManager


mtunique:
Sometime developers don't inherit `ReactActivity`. When the app(react native app) restart, the activity will recreate.  The experience is very poor.
Maybe we can set the ReactInstanceManager by ourselves.

geof90:
@mtunique Are you linking the plugin via this way: https://facebook.github.io/react-native/docs/integration-with-existing-apps.html#add-native-code?

The main difficulty is your `ReactInstanceManager` would be initialized as such:
```
mReactInstanceManager = ReactInstanceManager.builder()
                ...
                .addPackage(new CodePush(...))
                .build();
```

So, we can't pass in the `mReactInstanceManager` instance to CodePush during construction, perhaps it could be a separate method call?

If you find an elegant way to implement this, feel free to send out a pull request!

geof90:
What if we let you pass in a custom `ReactInstanceManagerHolder` interface, that you would need to implement? The interface would expose a method `ReactInstanceManager getReactInstanceManager() { ... }` to us.

mtunique:
Maybe it is better than a separate method to set the `ReactInstanceManager`.

geof90:
@mtunique I just sent out https://github.com/Microsoft/react-native-code-push/pull/470, which should unblock you, feel free to take a look if you are interested.

geof90:
I published #470 to NPM as v1.14.4-beta so it should work if you pull the latest bits!

igrayson:
We have a related use-case of running our react instance entirely without an activity. I've created a PR, exposing the change we're already using, which happens to take the approach initially discussed here.

https://github.com/Microsoft/react-native-code-push/pull/491

mtunique:
It work in react native 0.29. Thanks.

PR #470, opened 2016-08-17
Remove ReactActivity dependency

geof90:
Add [w=1](https://github.com/Microsoft/react-native-code-push/pull/470/files?w=1) to ignore whitespace changes.

As suggested by https://github.com/Microsoft/react-native-code-push/issues/465, sometimes the activity that hosts the React Native root view does not extend ReactActivity, especially if the app is an existing android app that was integrated with React Native as described in https://facebook.github.io/react-native/docs/integration-with-existing-apps.html (Android section).

This PR modifies our `loadBundle` logic slightly to look for the `mReactInstanceManager` field regardless of whether the Activity is a `ReactActivity`. This should improve the experience for the types of apps mentioned above, assuming that developers will be following the suggested code patches in the Facebook docs exactly down to the naming of fields. Obviously, that assumption is brittle, but at least this change exposes an option to these developers to still be able to leverage on our shiny `loadBundle` experience.

Going forward, we should discuss an API change that allows developers to either pass the `ReactInstanceManager` instance to our CodePush runtime, or perhaps create a custom interface that developers can implement in their activity that exposes a `getReactInstanceManager` method to us.

### DIFF
--- a/android/app/src/main/java/com/microsoft/codepush/react/CodePushNativeModule.java
+++ b/android/app/src/main/java/com/microsoft/codepush/react/CodePushNativeModule.java
@@ -126,7 +126,10 @@ public class CodePushNativeModule extends ReactContextBaseJavaModule {
                 // The React Native version might be older than 0.29, or the activity does not
                 // extend ReactActivity, so we try to get the instance manager via the
                 // "mReactInstanceManager" field.
-                Field instanceManagerField = currentActivity.getClass().getDeclaredField("mReactInstanceManager");
+                Class instanceManagerHolderClass = currentActivity instanceof ReactActivity
+                        ? ReactActivity.class
+                        : currentActivity.getClass();
+                Field instanceManagerField = instanceManagerHolderClass.getDeclaredField("mReactInstanceManager");
                 instanceManagerField.setAccessible(true);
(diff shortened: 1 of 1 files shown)


android/app/src/main/java/com/microsoft/codepush/react/CodePushNativeModule.java
```diff
@@ -105,67 +105,61 @@ private void loadBundle() {
...
+                        // The recreation method threw an unknown exception
+                        // so just simply fallback to restarting the Activity
+                        loadBundleLegacy(currentActivity);
+                    }
+                }
+            });
+        } catch (Exception e) {
+            // Our reflection logic failed somewhere
+            // so fall back to restarting the Activity
+            loadBundleLegacy(currentActivity);
```
silhouettes: So if `mReactInstanceManager` is undefined, we definitely fall back to this?

silhouettes:
Looks good!

android/app/src/main/java/com/microsoft/codepush/react/CodePushNativeModule.java
geof90: Yep.. This catches all exceptions including NullPointerExceptions

Issue #319, opened 2016-04-28
Update contents fail data integrity check


tomauty:
Hi, hopefully this is just something I'm doing incorrectly. Lately I've had to deploy updates twice in a row to get them accepted. The first time I send one out, it always returns that I failed a data integrity check. This is the function I run to generate my release

```
/Users/tom/.nvm/versions/node/v4.2.1/bin/react-native bundle \
--reset-cache=true \
--entry-file=index.ios.js \
--platform=ios --dev=false \
--minify=true \
--bundle-output ./release/main.jsbundle \
--sourcemap-output ./release/main.map \
--assets-dest ./release
```

And then I follow up with

`code-push release MyAppName ./release "*" --mandatory --deploymentName Staging`

What other info might you need to help me out with this? :) Thanks

tomauty:
Note that if I run that top command twice (without changing anything the second time) and do a second deploy, it will work! I would think they would have the same checksum. Not sure what's going on.

lostintangent:
@tomauty Is it the "this release is identical to the existing one" message that you're seeing? If so, that may be due to React Native packager caching issues. 

Have you tried using the new release-react command in the CodePush CLI? It will handle the bundling process for you and will also make sure to clear the packager cache (via "rm -rf $TEMP/react-*") before hand. Using it, you could simply type "code-push release-react MyApp ios -m -t '*'" to achieve the same affect as your two scripts.

tomauty:
Unfortunately no, it's this one:
`The update contents failed the data integrity check.`

I could try `release-react`, but my first command above should reset the packager cache anyway, I would think.

tomauty:
Good to know that I can shorten this build process though!

lostintangent:
@tomauty OK thanks for clarifying. We'll investigate this issue since you definitely shouldn't be seeing it. In the meantime, are you unblocked? It's just a pain to have to release twice? (I just want to make sure I know the severity of the issue :))

tomauty:
Yeah I don't mind releasing twice. If it helps, this also only happens when I'm updating from an existing codepush bundle. If I reinstall, I'll get the newest package without a second push.

lostintangent:
OK cool, thanks for this detail @tomauty. We'll let you know the findings of investigating this issue.

geof90:
[...]

PR #471, opened 2016-08-17
Fix the hash mismatch happening due to .DS_Store [Simulator]

rahuljiresal:
Mac OS creates a `.DS_Store` file in each folder. When running CodePush on the simulator, this creates a hash mismatch. We need to skip this file while calculating the hash of the downloaded bundles.

This likely resolves the issue mentioned at https://github.com/Microsoft/react-native-code-push/issues/319

### DIFF
--- a/ios/CodePush/CodePushUpdateUtils.m
+++ b/ios/CodePush/CodePushUpdateUtils.m
@@ -20,6 +20,14 @@ NSString * const ManifestFolderPrefix = @"CodePush";
     }
     
     for (NSString *fileName in folderFiles) {
+#if TARGET_IPHONE_SIMULATOR
+        // If the user is running this in a simulator, there is a chance
+        // that the .DS_Store file will mess up the calculated hash
+        // We must skip this file.
+        if ([fileName isEqualToString:@".DS_Store"]) {
+            continue;
+        }
+#endif
         NSString *fullFilePath = [folderPath stringByAppendingPathComponent:fileName];
         NSString *relativePath = [pathPrefix stringByAppendingPathComponent:fileName];
         BOOL isDir = NO;
--- a/ios/CodePush/CodePushUpdateUtils.m
+++ b/ios/CodePush/CodePushUpdateUtils.m
@@ -20,14 +20,11 @@ NSString * const ManifestFolderPrefix = @"CodePush";
     }
     
     for (NSString *fileName in folderFiles) {
-#if TARGET_IPHONE_SIMULATOR
-        // If the user is running this in a simulator, there is a chance
-        // that the .DS_Store file will mess up the calculated hash
-        // We must skip this file.
-        if ([fileName isEqualToString:@".DS_Store"]) {
+        // We must skip the macOS generated files.
+        if ([fileName isEqualToString:@".DS_Store"] || [fileName isEqualToString:@"__MACOSX"]) {
             continue;
         }
-#endif
+
         NSString *fullFilePath = [folderPath stringByAppendingPathComponent:fileName];
         NSString *relativePath = [pathPrefix stringByAppendingPathComponent:fileName];
         BOOL isDir = NO;


rahuljiresal:
@silhouettes Could you please take a look at this? Thanks!

ios/CodePush/CodePushUpdateUtils.m
```diff
@@ -20,6 +20,14 @@ + (void)addContentsOfFolderToManifest:(NSString *)folderPath
     }
     
     for (NSString *fileName in folderFiles) {
+#if TARGET_IPHONE_SIMULATOR
```
geof90: Let's remove this #if guard. I think we should skip this file in all cases, because that's what the server does.

geof90:
@rahuljiresal Awesome that you managed to figure this out! I'll let @silhouettes sign off before I merge it.

We should also check if Android has the same problem.

ios/CodePush/CodePushUpdateUtils.m
rahuljiresal: This file is never generated on iOS's file system. Its only Mac OS X that generates this file. The iPhone simulator (being a simulator, and not an emulator) uses OS X's file system, and hence you see this file on the simulator. So I'm guessing checking for the filename would be useless when the app is being built for release. But if you still insist, I can remove it.

rahuljiresal:
@geof90 Android will not have the same problem because the Android emulator never generates these files. The emulator is actually a Virtual Machine running on your computer. Unlike the iPhone simulator, which runs on top of Mac OS (uses its libraries, sometimes with some changes), and "simulates" the behavior of iPhone.

ios/CodePush/CodePushUpdateUtils.m
geof90: The IPA containing the app bundle, which a developer generates and then releases to the store, is still generated by the OSX machine right? From what I understand about this file, it is just used for saving icon display preferences for Finder. I'm not really sure how it got there in the first place even for the simulator, unless a user actually opened up the folder on disk using Finder. Either way, somehow OSX put it there when it generated the app bundle, so I presume its also entirely possible that it would generate it when making a release IPA for the device?

silhouettes: So I believe this error will also happen if the package you release was created on OS X (with accompanying `.DS_Store`), but downloaded on a real mobile device, because the server's hashing algorithm explicitly ignores `.DS_Store` and any `__MACOSX/` directory at the root.

In other words, the problem here is an inconsistency between the hashing algorithm used on the server and the hashing algorithm used on the client. So, I think you can remove this `#ifdef`, and if you like, you can also check for the `__MACOSX` directory, but that's up to you (I'm happy to fix it afterwards otherwise).

geof90:
@rahuljiresal I don't know if problem has anything to do with whether its a simulator or emulator. When CodePush CLI uploads the folder containing these update contents, it would zip up the whole folder, including any .DS_Store files along with it. When the device downloads the whole update contents as and unzips it, then tries to verify the hash, it would still add the .DS_Store file to the manifest used to compute the hash. So whether its a simulator or not, theoretically it would still have the same problem.

The android change doesn't have to be part of your PR. I was just leaving a note.

silhouettes:
Wow! That was fast! Well done on diagnosing the issue, and thanks for the PR. I'm happy to merge it as long as the `#if` guard is removed, and it's up to you if you'd like to make the other changes (I'm happy to make them if you don't want to).

rahuljiresal:
@geof90 @silhouettes Made the changes to add `__MACOSX` to the ignore list.
Thanks for looking into this. 

If possible, can you tag/release a new version on npm so I can start using that?

geof90:
Thanks for your contribution! I'll ping you when it's published to NPM.

geof90:
@rahuljiresal Its been published as v1.14.4-beta!

zhaogao:
great ！  #319  this PR resolve my question ,for "the update contents failed the data integrity check" ,my code-push version is 1.13.2-beta

Issue #473, opened 2016-08-18
binding errors


eladgel:
We define our root class like this

```
var Root = React.createClass({
....
...
...});
module.exports = CodePush(CodePushOptions)(Root);
```

When running the app we get warning messages because of our bindings in the next two calls

`            syncStatusCallback = rootComponentInstance.codePushStatusDidChange.bind(rootComponentInstance);
            downloadProgressCallback = rootComponentInstance.codePushDownloadDidProgress.bind(rootComponentInstance);
`

When moving on to class definition like so
`class Root extends React.Component`
All of the warnings disappear

Is there a way to avoid these warnings other then forking the repo?
Thanks

geof90:
I see. We'll see if we can get this fixed today.

PR #478, opened 2016-08-18
Don't bind lifecycle hooks if root component instance is not ES6 declared

geof90:
Fixes https://github.com/Microsoft/react-native-code-push/issues/473. The issue occurs because component functions in components declared via `React.createClass` are autobound by default, while those declared via the ES6 syntax are not.

### DIFF
--- a/CodePush.js
+++ b/CodePush.js
@@ -430,13 +430,19 @@ function codePushify(options = {}) {
           let rootComponentInstance = this.refs.rootComponent;
 
           let syncStatusCallback;
-          if (rootComponentInstance && rootComponentInstance.codePushStatusDidChange && rootComponentInstance instanceof React.Component) {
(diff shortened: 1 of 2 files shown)


CodePush.js
```diff
@@ -431,12 +431,18 @@ function codePushify(options = {}) {
 
           let syncStatusCallback;
           if (rootComponentInstance && rootComponentInstance.codePushStatusDidChange) {
-            syncStatusCallback = rootComponentInstance.codePushStatusDidChange.bind(rootComponentInstance);
+            syncStatusCallback = rootComponentInstance.codePushStatusDidChange;
+            if (rootComponentInstance instanceof React.Component) {
```
silhouettes: Shouldn't we still bind `this` to the object that was passed into `createClass()`? Or is it just too difficult to get access to that object?

PR #480, opened 2016-08-19
enhanced download package tolerance

ymmuse:
When download a update package, if the request status code is >=400, the method connectionDidFinishLoading will get a 0 bytes file of download.zip in download directory, and the variable isZip is FALSE and download.zip will be renamed app.bundle that occurring by self.doneCallback(isZip) method. The app will crash cause we get a empty file named app.bundle.

### DIFF
--- a/ios/CodePush/CodePushDownloadHandler.m
+++ b/ios/CodePush/CodePushDownloadHandler.m
@@ -48,6 +48,20 @@ failCallback:(void (^)(NSError *err))failCallback {
 }
 
 - (void)connection:(NSURLConnection *)connection didReceiveResponse:(NSURLResponse *)response {
+    if ([response isKindOfClass:[NSHTTPURLResponse class]]) {
+        NSInteger statusCode = [(NSHTTPURLResponse *)response statusCode];
+        if (statusCode >= 400) {
+            [self.outputFileStream close];
+            [connection cancel];
+            
+            NSError *err = [NSError errorWithDomain:@"http download error"
+                                               code:statusCode
+                                           userInfo:nil];
+            self.failCallback(err);
+            return;
+        }
+    }
+    
     self.expectedContentLength = response.expectedContentLength;
     [self.outputFileStream open];
 }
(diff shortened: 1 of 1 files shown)


geof90:
Looks legit. Thanks for your contribution! I will look at this tomorrow.

ios/CodePush/CodePushDownloadHandler.m
```diff
@@ -98,13 +112,24 @@ - (void)connection:(NSURLConnection*)connection didFailWithError:(NSError*)error
...
 - (void)connectionDidFinishLoading:(NSURLConnection *)connection {
+    [self.outputFileStream close];
+    
+    // MUST have received bytes from remote server. do not do this, we will get a 0 bytes file of download.zip in download directory,
+    // the variable isZip is FALSE and download.zip will be renamed app.bundle that occurring by self.doneCallback(isZip) method.
+    if (self.receivedContentLength < 1) {
+        NSError *err = [NSError errorWithDomain:@"received nothing"
+                                           code:-1
+                                       userInfo:nil];
+        self.failCallback(err);
```
geof90: If `didReceiveResponse` gets called above, will `connectionDidFinishLoading` get called again after? Since we already called the `failCallback` once before, we don't want to call it again, because we will end up rejecting the promise to JS twice which will result in an error.

ymmuse: Actually, when got a 400 code, the code
```objective-c
+        if (statusCode >= 400) {
 +            [self.outputFileStream close];
 +            [connection cancel];
 +            
```
 connection will be canceled and didReceiveResponse does not be called.

ymmuse: [Sent when a connection has finished loading successfully.](https://developer.apple.com/library/mac/documentation/Foundation/Reference/NSURLConnectionDataDelegate_protocol/#//apple_ref/occ/intfm/NSURLConnectionDataDelegate/connectionDidFinishLoading:)

I have tested, it did.

geof90:
Thanks @ymmuse!

PR #488, opened 2016-08-25
iOS SDK: Expose isPendingUpdate and isFailedHash publicly

### LINKED ISSUE #487: Make "isPendingUpdate" and "isFailedHash" public for iOS SDK
- These methods are public in the Android SDK, but private in the iOS SDK
- Would be nice if they were both public to maintain consistency
- If the owners are open to PRs, I can make this change.

daniel-beard:
- Brings the iOS SDK up to par with the Android SDK (these methods are exposed through `SettingsManager` on Android).
- Fixes #487

### DIFF
--- a/ios/CodePush/CodePush.h
+++ b/ios/CodePush/CodePush.h
@@ -48,6 +48,19 @@
  */
 + (void)setDeploymentKey:(NSString *)deploymentKey;
 
+/*
+ * This method checks to see whether a specific package hash
+ * has previously failed installation.
+ */
+- (BOOL)isFailedHash:(NSString*)packageHash;
+
+/*
+ * This method checks to see whether a specific package hash
+ * represents a downloaded and installed update, that hasn't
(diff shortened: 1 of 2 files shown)


geof90:
@daniel-beard Can you share how you are using these methods from native code, that they would need to be exposed publicly? Its true that they are exposed in Android but they are also completely undocumented, and it would be great if we have a good reason for supporting them going forward.

daniel-beard:
I'm creating a dev settings screen for QA that shows the current status of CodePush which shows the current status of CodePush (including whether it has a failed or pending update)

geof90:
I see. Is there a reason why you can get this information from JS?

daniel-beard:
I'm calling this from native, so I'd rather avoid the round trip from iOS -> JS -> iOS

geof90:
I'll go ahead and merge this, the implementation looks clean. Thanks @daniel-beard!

PR #491, opened 2016-08-25
[Android] Support React instances with no Activity

igrayson:
This change aims to support applications which run React Native's catalyst instance in the background, sometimes without an `Activity`.

We're using this change today in an RN 0.27 application, which overrides `ReactActivity#createReactInstanceManager` to obtain control of the instance's lifecycle, so that we may run and launch without an `Activity` (i.e., in response to push notifications and other triggers).

We call `CodePush. setReactInstanceManager()` after building our instance (foreground or background).

If we agree with this approach, I'll update this PR with documentation/etc.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -132,6 +132,7 @@ gen/
 .gradle/
 build/
 */build/
+android/app/gradle*
 
 # Local configuration file (sdk path, etc)
 local.properties
--- a/android/app/src/main/java/com/microsoft/codepush/react/CodePush.java
+++ b/android/app/src/main/java/com/microsoft/codepush/react/CodePush.java
@@ -1,5 +1,6 @@
 package com.microsoft.codepush.react;
 
+import com.facebook.react.ReactInstanceManager;
 import com.facebook.react.ReactPackage;
 import com.facebook.react.bridge.JavaScriptModule;
 import com.facebook.react.bridge.NativeModule;
@@ -44,6 +45,7 @@ public class CodePush implements ReactPackage {
     private Context mContext;
     private final boolean mIsDebugMode;
 
+    private static ReactInstanceHolder mReactInstanceHolder;
     private static CodePush mCurrentInstance;
 
     public CodePush(String deploymentKey, Context context) {
@@ -277,6 +279,17 @@ public class CodePush implements ReactPackage {
         mSettingsManager.removeFailedUpdates();
     }
 
+    public static void setReactInstanceHolder(ReactInstanceHolder reactInstanceHolder) {
+        mReactInstanceHolder = reactInstanceHolder;
+    }
+
+    static ReactInstanceManager getReactInstanceManager() {
+        if (mReactInstanceHolder == null) {
+            return null;
+        }
+        return mReactInstanceHolder.getReactInstanceManager();
+    }
+    
     @Override
     public List<NativeModule> createNativeModules(ReactApplicationContext reactApplicationContext) {
         CodePushNativeModule codePushModule = new CodePushNativeModule(reactApplicationContext, this, mUpdateManager, mTelemetryManager, mSettingsManager);
--- a/android/app/src/main/java/com/microsoft/codepush/react/CodePushNativeModule.java
+++ b/android/app/src/main/java/com/microsoft/codepush/react/CodePushNativeModule.java
@@ -1,8 +1,9 @@
 package com.microsoft.codepush.react;
 
 import android.app.Activity;
-import android.content.Context;
 import android.os.AsyncTask;
+import android.os.Handler;
+import android.os.Looper;
 import android.provider.Settings;
 import android.view.Choreographer;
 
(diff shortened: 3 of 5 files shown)


geof90:
Hi @igrayson, overall the approach looks sound! Thanks for your contribution. I have some comments about the implementation, I will post them later.

android/app/src/main/java/com/microsoft/codepush/react/CodePushNativeModule.java
```diff
@@ -99,39 +104,46 @@ public void run() {
...
-                Class instanceManagerHolderClass = currentActivity instanceof ReactActivity
-                        ? ReactActivity.class
-                        : currentActivity.getClass();
-                Field instanceManagerField = instanceManagerHolderClass.getDeclaredField("mReactInstanceManager");
-                instanceManagerField.setAccessible(true);
-                instanceManager = (ReactInstanceManager)instanceManagerField.get(currentActivity);
+            if (instanceManager == null) {
+                final Activity currentActivity = getCurrentActivity();
+
+                if (!ReactActivity.class.isInstance(currentActivity)) {
```
geof90: This is a regression of https://github.com/Microsoft/react-native-code-push/pull/470. Could you take a look at that and fix this? Basically we no longer want to immediately fallback to the legacy logic the moment the activity is not a ReactActivity.

android/app/src/main/java/com/microsoft/codepush/react/CodePush.java
```diff
@@ -277,6 +278,14 @@ public void clearUpdates() {
         mSettingsManager.removeFailedUpdates();
     }
 
+    public static void setReactInstanceManager(ReactInstanceManager reactInstanceManager) {
+        mReactInstanceManager = reactInstanceManager;
+    }
+
+    static ReactInstanceManager getReactInstanceManager() {
```
geof90: private?

android/app/src/main/java/com/microsoft/codepush/react/CodePushNativeModule.java
```diff
@@ -99,39 +104,46 @@ public void run() {
...
+                    // Our preferred reload logic relies on the user's Activity inheriting
+                    // from the core ReactActivity class, so if it doesn't, we fallback
+                    // early to our legacy behavior.
+                    loadBundleLegacy(currentActivity);
+                    return;
+                } else {
+                        ReactActivity reactActivity = (ReactActivity) currentActivity;
```
geof90: Again a regression of #470, we shouldn't have to cast this.

android/app/src/main/java/com/microsoft/codepush/react/CodePushNativeModule.java
```diff
@@ -87,6 +87,11 @@ private boolean isReactApplication(Context context) {
     }
 
     private void loadBundleLegacy(final Activity currentActivity) {
+        if (currentActivity == null) {
```
geof90: Since with your change we always pass `getCurrentActivity()` to this method.. I think it would be cleaner to just have `loadBundleLegacy` not take any method params, but simply call `getCurrentActivity()` here and check for nulls here.

android/app/src/main/java/com/microsoft/codepush/react/CodePush.java
igrayson: It's used by `CodePushNativeModule`

igrayson:
Fixed regressions, caused by basing off of an older `master`.

igrayson:
Updated, updated `README`, and moved to a `InstanceHolder` abstraction.

.gitignore
```diff
@@ -132,6 +132,7 @@ gen/
 .gradle/
 build/
 */build/
+android/app/gradle*
```
geof90: Why was this addition needed?

igrayson: It may not be. The only way I could wrestle android studio into being able
to navigate/compile the project was to allow it to install a gradle wrapper
at `android/app`.

Separately, it would be great if we had developer setup instructions for
Android Studio, in this repo.

geof90: Ah. Normally the way I do development with is to open the demo app project, instead of the library project itself in Android Studio. The demo app project is configured to include the library project as a dependency so those code files would show up.

geof90: Could you revert this change then? Or I can remove it after merging.

android/app/src/main/java/com/microsoft/codepush/react/CodePushNativeModule.java
```diff
@@ -143,27 +114,60 @@ private void loadBundle() {
...
         } catch (Exception e) {
             // Our reflection logic failed somewhere
-            // so fall back to restarting the Activity
-            loadBundleLegacy(currentActivity);
+            // so fall back to restarting the Activity (if it exists)
+            loadBundleLegacy();
+        }
+    }
+    
+    private ReactInstanceManager resolveInstanceManager() throws NoSuchFieldException, IllegalAccessException {
```
geof90: I like this refactoring 👍

geof90:
Thanks @igrayson!

PR #492, opened 2016-08-25
Proguard rules

igrayson:
For us, without this rule, `restartApp` always fell back to using the `currentActivity` (the reflection technique always failed @ https://github.com/Microsoft/react-native-code-push/blob/235450604feca12c902746decfcbb2e715f4a071/android/app/src/main/java/com/microsoft/codepush/react/CodePushNativeModule.java#L145 )

This change was necessary to get codepush working in an app that ran without an activity (ref https://github.com/Microsoft/react-native-code-push/pull/491)

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -14,6 +14,10 @@ android {
     lintOptions {
         abortOnError false
     }
+
+    defaultConfig {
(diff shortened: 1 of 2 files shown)


geof90:
LGTM!

Issue #495, opened 2016-08-26
[Android] react native > 0.30 use XReactInstanceManagerImpl


mtunique:
React native > 0.30 use XReactInstanceManagerImpl. XReactInstanceManagerImpl doesn't have mJSBundleFile.  So this doesn't work.

CodePushNativeModule 140 line.
```java
// #2) Update the locally stored JS bundle file path
Field jsBundleField = instanceManager.getClass().getDeclaredField("mJSBundleFile");
jsBundleField.setAccessible(true);
jsBundleField.set(instanceManager, latestJSBundleFile);
```

PR #496, opened 2016-08-26
Fix 0.32 compat

geof90:
Fixes https://github.com/Microsoft/react-native-code-push/issues/495 as well.

Our reflection based loadBundle is getting far too brittle and sensitive to changes in React Native's android interface. I will try to work with them to get a more stable solution going forward.


silhouettes:
Looks good!

Issue #493, opened 2016-08-26
Crash: -[CodePush saveFailedUpdate:]


avishayil:
http://crashes.to/s/6f41b62b9ee

CodePush version: 1.14.5-beta

Stack trace:
````
Fatal Exception: NSInvalidArgumentException
0  CoreFoundation                 0x18339adb0 __exceptionPreprocess
1  libobjc.A.dylib                0x1829fff80 objc_exception_throw
2  CoreFoundation                 0x18327e708 CFStringConvertNSStringEncodingToEncoding
3  Geektime                       0x1001b51f4 -[CodePush saveFailedUpdate:]
4  Geektime                       0x1001b50d0 -[CodePush rollbackPackage]
5  Geektime                       0x1001b4aec -[CodePush initializeUpdateAfterRestart]
6  Geektime                       0x1001b49f0 -[CodePush init]
7  Geektime                       0x10014ceac -[RCTModuleData setUpInstanceAndBridge]
8  Geektime                       0x100159ed4 RCTExecuteOnMainThread
9  Geektime                       0x10014d578 -[RCTModuleData instance]
10 Geektime                       0x10014787c __49-[RCTBatchedBridge initModulesWithDispatchGroup:]_block_invoke
11 libdispatch.dylib              0x182de54bc _dispatch_call_block_and_release
12 libdispatch.dylib              0x182de547c _dispatch_client_callout
13 libdispatch.dylib              0x182deab84 _dispatch_main_queue_callback_4CF
14 CoreFoundation                 0x183350d50 __CFRUNLOOP_IS_SERVICING_THE_MAIN_DISPATCH_QUEUE__
15 CoreFoundation                 0x18334ebb8 __CFRunLoopRun
16 CoreFoundation                 0x183278c50 CFRunLoopRunSpecific
17 GraphicsServices               0x184b60088 GSEventRunModal
18 UIKit                          0x188566088 UIApplicationMain
19 Geektime                       0x10007797c main (main.m:16)
20 libdispatch.dylib              0x182e168b8 (Missing)
````

geof90:
[...]

PR #497, opened 2016-08-28
Prevent crashes during rollbacks

geof90:
This PR prevents a crash caused by nil object insertion during rollbacks reported in https://github.com/Microsoft/react-native-code-push/issues/493, and also adds extra logging.

### DIFF
--- a/ios/CodePush/CodePush.m
+++ b/ios/CodePush/CodePush.m
@@ -455,9 +455,13 @@ static NSString *bundleResourceSubdirectory = nil;
 {
     NSError *error;
     NSDictionary *failedPackage = [CodePushPackage getCurrentPackage:&error];
-    if (error) {
-        CPLog(@"Error getting current update metadata during rollback: %@", error);
-    } else if (failedPackage) {
+    if (!failedPackage) {
+        if (error) {
+            CPLog(@"Error getting current update metadata during rollback: %@", error);
+        } else {
+            CPLog(@"Attempted to perform a rollback when there is no current update");
+        }
+    } else {
         // Write the current package's metadata to the "failed list"
         [self saveFailedUpdate:failedPackage];
     }
--- a/ios/CodePush/CodePush.m
+++ b/ios/CodePush/CodePush.m
@@ -455,9 +455,12 @@ static NSString *bundleResourceSubdirectory = nil;
 {
     NSError *error;
     NSDictionary *failedPackage = [CodePushPackage getCurrentPackage:&error];
-
-    // Write the current package's metadata to the "failed list"
-    [self saveFailedUpdate:failedPackage];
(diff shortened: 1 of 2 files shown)


ios/CodePush/CodePush.m
```diff
@@ -455,9 +455,12 @@ - (void)rollbackPackage
 {
     NSError *error;
     NSDictionary *failedPackage = [CodePushPackage getCurrentPackage:&error];
-
-    // Write the current package's metadata to the "failed list"
-    [self saveFailedUpdate:failedPackage];
+    if (error) {
```
daniel-beard: This is known as a "fragile" error check. Methods that return a value and accept an error should *always* check the return value from the method first. 

E.g. 

```
NSDictionary *failedPackage = [CodePushPackage getCurrentPackage:&error];
if (!failedPackage) {
    // handle error here.
}
```

geof90: It's already checked below in the `else if`. The current implementation is much better because it does both: prints the error if one exists, and if it doesnt (which signifies a problem with the method's contract) that gets handled too.

daniel-beard: A non-nil error object doesn't always indicate that an error *actually* occurred.

geof90: That example just says to me that an error can occur without the error pointer being set. The example there also shows a `success` flag being returned by the method. I completely agree with that example there that the success flag should be checked.

However, you are saying that an error pointer can be set without an error actually happening, i fail to see why especially considering the method we are calling into is also implemented by us.

geof90: Anyway, I don't see what change you are suggesting either. Are you saying we should remove this error pointer check save the `failedPackage` even if we find that the error pointer is set? Because that doesn't sound safe to me at all. I would rather take the safe route and not write the `failedPackage` if we detect either of the failure conditions.

daniel-beard: It's a code smell to check the error pointer directly, and yes, it does happen occasionally that a method will return a non-nil error object even if the call succeeds. https://www.bignerdranch.com/blog/an-nserror-error/

An alternative here would be this:

```
NSError *error;
NSDictionary *failedPackage = [CodePushPackage getCurrentPackage:&error];

if (!failedPackage) {
    CPLog(@"Error getting current update metadata during rollback: %@", error);
} else {
    // Write the current package's metadata to the "failed list"
    [self saveFailedUpdate:failedPackage];
}
```

geof90: Ok, I've updated the implementation according to your suggestion.

daniel-beard: 👍

silhouettes:
Looks good!

PR #499, opened 2016-08-29
Guard against NSError null pointer dereferences

### LINKED ISSUE #490: [iOS] Methods that take NSError references should guard against null pointer dereference
- Methods in the iOS SDK do not currently guard against null pointer dereferences, they should.
- Example:

```Objective-C
+ (NSMutableDictionary *)getCurrentPackageInfo:(NSError **)error
{
    NSString *statusFilePath = [self getStatusFilePath];
    if (![[NSFileManager defaultManager] fileExistsAtPath:statusFilePath]) {
        return [NSMutableDictionary dictionary];
    }
    
    NSString *content = [NSString stringWithContentsOfFile:statusFilePath
                                                  encoding:NSUTF8StringEncoding
                                                     error:error];
    if (*error) {
        return nil;
    }
    
    NSData *data = [content dataUsingEncoding:NSUTF8StringEncoding];
    NSDictionary* json = [NSJSONSerialization JSONObjectWithData:data
                                                         options:kNilOptions
                                                           error:error];
    if (*error) {
        return nil;
    }
    
    return [json mutableCopy];
}
```

- This method crashes if called like this: `[CodePush getCurrentPackageInfo:nil];`
- This method should use a `localError` object for the internal method calls that pass back an `NSError` instance, then check `if (error)` rather than `if (*error)`

daniel-beard:
- This fixes #490 
- Guards against null dereference of `NSError **` pointers.
- Fixes some of the fragile error checks.
- Future fixes here may require changing some public method signatures, as all methods that take an `NSError **` parameter should at *least* return a `BOOL` indicating success/failure. I haven't done that in this PR, as I wasn't sure how this team handles breaking API changes.
- Conform to best practices outlined here: https://developer.apple.com/library/ios/documentation/Cocoa/Conceptual/ProgrammingWithObjectiveC/ErrorHandling/ErrorHandling.html

### DIFF
--- a/ios/CodePush/CodePushPackage.m
+++ b/ios/CodePush/CodePushPackage.m
@@ -293,10 +293,10 @@ static NSString *const UnzippedFolderName = @"unzipped";
 + (NSDictionary *)getCurrentPackage:(NSError **)error
 {
     NSString *packageHash = [CodePushPackage getCurrentPackageHash:error];
-    if (*error || !packageHash) {
+    if (!packageHash) {
         return nil;
     }
-    
+
     return [CodePushPackage getPackage:packageHash error:error];
 }
 
@@ -304,14 +304,14 @@ static NSString *const UnzippedFolderName = @"unzipped";
 {
     NSString *packageFolder = [self getCurrentPackageFolderPath:error];
     
-    if(*error) {
-        return NULL;
+    if (!packageFolder) {
+        return nil;
     }
     
     NSDictionary *currentPackage = [self getCurrentPackage:error];
     
-    if(*error) {
-        return NULL;
+    if (!currentPackage) {
+        return nil;
     }
     
     NSString *relativeBundlePath = [currentPackage objectForKey:RelativeBundlePathKey];
@@ -325,7 +325,7 @@ static NSString *const UnzippedFolderName = @"unzipped";
 + (NSString *)getCurrentPackageHash:(NSError **)error
 {
     NSDictionary *info = [self getCurrentPackageInfo:error];
-    if (*error || !info) {
+    if (!info) {
         return nil;
     }
     
@@ -336,14 +336,14 @@ static NSString *const UnzippedFolderName = @"unzipped";
 {
     NSDictionary *info = [self getCurrentPackageInfo:error];
     
-    if (*error) {
-        return NULL;
+    if (!info) {
+        return nil;
     }
     
     NSString *packageHash = info[@"currentPackage"];
     
     if (!packageHash) {
-        return NULL;
+        return nil;
     }
     
     return [self getPackageFolderPath:packageHash];
@@ -359,7 +359,7 @@ static NSString *const UnzippedFolderName = @"unzipped";
     NSString *content = [NSString stringWithContentsOfFile:statusFilePath
                                                   encoding:NSUTF8StringEncoding
                                                      error:error];
-    if (*error) {
+    if (!content) {
         return nil;
     }
     
@@ -367,7 +367,7 @@ static NSString *const UnzippedFolderName = @"unzipped";
     NSDictionary* json = [NSJSONSerialization JSONObjectWithData:data
                                                          options:kNilOptions
                                                            error:error];
-    if (*error) {
+    if (!json) {
         return nil;
     }
     
@@ -392,8 +392,7 @@ static NSString *const UnzippedFolderName = @"unzipped";
     NSString *updateMetadataString = [NSString stringWithContentsOfFile:updateMetadataFilePath
                                                                encoding:NSUTF8StringEncoding
                                                                   error:error];
-    
-    if (*error) {
+    if (!updateMetadataString) {
         return nil;
     }
     
@@ -411,7 +410,7 @@ static NSString *const UnzippedFolderName = @"unzipped";
 + (NSDictionary *)getPreviousPackage:(NSError **)error
 {
     NSString *packageHash = [self getPreviousPackageHash:error];
-    if (*error || !packageHash) {
+    if (!packageHash) {
         return nil;
     }
     
@@ -421,7 +420,7 @@ static NSString *const UnzippedFolderName = @"unzipped";
 + (NSString *)getPreviousPackageHash:(NSError **)error
 {
     NSDictionary *info = [self getCurrentPackageInfo:error];
-    if (*error) {
+    if (!info) {
         return nil;
     }
     
@@ -438,25 +437,25 @@ static NSString *const UnzippedFolderName = @"unzipped";
     return [[self getCodePushPath] stringByAppendingPathComponent:UnzippedFolderName];
 }
 
-+ (void)installPackage:(NSDictionary *)updatePackage
++ (BOOL)installPackage:(NSDictionary *)updatePackage
    removePendingUpdate:(BOOL)removePendingUpdate
                  error:(NSError **)error
 {
     NSString *packageHash = updatePackage[@"packageHash"];
     NSMutableDictionary *info = [self getCurrentPackageInfo:error];
     
-    if (*error) {
-        return;
+    if (!info) {
+        return NO;
     }
     
     if (packageHash && [packageHash isEqualToString:info[@"currentPackage"]]) {
         // The current package is already the one being installed, so we should no-op.
-        return;
+        return YES;
     }
 
     if (removePendingUpdate) {
         NSString *currentPackageFolderPath = [self getCurrentPackageFolderPath:error];
-        if (!*error && currentPackageFolderPath) {
+        if (currentPackageFolderPath) {
             // Error in deleting pending package will not cause the entire operation to fail.
             NSError *deleteError;
             [[NSFileManager defaultManager] removeItemAtPath:currentPackageFolderPath
@@ -467,7 +466,7 @@ static NSString *const UnzippedFolderName = @"unzipped";
         }
     } else {
         NSString *previousPackageHash = [self getPreviousPackageHash:error];
-        if (!*error && previousPackageHash && ![previousPackageHash isEqualToString:packageHash]) {
+        if (previousPackageHash && ![previousPackageHash isEqualToString:packageHash]) {
             NSString *previousPackageFolderPath = [self getPackageFolderPath:previousPackageHash];
             // Error in deleting old package will not cause the entire operation to fail.
             NSError *deleteError;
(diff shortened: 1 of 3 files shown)


ios/CodePush/CodePushPackage.m
```diff
@@ -292,26 +292,41 @@ + (NSString *)getCodePushPath
...
+        if (localError && error) {
+            *error = localError;
+        }
         return nil;
     }
-    
-    return [CodePushPackage getPackage:packageHash error:error];
+
+    NSDictionary *package = [CodePushPackage getPackage:packageHash error:&localError];
+    if (!package && localError && error) {
```
geof90: This seems a little unnecessary (at least to me). Why not just `return [CodePushPackage getPackage:packageHash error:error];`? If `[CodePushPackage getPackage]` deals with the null pointer deference problem too, then that line already achieves all of your goals - 1. To return the result anyway, 2. To only set the `error` pointer pointer if its not nil.

daniel-beard: Good point, I'll update this.

daniel-beard:
Not sure how I can sign the CLA, can someone point me in the right direction?

ios/CodePush/CodePushPackage.m
```diff
@@ -391,16 +418,26 @@ + (NSDictionary *)getPackage:(NSString *)packageHash
...
     }
     
     NSData *updateMetadata = [updateMetadataString dataUsingEncoding:NSUTF8StringEncoding];
-    return [NSJSONSerialization JSONObjectWithData:updateMetadata
+    id result = [NSJSONSerialization JSONObjectWithData:updateMetadata
                                            options:kNilOptions
-                                             error:error];
+                                             error:&localError];
+
+    if (!result) {
```
geof90: Same comment here.. NSJSONSerialization accepts a nullable error parameter.

ios/CodePush/CodePushPackage.m
```diff
@@ -410,18 +447,33 @@ + (NSString *)getPackageFolderPath:(NSString *)packageHash
...
+        if (localError && error) {
+            *error = localError;
+        }
         return nil;
     }
     
-    return [CodePushPackage getPackage:packageHash error:error];
+    NSDictionary *package = [CodePushPackage getPackage:packageHash error:&localError];
+
+    if (!package) {
```
geof90: Same here.

ios/CodePush/CodePushPackage.m
daniel-beard: If this fails, wouldn't you want the caller of this method to know that it was a failure within `NSJSONSerialization` ?

ios/CodePush/CodePushPackage.m
daniel-beard: Will Update

ios/CodePush/CodePushPackage.m
```diff
@@ -481,28 +537,33 @@ + (void)installPackage:(NSDictionary *)updatePackage
...
     
     [info setValue:packageHash forKey:@"currentPackage"];
-    
+
+    //TODO: This method should return a BOOL indication of success.
     [self updateCurrentPackageInfo:info
-                             error:error];
+                             error:&localError];
+
+    if (localError && error) {
```
geof90: Here too

ios/CodePush/CodePushPackage.m
```diff
@@ -481,28 +537,33 @@ + (void)installPackage:(NSDictionary *)updatePackage
...
+    if (!currentPackageFolderPath) {
         return;
     }
     
     NSError *deleteError;
-    [[NSFileManager defaultManager] removeItemAtPath:currentPackageFolderPath
+    BOOL result = [[NSFileManager defaultManager] removeItemAtPath:currentPackageFolderPath
                                                error:&deleteError];
-    if (deleteError) {
+    if (!result) {
```
geof90: This probably isn't part of your change scope, but I think we should print the error too.

geof90:
@daniel-beard You can sign the CLA here: https://cla.microsoft.com/

ios/CodePush/CodePushPackage.m
geof90: Yes, but that would be achieved by passing the `error` pointer that the caller gave to us directly no?

daniel-beard: That's happening below:

```
if (!result) {
  if (localError && error) {   // Makes sure we are dereferencing a null pointer
    *error = localError;         // Assigns the value to the `error` parameter that is passed into the method
  }
}
```

ios/CodePush/CodePushUpdateUtils.m
```diff
@@ -238,18 +273,23 @@ + (BOOL)verifyHashForDiffUpdate:(NSString *)finalUpdateFolder
...
 {
+    NSError *localError;
     NSMutableArray *updateContentsManifest = [NSMutableArray array];
     [self addContentsOfFolderToManifest:finalUpdateFolder
                              pathPrefix:@""
                                manifest:updateContentsManifest
-                                  error:error];
-    if (*error) {
+                                  error:&localError];
+    if (localError && error) {
```
geof90: I think we should return `NO` if the `localError` is set, regardless of whether the `error` is nil?

If you don't like that, feel free to modify the methods in this file to return boolean flags indicating success, since this file doesn't contain public documented APIs.

daniel-beard: Yeah I think that would be best, I'll update.

ios/CodePush/CodePushPackage.m
geof90: Right.. what I meant was `NSJSONSerialization` takes care of not dereferencing null pointers for you since the documentation says the error param can be nullable. Ultimately, what I'm suggesting is just for brevity, since your code does achieve the same thing

geof90:
@daniel-beard Sorry do you mind resolving the conflicts? I want to get this merged. Thanks!

daniel-beard:
For sure, updating now.

daniel-beard:
Ok @geof90 updated and squashed.

ios/CodePush/CodePushPackage.m
```diff
@@ -292,26 +292,37 @@ + (NSString *)getCodePushPath
...
+    if (!packageHash) {
+        if (localError && error) {
+            *error = localError;
+        }
         return nil;
     }
-    
-    return [CodePushPackage getPackage:packageHash error:error];
+
+    return [CodePushPackage getPackage:packageHash error:&localError];
```
geof90: Shouldn't this be `error:error`?

ios/CodePush/CodePushUpdateUtils.m
```diff
@@ -196,19 +229,22 @@ + (NSString *)getHashForBinaryContents:(NSURL *)binaryBundleUrl
...
+            }
             return nil;
         }
     }
     
     [self addFileToManifest:binaryBundleUrl manifest:manifest];
     [self addFileToManifest:[binaryBundleUrl URLByAppendingPathExtension:@"meta"] manifest:manifest];
 
-    binaryHash = [self computeFinalHashFromManifest:manifest error:error];
+    binaryHash = [self computeFinalHashFromManifest:manifest error:&localError];
```
geof90: `error` instead of `&localError`?

daniel-beard: Fixed

ios/CodePush/CodePushPackage.m
daniel-beard: Fixed

geof90:
I'm sorry for dragging out this PR longer than it needs to, but thinking about this more, I still have a question over why this is the right practice:

1. I see that the motivation for this fix is that we should guard against nil dereferencing `NSError **` pointers, e.g. doing things like `if (*error) { return nil }`.

2. However, you brought out a second problem, which is that checking the error object in the first place is "fragile". I think I am convinced about that.

3. As a result, I'd argue that the fix for 2 would also fix 1 (because we no longer need nor should check returned error pointers, and therefore, dereference them). So why are all the local `*localError` variables needed then?

I'll leave some inline comments to better illustrate what I mean.

ios/CodePush/CodePushPackage.m
```diff
@@ -292,26 +292,37 @@ + (NSString *)getCodePushPath
 
 + (NSDictionary *)getCurrentPackage:(NSError **)error
 {
-    NSString *packageHash = [CodePushPackage getCurrentPackageHash:error];
-    if (*error || !packageHash) {
+    NSError *localError;
+    NSString *packageHash = [CodePushPackage getCurrentPackageHash:&localError];
```
geof90: What would be the drawback if we simply passed `error` instead of `&localError` here, if we know that `getCurrentPackageHash` already guards against nil dereferencing `error`?

ios/CodePush/CodePushPackage.m
```diff
@@ -292,26 +292,37 @@ + (NSString *)getCodePushPath
...
+    if (!packageHash) {
+        if (localError && error) {
+            *error = localError;
+        }
```
geof90: I think if we passed `error` directly instead of `&localError`, we could simply have

```
if (!packageHash) {
    return nil;
}
```

and get rid of these lines:
```
if (localError && error) {
    *error = localError;
}
```

ios/CodePush/CodePushPackage.m
```diff
@@ -292,26 +292,37 @@ + (NSString *)getCodePushPath
...
     }
-    
+
     return [CodePushPackage getPackage:packageHash error:error];
 }
 
 + (NSString *)getCurrentPackageBundlePath:(NSError **)error
 {
-    NSString *packageFolder = [self getCurrentPackageFolderPath:error];
+    NSError *localError;
```
geof90: If possible, I would like to get rid of the need to create this new `localError` entirely, here and the rest of your changes, unless they serve some other purpose which I am missing.

geof90:
Hi @daniel-beard, I want to respect your time, so if you want me to just merge this and I can continue where you left off later, feel free to let me know!

daniel-beard:
I can address the rest of the changes tomorrow :D

daniel-beard:
Ok @geof90 I scrubbed through and removed all the `localError` vars, and made sure there aren't any more fragile error checks. This should be good to go now.

geof90:
Yes! this looks a lot cleaner (and shorter). I will merge this, test and get it published ASAP.

PR #506, opened 2016-09-01
Check against self class in initialize methods

daniel-beard:
- Best practice for `initialize` methods is to check against the current class, otherwise if someone subclasses, the method may be run more than once.
- See: https://developer.apple.com/library/mac/documentation/Cocoa/Reference/Foundation/Classes/NSObject_Class/index.html#//apple_ref/occ/clm/NSObject/initialize


geof90:
Thanks for your contribution!

PR #507, opened 2016-09-01
Fix implicit conversion warning

daniel-beard:
- Fixes a build warning about implicit conversion loss of precision.
- Fixed by adding an explicit cast to the correct type (`CC_LONG`).


geof90:
Thanks!

Issue #501, opened 2016-08-31
Cannot link Android library because of directory structure assumptions


Zenton:
Hi,

I'm trying to integrate CodePush into my React-Native apps. I'm running into issues when linking the project as the compilation code (specifically for Android) makes assumptions about the directory structure (I've also had to manually link iOS because of structure assumptions).

My structure is `src/main/javascript/com/[company]/mobile/apps/[app]/[react root here]`, which means when CodePush's files use `../../node_modules` (examples below) it cannot find the file as `node_modules` is at the root of the repo.

Is there anyway to pass a root flag to CodePush so it uses the right path?

Thanks

Examples:
1. https://github.com/Microsoft/react-native-code-push/blob/5ab949ab9267cdc22cc00723818d109ee4a9c0d5/scripts/postlink/android/postlink.js
2. https://github.com/Microsoft/react-native-code-push/blob/1bf4a12feddd0bda80f61216b72dd1be0c6841a5/android/codepush.gradle#L43

geof90:
Hi @Zenton, so your node_modules folder doesn't live in the react root? Because if it does it should technically just work.

It will be a nice feature to add, I agree. We also documented the way to manually link our android native library [here](https://github.com/Microsoft/react-native-code-push#plugin-installation-android---manual) so you could use that as a workaround for now.

Zenton:
Hi @geof90,

My node_modules lives at my repo root (i.e. my node root). I just have a few more directories between it and my react `ios/android` directories (standard structuring of code based on purpose).

Before posting I tried using the Android manual link and it fails because it tries to run the second script I linked to, which also makes assumptions about directory structure. 

So unfortunately I can't use manual linking (without manually editing files in my `node_modules` directory which obviously doesn't work with a team).

Do you have any ideas for a different workaround?

Thanks

geof90:
How about if we have you declare a project variable, like so:
```
project.ext.set("nodeModulesPath", "path/to/node_modules");
```

And we can update codepush.gradle to use that if its set, otherwise, fall back to the default location. Does that sound ok to you?

Zenton:
Good idea, that'd work well.

silhouettes:
This fix is now available in `1.15.0-beta`. Release notes can be found at: https://github.com/Microsoft/react-native-code-push/releases/tag/v1.15.0-beta. Thanks so much for the patience, and we look forward to your feedback! Please let us know if you still run into any issues.

PR #508, opened 2016-09-02
Allow node_modules path to be overridden in codepush.gradle

geof90:
Addresses https://github.com/Microsoft/react-native-code-push/issues/501.

### DIFF
--- a/android/codepush.gradle
+++ b/android/codepush.gradle
@@ -18,6 +18,12 @@ gradle.projectsEvaluated {
     def buildTypes = android.buildTypes.collect { type -> type.name }
     def productFlavors = android.productFlavors.collect { flavor -> flavor.name }
     if (!productFlavors) productFlavors.add('')
+    def nodeModulesPath;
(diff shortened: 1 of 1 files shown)


android/codepush.gradle
```diff
@@ -18,6 +18,12 @@ gradle.projectsEvaluated {
     def buildTypes = android.buildTypes.collect { type -> type.name }
     def productFlavors = android.productFlavors.collect { flavor -> flavor.name }
     if (!productFlavors) productFlavors.add('')
+    def nodeModulesPath;
+    if (project.hasProperty('nodeModulesPath')) {
+        nodeModulesPath = project.nodeModulesPath
```
silhouettes: When people override the module path, what location are they expecting to set the path relative to?

silhouettes:
Looks good!

android/codepush.gradle
geof90: It should be relative to their android project root.

PR #525, opened 2016-09-22
[doc] advice when targeting both android/ios

lklots:
Based on a discussion with silhouettes

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ We try our best to maintain backwards compatability of our plugin with previous 
 
 ## Supported Components
 
-When using the React Native assets sytem (i.e. using the `require("./foo.png")` syntax), the following list represents the set of core components (and props) that support having their referenced images updated via CodePush:
+When using the React Native assets system (i.e. using the `require("./foo.png")` syntax), the following list represents the set of core components (and props) that support having their referenced images updated via CodePush:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -82,7 +82,7 @@ Once you've followed the general-purpose ["getting started"](http://codepush.too
 npm install --save react-native-code-push@latest
 ```
 
-As with all other React Native plugins, the integration experience is different for iOS and Android, so perform the following setup steps depending on which platform(s) you are targetting.
+As with all other React Native plugins, the integration experience is different for iOS and Android, so perform the following setup steps depending on which platform(s) you are targeting. Note, if you are targeting both platforms it is recommended to create separate code push applications for each platform.
```
silhouettes: Could you change `code push` to `CodePush`?

vojtechnovak:
Hello! I'd be interested in knowing why exactly this is recommended.? Is it only to control whether an update that affects say only android will be downloaded by iOS devices too?

silhouettes:
@vojtechnovak - yes - currently, we don't have any logic to distinguish between the two platforms, so it's important that they are kept separate.

README.md
silhouettes: I will merge this and then fix it up afterwards. Thanks for the PR!

PR #535, opened 2016-09-28
Check that the previous package is not null

cmcewen:
This should hopefully prevent #515 from happening. https://github.com/Microsoft/react-native-code-push/blob/master/android/app/src/main/java/com/microsoft/codepush/react/CodePushNativeModule.java#L288 checks that the currentPackage is not null, so the only way that `convertJsonObjectToWritable` is called with a null argument is if `mUpdateManager.getPreviousPackage()` returns null.


cmcewen:
I'm pretty confident this fixes the crash. We were able to consistently reproduce this crash before by installing a version of the app that had an older JS bundle, waiting for the codepush update to install in the background, then switching to a screen that would call `CodePush.getUpdateMetadata();`

Now following those same steps, we don't see a crash.

PR #545, opened 2016-10-04
fix(AppRestart): Fix update-on-app-restart for UWP

rozele:
The following sequence seems to be locking up the application for update-on-app-restart:
1) Start app without any updates on CodePush
2) Push a CodePush update
3) Restart the app, the update will be downloaded and installed
4) Restart the app again, the app will claim that a rollback is required and hang

The source of the issue is that CodePushReactPackage.InitializeUpdateAfterRestart() is called twice, once in the constructor of CodePushReactPackage, when it sets PendingUpdateIsLoading to true. The second time it's called is from CodePushNativeModule.Initialize, at this point the "PendingUpdateIsLoading" is read out as true, and the app thinks a pending update was not finished, so it tries to Rollback.

Everything seems to work still after having removed the constructor call.

### DIFF
--- a/windows/CodePushReactPackage.cs
+++ b/windows/CodePushReactPackage.cs
@@ -39,7 +39,6 @@ namespace CodePush.ReactNative
             // TODO implement telemetryManager 
             // _codePushTelemetryManager = new CodePushTelemetryManager(this.applicationContext, CODE_PUSH_PREFERENCES);
 
-            InitializeUpdateAfterRestart();
             if (CurrentInstance != null)
             {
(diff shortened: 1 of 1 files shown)


geof90:
Hi @rozele, did you test this with the immediate and on resume install modes?

rozele:
@geof90 testing now.

rozele:
@geof90 works with ON_APP_RESUME but I can't get IMMEDIATE to work because of #546.  Trying a manual check now.

rozele:
Actually, nevermind.  Pushed up a second update, restarted the app (with an installed update already in queue), the second update was downloaded and app reloaded without any problems.

geof90:
Thanks @rozele!

PR #585, opened 2016-10-27
android: allow `assetsDir` and `resourcesDir` can be overridden

ianlin:
I'm building an android library project using react-native and code-push.
### assetsDir

When compiling an android library project the `assets` directory is located at

`${buildDir}/intermediates/bundles/${targetPath}/assets`

instead of

`${buildDir}/intermediates/assets/${targetPath}`

So I need to specify not only `jsBundleDir` but `assetsDir`, otherwise it cannot find the correct assets path in `generateBundledResourcesHash.js`
### resourcesDir

As of `resourcesDir`, it was originally overridden by `jsBunbleDir`. It would be better to be overridden by a separated variable like that in [react-native/react.gradle](https://github.com/facebook/react-native/blob/v0.36.0/react.gradle#L46)

### DIFF
--- a/android/codepush.gradle
+++ b/android/codepush.gradle
@@ -32,12 +32,15 @@ gradle.projectsEvaluated {
                     "${productFlavorName}/${buildTypeName}" :
                     "${buildTypeName}"
 
+            def assetsDirConfigName = "assetsDir${targetName}"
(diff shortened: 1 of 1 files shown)


silhouettes:
Closing this until we get more feedback that it is needed!

PR #587, opened 2016-11-01
Fix comparison method in setJSBundle

Swordsman-Inaction:
Simple Fix:

The old comparison method is wrong:
`if (method.getName() == "createFileLoader")`

`==` comparison fails on Samsung S5,

Change it to:
`if (method.getName().equals("createFileLoader"))`


silhouettes:
Great catch. It's strange because this had definitely been tested and was working before - from my brief research it looks like Java often (but not always) does optimizations to re-use the same object when they are the same string. Thanks so much for finding this!

silhouettes:
I will make sure this is released soon.

Swordsman-Inaction:
@silhouettes normally it should work, I've tested it on several devices too, it only fails on Samsung S5, just a edge case.

PR #601, opened 2016-11-11
Update broken 'install RN with cocoapods' link

dingbat:
Small docs PR, just updates the link to the React Native documentation page of instructions on specifying React as a pod, since the current one now redirects to a different page.


silhouettes:
Thanks for your contribution @dingbat!

Issue #530, opened 2016-09-26
rnpm link react-native-code-push make mistake


fanyuRN:
![c30a65b6-7460-471d-b885-95666af14fe7](https://cloud.githubusercontent.com/assets/20810414/18825373/a55556c0-83fa-11e6-8c2f-42323087ef1f.png)
my rn is 0.32;code-push is 1.14.6 beta,thx

silhouettes:
Could you try running `react-native link react-native-code-push` instead of `rnpm link`?

joeytat:
I got the same issue. Turned out you have to put `jsCodeLocation = [CodePush bundleURL];` into the AppDelegate.m. Anywhere else will cause this issue.

silhouettes:
Hi @joeytat, @fanyuRN, @Lughino - I am investigating this now and would really appreciate your input. The linking script looks for the following lines (or similar) in `AppDelegate.m` in order to patch them for CodePush, and appears to be unable to find them:

```
jsCodeLocation = [[RCTBundleURLProvider sharedSettings] jsBundleURLForBundleRoot:@"index.ios" fallbackResource:nil];

  RCTRootView *rootView = [[RCTRootView alloc] initWithBundleURL:jsCodeLocation
                                                      moduleName:@"DummyApp"
                                               initialProperties:nil
                                                   launchOptions:launchOptions];
```

Can I ask if you have these lines in some other file (if so, which)? Was your app created from scratch with React Native, or are you integrating RN into an app that was originally built natively?

joeytat:
@silhouettes 
1) Yes, I thought keep `AppDelegate.m` neat is nice. so I put those code in another file called 'XXRNManager'. 
2) Integration with existing app.

Lughino:
Hi @silhouettes 
* No
* App created from scratch with react native

silhouettes:
Thanks, this is really helpful.

@Lughino - Did you have a `jsCodeLocation = ...` line in your `AppDelegate.m` originally, before you ran `react-native link`? Is it possible you had multiple `AppDelegate.m` files, for example if there was a sample app embedded in your repo somewhere?

Lughino:
@silhouettes - Yes, that line of code is present in the `AppDelegate.m` before launching `react-native link`.
I also verified by creating a project from scratch with no external dependencies and no multiple `AppDelegate.m` in any project folder.

I hope I was helpful

Lughino:
@fanyuRN why you closed this issue?

[...]

PR #603, opened 2016-11-13
Fix occasionally break in oldJsCodeLocationAssignmentStatement

Lughino:
Fix issue #530.
Every so often you run the command `react-native link` an exception is thrown because the `\n` in regex does not lead to the line.


silhouettes:
@Lughino - Ahh, this explains a lot! I'm guessing this is a Windows vs OS X problem. Thanks so much for your contribution, I will merge this.

PR #609, opened 2016-11-16
Correct console message in status change example

tristanbbq:


silhouettes:
Thanks for your contribution!

PR #620, opened 2016-11-24
Update README for iOS Manual plugin installation

sogko:
Wrong/incomplete path to Codepush iOS library set in the `Header Search Paths` causes Xcode to fail without specifying any errors.

The path to the library should be changed from 

```
$(SRCROOT)/../node_modules/react-native-code-push

to

$(SRCROOT)/../node_modules/react-native-code-push/ios
```

screen shot 2016-11-24 at 10 26 36 am

Small detail but might cause a developer significant amount of time trying to pin down the cause of the build failure, especially when Xcode does not return any useful error.

The Xcode log when the header search path is wrong:

```
# Exit Status 65
Verify final result code for completed build operation

Build operation failed without specifying any errors. Individual build tasks may have failed for unknown reasons.
One possible cause is if there are too many (possibly zombie) processes; in this case, rebooting may fix the problem.
Some individual build task failures (up to 12) may be listed below.

```

screen shot 2016-11-24 at 10 12 01 am

Cheers!

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,7 +165,7 @@ And that's it! Isn't RNPM awesome? :)
 6. Under the "Build Settings" tab of your project configuration, find the "Header Search Paths" section and edit the value.
 Add a new value, `$(SRCROOT)/../node_modules/react-native-code-push/ios` and select "recursive" in the dropdown.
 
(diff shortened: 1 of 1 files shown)


silhouettes:
Thanks for the contribution @sogko!

PR #622, opened 2016-11-27
Remove MoveIt! from example section

sotayamashita:
Remove Movelt! bacause it is 404 and not maintained anymore according to https://github.com/multunus/moveit-mobile.


silhouettes:
Thanks for the contribution @sotayamashita!

PR #623, opened 2016-11-27
Change codePush -> CodePush

sotayamashita:


silhouettes:
Hi @sotayamashita, I don't think we should accept this change, because we use the camel-case `codePush` all throughout the docs. If you have any other thoughts or contributions we are all ears though!

PR #624, opened 2016-11-28
Upgrade to rn 0.38

buptkang:
There is no issues detected by using RN 0.38 instead of previous RN 0.37. Only readme file is updated.


Examples/CodePushDemoApp/demo.js
```diff
@@ -103,7 +103,7 @@ class CodePushDemoApp extends Component {
     return (
       <View style={styles.container}>
         <Text style={styles.welcome}>
-          Welcome to CodePush!
+          Welcome to CodePush, test.
```
silhouettes: Might want to remove the `test` addition here

PR #637, opened 2016-12-12
Update header paths to compile with RN@0.40

vladimir-kotikov:
Since RN 0.40 headers are restructured and copied to product directory (see [facebook/react-native/e1577df)](https://github.com/facebook/react-native/commit/e1577df1fd70049ce7f288f91f6e2b18d512ff4d). To compile CodePush library which has a dependency on these headers, we need to add new location to headers search paths. Old paths still need to exist to keep compatibility w/ earlier RN versions

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,7 +163,12 @@ And that's it! Isn't RNPM awesome? :)
     *Note: Alternatively, if you prefer, you can add the `-lz` flag to the `Other Linker Flags` field in the `Linking` section of the `Build Settings`.*
 
 6. Under the "Build Settings" tab of your project configuration, find the "Header Search Paths" section and edit the value.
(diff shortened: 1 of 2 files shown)


vladimir-kotikov:
Looks like RN has changed their headers location again (see [63d5bcb](https://github.com/facebook/react-native/commit/63d5bcb49bec51123fe36e93ad0e44aaeef966d5)). Need to double check if this PR is still required

silhouettes:
@vladimir-kotikov - whatever the final fix is, please make sure the compatibility table in the README.md is updated, and that the fix is backwards compatible with older RN versions (or if not, that the breaking change is a conscious decision and is reflected in the compatibility table)

PR #638, opened 2016-12-12
Upgrade example app to react-native 0.40.0

vladimir-kotikov:
@silhouettes, not sure if we need to merge this since RN@0.40 is in RC state yet - we might want to wait unitl 0.40 is released, test it and update the PR accordingly (if needed)

UPD.: Refreshed the PR since react-native@0.40 has been finally released.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,8 +53,8 @@ We try our best to maintain backwards compatability of our plugin with previous 
 | v0.29-v0.30             | v1.13.0+ *(RN refactored native hosting code)* |
 | v0.31-v0.33             | v1.14.6+ *(RN refactored native hosting code)* |
 | v0.34-v0.35             | v1.15.0+ *(RN refactored native hosting code)* |
-| v0.36-v0.39             | v1.16.0+ *(RN refactored resume handler)*      |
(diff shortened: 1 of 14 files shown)


silhouettes:
Great stuff! I agree - we should confirm 0.40 compatibility is solid before merging, as that is the version our users will install

README.md
```diff
@@ -53,8 +53,8 @@ We try our best to maintain backwards compatability of our plugin with previous
 | v0.29-v0.30             | v1.13.0+ *(RN refactored native hosting code)* |
 | v0.31-v0.33             | v1.14.6+ *(RN refactored native hosting code)* |
 | v0.34-v0.35             | v1.15.0+ *(RN refactored native hosting code)* |
-| v0.36-v0.39             | v1.16.0+ *(RN refactored resume handler)*      |
-| v0.40+                  | TBD :) We work hard to respond to new RN releases, but they do occasionally break us. We will update this chart with each RN release, so that users can check to see what our "official" support is.
+| v0.36-v0.40             | v1.16.0+ *(RN refactored resume handler)*      |
```
silhouettes: @buptkang - I think this line is incorrect, right?

Issue #584, opened 2016-10-26
Execution failed for task ':app:recordFilesBeforeBundleCommandDebug'.


zhangjunhou:
:app:compileDebugAidl
:app:compileDebugRenderscript
:app:generateDebugBuildConfig
:app:generateDebugAssets UP-TO-DATE
:app:mergeDebugAssets
:app:generateDebugResValues
:app:generateDebugResources
:app:mergeDebugResources
AAPT: /Users/zhangjunhou/APP/android/app/src/main/res/drawable-xxxhdpi/splash.png: libpng warning: iCCP: Not recognizing known sRGB profile that has been edited
AAPT: /Users/zhangjunhou/APP/android/app/src/main/res/drawable-mdpi/images_share.png: libpng warning: iCCP: Not recognizing known sRGB profile that has been edited
AAPT: /Users/zhangjunhou/APP/android/app/build/intermediates/exploded-aar/lucky/rn-splash-screen/unspecified/res/drawable/splash.png: libpng warning: iCCP: Not recognizing known sRGB profile that has been edited
AAPT: /Users/zhangjunhou/APP/android/app/src/main/res/drawable-xxhdpi/splash.png: libpng warning: iCCP: Not recognizing known sRGB profile that has been edited
AAPT: /Users/zhangjunhou/APP/android/app/src/main/res/drawable-hdpi/splash.png: libpng warning: iCCP: Not recognizing known sRGB profile that has been edited
AAPT: /Users/zhangjunhou/APP/android/app/src/main/res/drawable-xhdpi/splash.png: libpng warning: iCCP: Not recognizing known sRGB profile that has been edited
AAPT: /Users/zhangjunhou/APP/android/app/src/main/res/drawable-mdpi/splash.png: libpng warning: iCCP: Not recognizing known sRGB profile that has been edited
:app:recordFilesBeforeBundleCommandDebug
fs.js:951
  return binding.readdir(pathModule._makeLong(path), options.encoding);
                 ^

Error: ENOENT: no such file or directory, scandir '/Users/zhangjunhou/APP/android/app/build/intermediates/res/merged/debug'
    at Error (native)
    at Object.fs.readdirSync (fs.js:951:18)
    at getFilesInFolder (/Users/zhangjunhou/APP/node_modules/react-native-code-push/scripts/getFilesInFolder.js:7:26)
[...]

PR #651, opened 2016-12-23
codepush.gradle: add support for Android Plugin for Gradle with version lower than 1.3.0

sergey-akhalkov:
In case version of `Android Plugin for Gradle` is lower than 1.3.0
`$buildDir` has slightly different structure - `merged` folder
does not exists so `${targetPath}` folder contains directly in `res` folder.

Fix issue https://github.com/Microsoft/react-native-code-push/issues/584


silhouettes:
Nice work!

Issue #650, opened 2016-12-22
getBinaryResourcesModifiedTime always returns the same time on Android


jacob-meacham:
* react-native-code-push version: 1.16.0-beta
* react-native version: 0.36.0
* iOS/Android/Windows version: Android 6.0.1, API 23, gradle 2.14.1

The timestamp returned in getBinaryResourcesModifiedTime() always returns 312796800000 (11/30/1979, 12:00:00 AM). This is expected as per https://code.google.com/p/android/issues/detail?id=220039. The workaround to set android.keepTimestampsInApk = true is an ok one, but in a default project, the binary version is never loaded once a codepush bundle has been installed.

I'm happy to cook up a PR to fix this issue, but I am not primarily an android dev and wanted to get some feedback on an appropriate fix. My initial proposal is to set a buildConfigField in codepush.gradle, and then check that at runtime.

Thanks!

vladimir-kotikov:
Looks like this is exactly the same problem as with cordova plugin here: https://github.com/Microsoft/cordova-plugin-code-push/issues/196

matrosov-nikita:
Hi @jacob-meacham 

Yeah, I see the timestamp within update and built apk time are equal, but there is [additional check](https://github.com/Microsoft/react-native-code-push/blob/master/android/app/src/main/java/com/microsoft/codepush/react/CodePush.java#L187) for binary versions equality. Even though the timestamps are the same, after build step plugin cleans out current local package (if you bump version) and anyway downloads new updates. Could you provide specific case which demonstrates your problem?

jacob-meacham:
@matrosov-nikita yeah, on debug builds right now we're not bumping version appropriately. We should definitely fix that on our side, but it's why I noticed this issue. If the code is just dead, maybe the better solution is just to remove it?

matrosov-nikita:
@jacob-meacham, I've sent the PR. It uses `buildConfigField` to get correct apk built time as you've already suggested. Could you pls check whether it works for you?

jacob-meacham:
This is similar to the hacked fix I had been using and works for me. Thanks!

PR #654, opened 2016-12-27
Get correct apk built time

matrosov-nikita:
We should use `buildConfigField` to get correct apk build time since gradle zeroes out built timestamp(https://github.com/Microsoft/react-native-code-push/issues/650)

### DIFF
--- a/README.md
+++ b/README.md
@@ -294,9 +294,11 @@ public class MainApplication extends Application implements ReactApplication {
         ...
         // 2. Override the getJSBundleFile method in order to let
         // the CodePush runtime determine where to get the JS
-        // bundle location from on each app start
+        // bundle location from on each app start. Set manually apk built time
+        // since gradle zeroes built timestamps
         @Override
         protected String getJSBundleFile() {
+            CodePush.setBuildTimestamp(BuildConfig.CODE_PUSH_APK_BUILD_TIME);
             return CodePush.getJSBundleFile();
         }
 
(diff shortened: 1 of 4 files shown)


panarasi:
@silhouettes Can you review this, and if it looks ok, lets merge this in.

panarasi:
Once this PR is merged, please close #650

silhouettes:
Tagging @sergey-akhalkov and @vladimir-kotikov seeing as @matrosov-nikita may not be able to follow up on this!

sergey-akhalkov:
@silhouettes, I've probably found the way we could get `apk_build_time` value without adding `CodePush.setBuildTimeStamp(...)` line. We need to add following things:
1. add this lines
```groovy
android.buildTypes.each {
    it.resValue 'string', 'CODE_PUSH_APK_BUILD_TIME', System.currentTimeMillis() + "L"
}
```
into `codepush.gradle` file. It allows us to store build time value locally into `app/src/main/res/values/strings.xml` file (content of `strings.xml` file [would not be changed physically](http://stackoverflow.com/a/30879430), just on the fly, but I should verify it twice). 
2. then we should read `CODE_PUSH_APK_BUILD_TIME` value from `app/src/main/res/values/strings.xml` file like this:
```Java
String packageName = this.mContext.getPackageName();
int codePushApkBuildTimeId = this.mContext.getResources().getIdentifier("CODE_PUSH_APK_BUILD_TIME", "string", packageName);
String codePushApkBuildTime = this.mContext.getResources().getString(codePushApkBuildTimeId);
```
and convert `codePushApkBuildTime ` variable value to `long`.

What do you think?

sergey-akhalkov:
@silhouettes, I've created https://github.com/Microsoft/react-native-code-push/pull/683. It does not contain manual installation steps and any docs changes, it uses `app/src/main/res/values/strings.xml` file to store apk build time value. `codepush.grade` does not change `app/src/main/res/values/strings.xml` file physically, it inserts `CODE_PUSH_APK_BUILD_TIME` key on the fly (while building apk).

silhouettes:
Closing this PR in favor of #683

Issue #662, opened 2017-01-06
'CodePush.h' file not found


sergey-akhalkov:
* react-native-code-push version: v1.16.1-beta
* react-native version: v0.40.0
* iOS/Android/Windows version: iOS
* Does this reproduce on a debug build or release build? all
* Does this reproduce on a simulator, or only on a physical device? all
* Repro steps and/or stack trace:
1. Execute `react-native init myapp`
2. Execute `cd myapp` then `npm install --save react-native-code-push@1.16.1-beta`
3. Execute `react-native link react-native-code-push`
4. Then after build and run `myapp` using xcode I've got following error:
![image](https://cloud.githubusercontent.com/assets/4928157/21715111/bca0ef60-d413-11e6-83af-d169ea9f37a0.png)

Seems like `react-native link react-native-code-push` command works wrong.
Created issue in react-native repo: https://github.com/facebook/react-native/issues/11747

SamMatthewsIsACommonName:
@sergey-akhalkov I'm having the same issue, have you been able to fix it?

sergey-akhalkov:
@SamMatthewsIsACommonName, hi! You could fix it by [installing CodePush plugin manually](https://github.com/Microsoft/react-native-code-push#plugin-installation-ios---manual) (look at pt.6).
I assume the issue will be fixed soon by React Native team so we wouldn't have to fix it manually.

SamMatthewsIsACommonName:
Hi @sergey-akhalkov I ended up using an earlier version of React Native and worked fine. Cheers!

anhtuank7c:
I have same problem with RN 0.40.0, RN-cli 2.0.1
```

/Users/anhtuank7c/Projects/reactnative/Clock/ios/Clock/AppDelegate.m:11:9: fatal error: 'CodePush.h' file not found
#import "CodePush.h"
        ^
1 error generated.


** BUILD FAILED **
```

2ducanhpham:
Thanks @anhtuank7c

fabriziogiordano:
I am having the same issue.

```
  "dependencies": {
    "react": "16.0.0-alpha.12",
    "react-native": "0.46.4",
    "react-native-code-push": "^4.1.0-beta"
  },
```

![screenshot 2017-07-21 10 29 06](https://user-images.githubusercontent.com/109390/28467995-a71af6e6-6dff-11e7-9930-44a3f6f8f3c8.png)

[...]

PR #665, opened 2017-01-09
[iOS][Build] Add `Headers` and `Copy Files` build phases

sergey-akhalkov:
Due to changes in new release of `react-native@0.40.0` there is `CodePush.h file not found` error occures.
To fix it I've added `CodePush.h` header file in `Headers` build phase and changed
`codePushHeaderImportStatement` variable value from `#import "CodePush.h"`
to `#import <CodePush/CodePush.h>` in `postlink.js` file.
Fix https://github.com/Microsoft/react-native-code-push/issues/662

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,10 +162,6 @@ And that's it! Isn't RNPM awesome? :)
 
     *Note: Alternatively, if you prefer, you can add the `-lz` flag to the `Other Linker Flags` field in the `Linking` section of the `Build Settings`.*
 
-6. Under the "Build Settings" tab of your project configuration, find the "Header Search Paths" section and edit the value.
-Add a new value, `$(SRCROOT)/../node_modules/react-native-code-push/ios` and select "recursive" in the dropdown.
-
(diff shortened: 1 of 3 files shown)


panarasi:
@sergey-akhalkov @BretJohnson Should this not have a phase to `Copy Headers` too ? I tried running it, the headers were included, but header was not copied over to `/Users/mYName/Library/Developer/Xcode/DerivedData/MYPROJ/Build/Products/Debug-iphonesimulator/include/CodePush/CodePush.a`

jy1989:
hi I've tried to get your modified sources and replace in my project. but the error still there. is it need to do some setup in my project?

sergey-akhalkov:
@panarasi, @jy1989 - hi!
I've updated PR and tested it with newly created React Native apps having react-native versions `v0.39.2` and `v0.41.0-rc.0` and now it should work, please verify.

grabbou:
Is this needed? I think the issue is that in new React Native link incorrectly fails to add path to header search path. When that gets fixed, all should work. No need to change the Codepush imports.

buptkang:
I test this PR with @silhouettes, and it works well.

Issue #672, opened 2017-01-13
getUpdateMetadata gives false package with DEBUG build


adbl:
Running a DEBUG build in simulator (and probably on device too), using this in AppDelegate.m:

```
 #ifdef DEBUG
  jsCodeLocation = [[RCTBundleURLProvider sharedSettings] jsBundleURLForBundleRoot:@"index.ios" fallbackResource:nil];
#else
  jsCodeLocation = [CodePush bundleURL];
#endif
```

which should make it impossible that a Code Push update gets loaded?

1. The first run this works as expected, `codePush.getUpdateMetadata()` returns `null`.
2. During this run, Code Push detected, downloaded and installed an update.
3. During next launch, `codePush.getUpdateMetadata()` now gives the package that was downloaded, however it clearly isn't running. It's running the dev bundle from packager. 

This had me extremely confused until I figured out that it wasn't actually running the update. It seems like `getUpdateMetadata` doesn't understand that an "installed" package may not be running, due to the CodePush bundle loader was not in use.

* react-native-code-push version: `1.16.1-beta`
* react-native version: `0.36.1`
* iOS/Android/Windows version: `iOS 10.2`
* Does this reproduce on a debug build or release build? deug
* Does this reproduce on a simulator, or only on a physical device? Simulator (both?)

sergey-akhalkov:
@adbl - hi!
`getUpdateMetadata` method returns info about latest installed package and supposes that local js code is in use, I don't sure if its behavior should be changed to determine js code location...
To make it easier to understand what js code location is in use we could output its location using this hack:
```
...
const { SourceCode } = require('NativeModules');
...
<Text style={styles.messages}>{SourceCode.scriptURL || ""}</Text>
...
```
Depending on `jsCodeLocation` variable value it prints different paths:
1. `jsCodeLocation = [CodePush bundleURL];` ![image](https://cloud.githubusercontent.com/assets/4928157/22059925/e148298a-dd7f-11e6-9d37-2d21868b3ca4.png)
2. `jsCodeLocation = [[RCTBundleURLProvider sharedSettings] jsBundleURLForBundleRoot:@"index.ios" fallbackResource:nil];` 
![image](https://cloud.githubusercontent.com/assets/4928157/22059987/1de4f328-dd80-11e6-8024-871fc31efb97.png)

As for debugging: 
``` 
#ifdef DEBUG
[...]

PR #673, opened 2017-01-18
CodePush: implement `isUsingReactNativeBundle` method.

sergey-akhalkov:
This helper method is used to determine if React Native dev bundle
currently is in use. In case `jsCodeLocation` variable value equals to
`[[RCTBundleURLProvider sharedSettings] jsBundleURLForBundleRoot:@"index.ios" fallbackResource:nil];`
React Native dev bundle is loaded each time app starts although CodePush updated is installed.

Relates to #672

### DIFF
--- a/CodePush.js
+++ b/CodePush.js
@@ -7,6 +7,7 @@ import log from "./logging";
 
 let NativeCodePush = require("react-native").NativeModules.CodePush;
 const PackageMixins = require("./package-mixins")(NativeCodePush);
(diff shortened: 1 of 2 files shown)


sergey-akhalkov:
Probably this method is surplus and PR should not be merged. It would be better to use workflow described in  https://github.com/Microsoft/react-native-code-push#ios to avoid the situation when it is hard to say what js bundle is in use.

panarasi:
We should close this PR, since it is better to mention this in docs.

PR #677, opened 2017-01-22
Update README.md

teodors:
Changed the end of line semicolon to a comma and added a bit more context. Since, this is an array there should be a comma not a semicolon. I did a blind copy/paste myself on initial setup and got a build fail.


silhouettes:
Looks good, thanks for the PR @teodors!

PR #670, opened 2017-01-12
Header changes required for 0.40

christopherdro:
These changes were required for me to properly build the project on the latest version of react native.
Related #665 #638

PR #678, opened 2017-01-24
Upgrade RN 0.40 Finalization

buptkang:
This PR finalizes the upgrade toward RN 0.40, which relates to previous PRs: #638, #665 and #670.

### DIFF
--- a/ios/CodePush/CodePush.h
+++ b/ios/CodePush/CodePush.h
@@ -1,4 +1,6 @@
 #if __has_include("RCTEventEmitter.h")
+#import "RCTEventEmitter.h"
+#elif __has_include(<React/RCTEventEmitter.h>)
 #import <React/RCTEventEmitter.h>
 #else
 #import "React/RCTEventEmitter.h"   // Required when used as a Pod in a Swift project
--- a/ios/CodePush/CodePush.h
+++ b/ios/CodePush/CodePush.h
@@ -1,7 +1,7 @@
-#if __has_include("RCTEventEmitter.h")
-#import "RCTEventEmitter.h"
-#elif __has_include(<React/RCTEventEmitter.h>)
+#if __has_include(<React/RCTEventEmitter.h>)
 #import <React/RCTEventEmitter.h>
+#elif __has_include("RCTEventEmitter.h")
+#import "RCTEventEmitter.h"
 #else
 #import "React/RCTEventEmitter.h"   // Required when used as a Pod in a Swift project
 #endif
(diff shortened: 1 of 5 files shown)


ios/CodePush/CodePush.h
```diff
@@ -1,4 +1,6 @@
 #if __has_include("RCTEventEmitter.h")
+#import "RCTEventEmitter.h"
+#elif __has_include(<React/RCTEventEmitter.h>)
```
silhouettes: Can we swap this with the statement above?

ios/CodePush/RCTConvert+CodePushInstallMode.m
```diff
@@ -1,5 +1,10 @@
 #import "CodePush.h"
+
+#if __has_include("RCTConvert.h")
```
silhouettes: Same here, please swap the ordering so that `<>` is the default

silhouettes: (and below)

BretJohnson: Should we swap the order, to do `#if __has_include(<React/RCTConvert.h>)` (that is, always check for the 0.40 case first)?  It shouldn't really matter, that that would make all the files consistent and seems slightly better.

mauron85: Hi guys. I'm just solving probably the [issue](https://github.com/mauron85/react-native-background-geolocation/issues/44) with ordering, like you did here. Can you please explain why ordering in if ... else statement is important? Just don't get it, but seems to be causing duplicate imports. Many thanks.

silhouettes: I think this was more of an aesthetic preference than anything else (put the case we expect to use first), so there wasn't necessarily a specific technical reason for it.

With the duplicate imports, I'm not sure without additional investigation - but I could hypothesize that both `"RCTConvert.h"` and `<React/RCTConvert.h>` exist in 0.40 and later, and you're meant to use the latter.

mauron85: > I could hypothesize that both "RCTConvert.h" and <React/RCTConvert.h> exist in 0.40

actually that is what i thought too, but `find node_modules/ -name RCTBridge*` finds only occurrence. Also even if there were two of them, I though that #if / #else is safe enough to not allow importing file more than once.

I saw this commit, where you're changing order and I though, you had same issue with duplicate imports, but if it's just cosmetic, then it remains mystery for me.

Anyway thank you for time and answer.

silhouettes: I don't think #if #else would protect you from duplicate imports. The #if only checks for existence - if it exists, it will try to import it. That's why my hypothesis is that the ordering might matter here - you should order by your preference.

PR #680, opened 2017-01-24
Bump package.json and update README.md for 1.17.0-beta

buptkang:


ThisIsMissEm:
Question, why is every release a beta release (marked by -beta)

silhouettes:
@ThisIsMissEm - tradition mostly. At the beginning of the project, we wanted to indicate that the project was still in its beta stages, hence the pre-release tag `-beta`. Of course, what we should have done is followed convention and numbered our releases with a major version of `0`, e.g. `0.17.0`.

ThisIsMissEm:
@silhouettes okay, interesting! So, it's not a beta semver release, but rather just the project is beta — that is, there is no 1.17.0 coming, there is only 1.17.0-beta

silhouettes:
right!

PR #684, opened 2017-01-31
Added React Native Windows dotNet46 support

abodalevsky:
Added support react-native-windows dotNet
Added example for react-native-windows UWP based
Added example for react-native-windows dotNet based

### Project structure:
_CodePush.Shared_ - shared code between UWP and dotNet
_CodePush_ - UWP specific code
_CodePush.Net46_ - dotNet specific code

For UWP solution it needs to be added the following projects:
* CodePush.Shared
* CodePush

For dotNet solution it needs to be added the following projects:
* CodePush.Shared
* CodePush.Net46

### Examples:
_Examples\CodePushDemoApp\windows\CodePushDemoApp.sln_ the solution contains both examples (UWP and dotNet).

### Notes
Example for ARM configuration has not been tested. Since there is no changes in UWP part of implementation, there is low risk of failure.

In this implementation we tried to reuse UWP library as much as possible. The following issues are relevant for both platforms:

1.  ZipFile.ExtractToDirectory is not reliable and throws exception if:
    * folder exists already
    * path is too long (> 250 chars)
2. Un-zipping is quite long operation. Does it make sense for _async_?
   * ` await UpdateUtils.UnzipBundleAsync(downloadFile.Path, unzippedFolder.Path);`


silhouettes:
Thank you so much for your contribution! As this is a large PR, please give us a week or so to work our way through it.

matthargett:
@silhouettes let me/us know if you'd like to do a video call to accelerate things. we did a few of these with @rozele for some of our larger contributions to React Native Windows and the ChakraCore team.

rozele:
@silhouettes I can attest to the great work @matthargett and his team have done to make react-native-windows work for Windows 7.

Examples/CodePushDemoApp/package.json
```diff
@@ -9,6 +9,7 @@
     "babel-preset-react-native-stage-0": "1.0.1",
     "react": "15.4.0",
     "react-native": "0.40.0",
-    "react-native-code-push": "file:../../"
+    "react-native-code-push": "file:../../",
+    "react-native-windows": "git://github.com/ReactWindows/react-native-windows.git"
   }
```
rozele: Since they are already locking to a specific version of react-native, you should also lock to a specific version of react-native-windows (these need to be kept in sync).

windows/CodePush/CodePushUtils.cs
```diff
@@ -2,13 +2,14 @@
 using System;
 using System.Diagnostics;
 using System.Threading.Tasks;
+using Windows.ApplicationModel;
 using Windows.Storage;
```
rozele: Not sure what this using belongs to, since no new type references were added.

windows/CodePush/FileUtils.cs
```diff
@@ -19,5 +19,21 @@ internal class FileUtils
                 await MergeFoldersAsync(sourceDirectory, nextTargetSubDir).ConfigureAwait(false);
             }
         }
+        internal async static Task ClearReactDevBundleCacheAsync()
```
rozele: add an empty line between methods.

windows/CodePush/UpdateManager.cs
```diff
@@ -98,6 +98,15 @@ internal class UpdateManager
                 await downloadFile.RenameAsync(expectedBundleFileName).AsTask().ConfigureAwait(false);
                 await downloadFile.MoveAsync(newUpdateFolder).AsTask().ConfigureAwait(false);
             }
+            catch (Exception e)
+            {
+				/*TODO: ZipFile.ExtractToDirectory is nor relible and throws exceptions if:
```
rozele: >nor relible [](start = 42, length = 11)

*not *reliable

rozele:
> using PCLStorage;

I know that we took this dependency for react-native-windows, but it's up to the CodePush team if they'll want to take this dependency here as well.  If you're not using PCLStorage on both sides (UWP and .NET 4.6), it may not be worth taking the dependency.

---
Refers to: windows/CodePush.Net46/FileUtils.cs:1 in d9c2028. [](commit_id = d9c20281b47cd723061d6bb8fa483f045f3be4b2, deletion_comment = False)

windows/CodePush.Net46/FileUtils.cs
```diff
@@ -0,0 +1,62 @@
...
+                await CopyFileAsync(sourceFile.Path, Path.Combine(target.Path, sourceFile.Name)).ConfigureAwait(false);
+            }
+
+            foreach (IFolder sourceDirectory in await source.GetFoldersAsync().ConfigureAwait(false))
+            {
+                var nextTargetSubDir = await target.CreateFolderAsync(sourceDirectory.Name, CreationCollisionOption.OpenIfExists).ConfigureAwait(false);
+                await MergeFoldersAsync(sourceDirectory, nextTargetSubDir).ConfigureAwait(false);
+            }
+        }
+        internal async static Task ClearReactDevBundleCacheAsync()
```
rozele: >internal [](start = 8, length = 8)

Empty newline between methods.

windows/CodePush.Net46/FileUtils.cs
```diff
@@ -0,0 +1,62 @@
...
+            {
+
+                var devBundleCacheFile = await FileSystem.Current.LocalStorage.GetFileAsync(CodePushConstants.ReactDevBundleCacheFileName).ConfigureAwait(false);
+
+                if (devBundleCacheFile != null)
+                {
+                    await devBundleCacheFile.DeleteAsync().ConfigureAwait(false);
+                }
+            }
+            catch (FileNotFoundException)
```
rozele: >catch [](start = 12, length = 5)

Is there a "TryGetFileAsync" or the like to prevent throwing an exception in a common case?  Is this method typically called when the bundle does not exist?

windows/CodePush.Net46/FileUtils.cs
```diff
@@ -0,0 +1,62 @@
...
+            {
+                //no files do nothing
+            }
+        }
+
+        internal static Task<long> GetBinaryResourcesModifiedTimeAsync(string fileName)
+        {
+            var pathToAssembly = CodePushUtils.GetAppFolder();
+            var pathToAssemblyResource = Path.Combine(pathToAssembly, CodePushConstants.AssetsBundlePrefix.Replace("ms-appx:///", String.Empty), fileName);
```
rozele: > CodePushConstants.AssetsBundlePrefix.Replace("ms-appx:///", String.Empty) [](start = 69, length = 74)

In react-native-windows, we eliminated the use of `ms-appx:///` in the context of react-native-wpf because the path prefix is controlled by the application.  Since ms-appx:/// is a UWP convention, it should not be used here.

windows/CodePush.Net46/UpdateManager.cs
```diff
@@ -0,0 +1,326 @@
+﻿using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+using System.IO;
+using System.IO.Compression;
+using System.Threading.Tasks;
+using System.Runtime.CompilerServices;
+using System;
+#if WINDOWS_UWP
+using Windows.Web.Http;
+#else
```
rozele: Since we're in the context of the .NET 4.6 project, do we really need these conditional compile directives?

windows/CodePush.Net46/UpdateManager.cs
```diff
@@ -0,0 +1,326 @@
...
+
+            await UpdateUtils.DownloadBundleAsync(downloadUrlString, downloadFile.Path, downloadProgress);
+
+            try
+            {
+                // Unzip the downloaded file and then delete the zip
+                var unzippedFolder = await GetUnzippedFolderAsync().ConfigureAwait(false);
+                /**
+                 * TODO:
+                 *  1) ZipFile.ExtractToDirectory is nor reliable and throws exception if:
```
rozele: >nor [](start = 53, length = 3)

not

windows/CodePush.Net46/UpdateManager.cs
```diff
@@ -0,0 +1,326 @@
...
+
+        internal async Task DownloadPackageAsync(JObject updatePackage, string expectedBundleFileName, Progress<HttpProgress> downloadProgress)
+        {
+            // Using its hash, get the folder where the new update will be saved 
+            var codePushFolder = await GetCodePushFolderAsync().ConfigureAwait(false);
+            var newUpdateHash = (string)updatePackage[CodePushConstants.PackageHashKey];
+            var newUpdateFolder = await GetPackageFolderAsync(newUpdateHash, false).ConfigureAwait(false);
+            if (newUpdateFolder != null)
+            {
+                // This removes any stale data in newPackageFolderPath that could have been left
```
rozele: >newPackageFolderPath [](start = 50, length = 20)

newUpdateFolder?

rozele: Maybe rename the variable to something like `existingUpdateFolder` to make it a bit more readable (then declare a new var below where you actually get the new reference to the folder)

---
In reply to: [99630777](https://github.com/Microsoft/react-native-code-push/pull/684#discussion_r99630777) [](ancestors = 99630777)

windows/CodePush.Net46/UpdateManager.cs
```diff
@@ -0,0 +1,326 @@
...
+            {
+                // Unzip the downloaded file and then delete the zip
+                var unzippedFolder = await GetUnzippedFolderAsync().ConfigureAwait(false);
+                /**
+                 * TODO:
+                 *  1) ZipFile.ExtractToDirectory is nor reliable and throws exception if:
+                 *      - folder exists already
+                 *      - path is too long (> 250 chars)
+                 *  
+                 *  2) Un-zipping is quite long operation. Does it make sense for async?
```
rozele: >) Un-zipping is quite long operation [](start = 21, length = 36)

It depends on where this is originally called from.  I would assume that this work is already being performed on a background thread, so there would be no reason to create yet another task to continue this work.

windows/CodePush.Net46/UpdateManager.cs
```diff
@@ -0,0 +1,326 @@
...
+            var relativeBundlePath = (string)currentPackage[CodePushConstants.RelativeBundlePathKey];
+
+            return relativeBundlePath == null
+                ? await packageFolder.GetFileAsync(bundleFileName).ConfigureAwait(false)
+                : await packageFolder.GetFileAsync(relativeBundlePath).ConfigureAwait(false);
+        }
+
+        internal async Task<string> GetCurrentPackageHashAsync()
+        {
+            JObject info = await GetCurrentPackageInfoAsync().ConfigureAwait(false);
```
rozele: >JObject [](start = 12, length = 7)

Kind of a nitpick, but there seems to be an inconsistent use of explicit and implicit variable types.  On the whole, does the code use mostly `var` or is there already a healthy mix? The C# coding convention says "Use implicit typing for local variables when the type of the variable is obvious from the right side of the assignment, or when the precise type is not important."  I rarely find a case where type information is really important.

windows/CodePush.Net46/UpdateUtils.cs
```diff
@@ -0,0 +1,65 @@
+﻿using CodePush.Net46.Adapters.Http;
+using Newtonsoft.Json.Linq;
+using PCLStorage;
+using System;
+using System.IO;
+using System.Net;
+using System.Threading.Tasks;
+
+
```
rozele: Delete extra newline.

windows/CodePush.Net46/UpdateUtils.cs
```diff
@@ -0,0 +1,65 @@
...
+        internal async static Task DownloadBundleAsync(string url, string fileName, IProgress<HttpProgress> downloadProgress)
+        {
+            var uri = new Uri(url);
+            var client = new WebClient();
+            client.DownloadProgressChanged += (s, e) =>
+            {
+                downloadProgress.Report(new HttpProgress
+                {
+                    BytesReceived = (UInt64)e.BytesReceived,           //conversion ulong to long is safe
```
rozele: >UInt64 [](start = 37, length = 6)

just use `ulong` here?

rozele: >ulong to long [](start = 84, length = 13)

*long to ulong

windows/CodePush.Shared/CodePushConstants.cs
```diff
@@ -27,5 +25,11 @@ internal class CodePushConstants
         internal const string RelativeBundlePathKey = "bundlePath";
         internal const string StatusFileName = "codepush.json";
         internal const string UnzippedFolderName = "unzipped";
+#if WINDOWS_UWP
+        internal const string AssetsBundlePrefix = "ms-appx:///ReactAssets/";
+        internal const string FileBundlePrefix = "ms-appdata:///local";
+#else
```
rozele: no file bundle prefix needed for WPF?

windows/CodePush.Shared/CodePushNativeModule.cs
```diff
@@ -240,17 +244,22 @@ public void notifyApplicationReady(IPromise promise)
...
         {
             // #1) Get the private ReactInstanceManager, which is what includes
             //     the logic to reload the current React context.
             FieldInfo info = typeof(ReactPage)
                 .GetField("_reactInstanceManager", BindingFlags.NonPublic | BindingFlags.Instance);
-
+#if WINDOWS_UWP
             var reactInstanceManager = (ReactInstanceManager)typeof(ReactPage)
                 .GetField("_reactInstanceManager", BindingFlags.NonPublic | BindingFlags.Instance)
                 .GetValue(_codePush.MainPage);
```
rozele: This is more of a question for the CodePush team, but should we just make the fields you need here part of the public API on react-native-windows?

windows/CodePushReactPackage.cs
```diff
@@ -1,25 +1,18 @@
...
-using Windows.Web.Http;
+
 
 namespace CodePush.ReactNative
 {
     public sealed class CodePushReactPackage : IReactPackage
     {
         private static CodePushReactPackage CurrentInstance;
-        private static bool NeedToReportRollback = false;
-
```
rozele: This is suspicious, is this some missing feature, or was it just totally dead code?

windows/CodePush.Shared/SettingsManager.cs
```diff
@@ -1,13 +1,21 @@
...
+#endif
 
 namespace CodePush.ReactNative
 {
     internal class SettingsManager
     {
+#if WINDOWS_UWP
         private static ApplicationDataContainer Settings = ApplicationData.Current.LocalSettings.CreateContainer(CodePushConstants.CodePushPreferences, ApplicationDataCreateDisposition.Always);
+#else
+        private static ApplicationDataContainer Settings = new ApplicationDataContainer();
```
rozele: >new ApplicationDataContainer(); [](start = 59, length = 31)

Are these settings in-memory only then?  I think ApplicationData.Current.LocalSettings is persisted across runs.

windows/CodePush.Shared/CodePushConstants.cs
abodalevsky: Since fix in [PR#950](https://github.com/ReactWindows/react-native-windows/pull/950) **ReactNative.Net46** uses absolute path, without prefix. Therefore all prefixes were removed for Net46 solution.

windows/CodePush.Net46/FileUtils.cs
abodalevsky: Unfortunately there is no _Try....._ functions. Replaced with preliminary verification if file exists. If file doesn't exists (which is most common case) it simply return no any exception will be thrown.

windows/CodePush.Net46/FileUtils.cs
abodalevsky: Correct, actually, it has been already fixed in `CodePushConstants.AssetsBundlePrefix` thus this code replaces nothing. Removed.

windows/CodePush.Net46/UpdateManager.cs
abodalevsky: Seems like _newUpdateFolder_ is correct name. Fixed.
This implementation (including comments :) ) replicates UWP implementation. I've tried to follow the idea not to change existing code as much as possible.  The proposed change make sense but looks like it should be done in both versions (UWP and NET46) that is I believe can be done as separate PR.

windows/CodePush.Net46/UpdateManager.cs
abodalevsky: Correct, fixed to the `var`

Examples/CodePushDemoApp/package.json
abodalevsky: Yep, totally agree in general. But the latest available version of react-native-windows is 0.39.0 This implementation requires more modern version.

windows/CodePush.Shared/SettingsManager.cs
abodalevsky: You are right absolutely! Thanks for this catch.
Fixed in 03af9fcf29e95a00dc8dbdee2214d0e4e2ab2bda

abodalevsky:
_Regarding PCLStorage_. I use it for Net46 project only. In fact this doesn't add additional dependency because it is already exists in ReactNative.Net46 project.

In general, probably, it make sense to use PCLStorage for UWP implementation as well. A benefit will be more code shared between UWP & NET implementations. As I mentioned in comments above I tried not to modify existent implementation UWP, as this is not the subject of this PR.

windows/CodePushReactPackage.cs
matthargett: It was part of @geof90 's original commit (according to git blame). The value is set to true, but never read from. Looks like it was transliterated in to C# incorrectly, maybe? We can file and issue and do a separate PR to fix -- we specifically have user stories in our project for update failure modes that will likely force us to fill missing gaps on that front.

windows/CodePush.Shared/CodePushNativeModule.cs
silhouettes: @rozele That would be ideal and would help us to avoid being broken by changes, but it's up to you if that's your priority. We do the same kind of reflection on the other platforms, although we would like to at some point in the distant future submit PR's so that we can get rid of it.

.gitignore
```diff
@@ -150,3 +150,37 @@ captures/
 
 # Remove after this framework is published on NPM
 code-push-plugin-testing-framework/node_modules
+
```
silhouettes: Can you mirror these changes to the `.npmignore` file, and if possible, could you also leave a comment at the top of this file informing future contributors that they need to do the same?

I'm also assuming there's nothing you want to exclude from the npm package that isn't already listed here, but please add them there if so!

windows/CodePushReactPackage.cs
silhouettes: @matthargett I suspect the purpose of this is to post to the CodePush server after a rollback happens, so that users can see a record of the number of rollbacks in the CodePush CLI. If you'd like to file an issue and do a separate PR I am fine with that!

silhouettes:
@rozele Do you think your and @matthargett's reviews are sufficient to merge this?

.gitignore
kevinvangelder: @silhouettes Updated. (Alex is offline for a few hours)

rozele:
@silhouettes I'll take another pass and then approve.

package.json
```diff
@@ -28,12 +28,14 @@
...
-    "gulp-typescript": "2.12.2",
     "gulp-insert": "latest",
     "gulp-tslint": "latest",
+    "gulp-typescript": "2.12.2",
     "mkdirp": "latest",
     "q": "^1.4.1",
-    "run-sequence": "latest"
+    "run-sequence": "latest",
+    "tslint": "^4.3.1",
+    "typescript": "^2.1.5"
```
rozele: Any specific reason why these were added?

windows/CodePush/UpdateManager.cs
```diff
@@ -98,6 +98,15 @@ internal class UpdateManager
...
             }
+            catch (Exception e)
+            {
+				/*TODO: ZipFile.ExtractToDirectory is not reliable and throws exceptions if:
+				 	- folder exists already
+					- path is too long
+					it needs to be handled			 
+				 */
+                throw e;
+            }
```
rozele: This will screw up the stack trace of the exception (the point where the exception is thrown is recalculated from here).   Since this has no net affect on the behavior other, please remove this catch block.

windows/CodePush.Net46/UpdateManager.cs
```diff
@@ -0,0 +1,317 @@
...
+            {
+                // Downloaded file is not a zip, assume it is a jsbundle
+                await downloadFile.RenameAsync(expectedBundleFileName).ConfigureAwait(false);
+                await downloadFile.MoveAsync(newUpdateFolder.Path, NameCollisionOption.ReplaceExisting).ConfigureAwait(false);
+            }
+            catch (Exception e)
+            {
+                throw e;    //left here for debug purposes to see the reason of ZipFile failure.
+            }
+
```
rozele: Another case of a catch block with no purpose.

rozele:
@rihua@microsoft.com one architectural comment about the shared code project.  If we were shipping code push as a compiled binary, it wouldn't matter so much, but now users have to add two projects to their react-native-windows solutions to see all code from the code push module.  Obviously the shared code approach saves a bunch of duplicate code for WPF.  The only other alternative would be to request that a copy of all code be made instead of using a shared project.  Just thought I should point this out, as it could impact perceived debuggability for code push users.

---
In reply to: [278792526](https://github.com/Microsoft/react-native-code-push/pull/684#issuecomment-278792526) [](ancestors = 278792526)

rozele:
Other than the final comments I've made, and the general comment about shared projects, this looks good.

---
In reply to: [278795588](https://github.com/Microsoft/react-native-code-push/pull/684#issuecomment-278795588) [](ancestors = 278795588,278792526)

matthargett:
@rozele we can follow up with a separate PR that creates a NuGet package for CodePush, if you like. Since CodePush isn't like React Native Windows where people may want to extend the core source project with new files, it should probably be a NuGet package regardless of this PR for the best DX.

package.json
abodalevsky: Nothing special, just wanted to avoid npm _UNMET PEER DEPENDENCY_ warnings:

>npm WARN gulp-tslint@7.1.0 requires a peer of tslint@>=4.0.0-dev but none was installed.
>npm WARN tslint@4.4.2 requires a peer of typescript@>=2.0.0 but none was installed

.gitignore
abodalevsky: Thank you @kevinvangelder! I also added UnitTest project to `.npmignore`

silhouettes:
I will merge this PR on Monday unless anyone still wants to make changes.

Thanks everyone for the great work and the very informative comments :)

maxim-pop:
Hi @abodalevsky ,
The Demo App contains a .pfx certificate file (/CodePushDemoApp/windows/CodePushDemoApp/CodePushDemoApp_TemporaryKey.pfx). 
Is it possible to remove it from repo? We have compliance issues with storing secrets in code.

abodalevsky:
Hi @maxim-pop 

VisualStudio has generated this certificate by default within the project. In general it should be used for signing the assembly for deployment purposes. But I don't see in code that it is used. Therefore I may assume that for this example the *_TemporaryKey.pfx may be safely removed. 

I have created [PR #741](https://github.com/Microsoft/react-native-code-push/pull/741), please take a look.

maxim-pop:
Thanks!

PR #687, opened 2017-02-03
Test automation

buptkang:
This script can be used:

1. Compatibility between react-native and react-native-code-push.
2. Reproduction of PR or issues that needs certain additional libraries or resources.


Examples/AutomateTest/README.md
```diff
@@ -0,0 +1,12 @@
+The goal of this script is to automate the testing of react-native, react-native-code-push and their internal compatibility. Run **./run.sh** to create an app and test it.
```
silhouettes: Can we have this simply as a comment at the top of the script? I don't think we need a separate README

Examples/AutomateTest/run.sh
```diff
@@ -0,0 +1,80 @@
...
+echo 'react native link to react native code push'
+react-native link react-native-code-push
+
+rm index.android.js
+rm index.ios.js
+cp ../skeleton/*js .
+mkdir images
+cp ../skeleton/images/* images
+
+if [ $mobile_env == 'ios' ]; then
```
silhouettes: Can we re-name this script to `generate-app.sh`, and remove the `mobile_env` variable and the `run-ios` and `run-android`? I don't think that part is necessary, and we typically want to test both anyway. In future, we can use `generate-app` however we like, for example as one of a series of steps.

Examples/AutomateTest/run.sh
```diff
@@ -0,0 +1,80 @@
...
+echo 'Automate synchronization testing between React Native and React Native Code Push';
+echo
+ 
+# Please make sure you installed react native in your mac machine
+
+echo '************************ Configuration ***********************************';
+
+####################  Code Push Config  #########################W#########################
+
+code_push_version=`code-push --version`;
```
silhouettes: This should be called `code_push_cli_version` for clarity

silhouettes: Also why do we need this variable? It doesn't seem to be used anywhere else

Examples/AutomateTest/README.md
silhouettes: You might be able to remove the gitignore as well, assuming that the test app can be generated in any directory

Examples/AutomateTest/README.md
```diff
@@ -0,0 +1,12 @@
+The goal of this script is to automate the testing of react-native, react-native-code-push and their internal compatibility. Run **./run.sh** to create an app and test it. 
+
+You can configure below variables in the bash.
```
silhouettes: I don't think these make sense as environment variables - I don't think of them as things that need to be persisted across sessions, you really only want to set them once for the execution of the script. Instead, I recommend doing an input prompt that can be overriden by command-line arguments.

In other words:

```
prompt> ./generate-app.sh
What react-native version would you like to use (default: latest)? <Hit return key>
What react-native-code-push version would you like to use (default: latest)? 1.17.0-beta

Using:
react-native@0.41.0
react-native-code-push@1.17.0-beta
```

Can be overriden like this:

```
prompt> ./generate-app.sh latest 1.17.0-beta

Using:
react-native@0.41.0
react-native-code-push@1.17.0-beta
```

Examples/AutomateTest/run.sh
```diff
@@ -0,0 +1,80 @@
...
+echo 'Code Push version: ' + ${code_push_version};
+
+####################  React Native Code Push Config  ######################################
+
+react_native_code_push_version='';  # manual label yo, yo, yo.
```
silhouettes: What does 'yo, yo, yo' here and below mean?

Examples/AutomateTest/run.sh
silhouettes: Also, we can probably put it directly in the `Examples` directory, if we can get it down to a single script file

Examples/AutomateTest/run.sh
```diff
@@ -0,0 +1,80 @@
...
+npm install --save react-native-code-push@$react_native_code_push_version
+
+echo 'react native link to react native code push'
+react-native link react-native-code-push
+
+rm index.android.js
+rm index.ios.js
+cp ../skeleton/*js .
+mkdir images
+cp ../skeleton/images/* images
```
silhouettes: Can we also automate the steps for setting up debug mode testing of CodePush?

buptkang:
@silhouettes According to your comments, I have updated the script.

PR #694, opened 2017-02-07
[docs][android] readme: improved “Releasing Updates” and “Multi-Deployment Testing” c…

sergey-akhalkov:
…hapters

CodePush updates should be tested in modes other than Debug due to in Debug mode React Native app always downloads JS bundle generated by packager, so JS bundle downloaded by CodePush does not apply.

Relates https://github.com/Microsoft/react-native-code-push/issues/614 and https://github.com/Microsoft/react-native-code-push/issues/682


sergey-akhalkov:
@buptkang, thank you for the correction, I've applied the fix

PR #702, opened 2017-02-14
[docs] readme: improved “Dynamic Deployment Assignment” chapter

sergey-akhalkov:
Deployment key specified by `codePush.sync({ deploymentKey: key })` will not persist across app restarts (user kills app process and launch it again) and will be taken from native-side on next app launch. In case app restarts programmatically (while using IMMEDIATE installation mode) deployment key will persists.

### DIFF
--- a/README.md
+++ b/README.md
@@ -730,6 +730,8 @@ In order to achieve this kind of workflow, all you need to do is specify the dep
 codePush.sync({ deploymentKey: userProfile.CODEPUSH_KEY });
 ```
 
+*NOTE: Deployment key specified by `codePush.sync({ deploymentKey: key })` will not persist across app restarts (user kills app process and launch it again) and will be taken from native-side on next app launch. In case app restarts programmatically (while using IMMEDIATE installation mode) deployment key will persists.*
+
 With that change in place, now it's just a matter of choosing how your app determines the right deployment key for the current user. In practice, there are typically two solutions for this:
 
 1. Expose a user-visible mechanism for changing deployments at any time. For example, your settings page could have a toggle for enabling "beta" access. This model works well if you're not concerned with the privacy of your pre-production updates, and you have power users that may want to opt-in to earlier (and potentially buggy) updates at their own will (kind of like Chrome channels). However, this solution puts the decision in the hands of your users, which doesn't help you perform A/B tests transparently.


README.md
```diff
@@ -730,6 +730,8 @@ In order to achieve this kind of workflow, all you need to do is specify the dep
 codePush.sync({ deploymentKey: userProfile.CODEPUSH_KEY });
 ```
 
+*NOTE: Deployment key specified by `codePush.sync({ deploymentKey: key })` will not persist across app restarts (user kills app process and launch it again) and will be taken from native-side on next app launch. In case app restarts programmatically (while using IMMEDIATE installation mode) deployment key will persists.*
```
silhouettes: I'm not sure what is meant by persisting after the app restarts programatically. As far as I understood, the key only persists for the duration of the call, regardless of when the call is made, so:

```
codePush.sync({ deploymentKey: key })   // This call uses the specified key, regardless of when it is called
    .then(() => {
        codePush.sync();    // This call uses the key specified natively, regardless of when it is called
    });
```

Have I gotten something confused here? If I haven't gotten it muddled up, then maybe this is what you mean?

```
*NOTE: The deployment key specified via `codePush.sync({ deploymentKey: key })` does not persist beyond the given call, and needs to be specified as an option each time it is used.*
```

I'm not sure that using `IMMEDIATE` install mode or not makes any difference - are you saying that the value set to the variable `key` is lost when the app restarts, but not when it restarts programmatically? Because that's not what I'd expect.

sergey-akhalkov: >I'm not sure what is meant by persisting after the app restarts programatically.

`IMMEDIATE` installation mode uses `RestartManager.restartApp()` method to restart app programatically (user does not kill the app process and launch it again)

>I'm not sure that using IMMEDIATE install mode or not makes any difference - are you saying that the value set to the variable key is lost when the app restarts, but not when it restarts programmatically? Because that's not what I'd expect.

I've created demo app and could confirm, that the value set to variable key is not probably lost when app restart programmatically. Here is the code I've used to reproduce the issue:
```javascript
  syncProduction() {
    CodePush.sync(
      { deploymentKey: "iDKl3x098ODmeItIQEyEMjV3M7EjVyM2ZvIuz" } // Production deployment key
    )
  }

  syncStaging() {
    CodePush.sync(); // Staging deployment key should be taken from native-side
  }

  restartApp() {
    CodePush.restartApp(); // both "CodePush.restartApp()" and "IMMEDIATE" installation mode are using "RestartManager.restartApp()" method to restart app programatically (user does not kill the app process and launch it again)
  }
```
Reproduction steps:
1. call `syncProduction()` method
2. call `restartApp()` method
3. call `syncStaging()` method

**Actual**: CodePush prints "App up to date." due to probably using Production deployment key
**Expected**: CodePush prints "Update installed and will be applied on restart." due to using staging deployment key

I believe I should fix this confusing behavior instead of updating the docs. I suppose that the expected behavior is `*NOTE: The deployment key specified via `codePush.sync({ deploymentKey: key })` does not persist beyond the given call, and needs to be specified as an option each time it is used.*` as you've described above.

sergey-akhalkov:
Closing this due to https://github.com/Microsoft/react-native-code-push/issues/762 has been created.

PR #704, opened 2017-02-14
Automate test

buptkang:
Automation test script to verify each version of react native.


silhouettes:
Will look at this tomorrow

Examples/CodePushDemoApp/demo.js
```diff
@@ -16,7 +16,7 @@ import CodePush from "react-native-code-push";
  * different check frequency, such as ON_APP_START, for a 'hands-off' approach where CodePush.sync() does not
  * need to be explicitly called. All options of CodePush.sync() are also available in this decorator.
  */
-@CodePush({ checkFrequency: CodePush.CheckFrequency.MANUAL })
+// @CodePush({ checkFrequency: CodePush.CheckFrequency.MANUAL })
```
silhouettes: Could you remove this line altogether?

Examples/CodePushDemoApp/demo.js
```diff
@@ -16,7 +16,7 @@ import CodePush from "react-native-code-push";
  * different check frequency, such as ON_APP_START, for a 'hands-off' approach where CodePush.sync() does not
  * need to be explicitly called. All options of CodePush.sync() are also available in this decorator.
  */
```
silhouettes: Can you move this comment down to just above line 159, where it is used? And can you add the `checkFrequency` argument back in so that the behavior is equivalent to what it was before?

Examples/generate-app.sh
```diff
@@ -0,0 +1,98 @@
...
+
+echo 'react native link to react native code push'
+react-native link react-native-code-push
+
+rm index.android.js
+rm index.ios.js
+cp ../CodePushDemoApp/*js .
+mkdir images
+cp ../CodePushDemoApp/images/* images
+
```
silhouettes: I think the `run-ios` and `run-android` commands below can be removed. For a good separation of concerns, I feel like it would be better for this script to only be responsible for generating the app. If desired in the future, other scripts can be responsible for running the app in different ways (or running integration tests on it), and they can be chained together with this one in different combinations.

Examples/generate-app.sh
```diff
@@ -0,0 +1,98 @@
...
+echo '********************* Running Android ***************************************';
+
+read -p "you need to open the android simulator manually before continue running this script. Enter [Enter] to continue" 
+
+react-native run-android
+
+# cd ..
+# rm -rf testapp
+# exit;
```
silhouettes: Please remove these three lines. In general don't keep commented code around unless there's a reason for it.

Examples/generate-app.sh
```diff
@@ -0,0 +1,98 @@
...
+# 1. `code_push_version`
+# 2. `react_native_code_push_version`
+# 3. `react_native_version`
+# 4. `mobile_env`
+
+# If you use other libs or packages, and fail to run this test, welcome to modify this script 
+# so that we can reproduce any issues that you have.
+
+
+echo 'Automate synchronization testing between React Native and React Native Code Push';
```
silhouettes: I think this line would be better modified to describe the script as a CodePush + RN sample app generation script

Examples/generate-app.sh
```diff
@@ -0,0 +1,98 @@
...
+
+# The goal of this script is to automate the testing of react-native, react-native-code-push and 
+# their internal compatibility. Run **./run.sh** to create an app and test it. 
+
+# You can configure below variables in the bash. 
+
+# 1. `code_push_version`
+# 2. `react_native_code_push_version`
+# 3. `react_native_version`
+# 4. `mobile_env`
```
silhouettes: These bash variables don't apply anymore, and can be removed

Examples/generate-app.sh
```diff
@@ -0,0 +1,98 @@
...
+
+# Copyright (c) 2015-present, Microsoft Inc.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree. An additional grant
+# of patent rights can be found in the PATENTS file in the same directory.
+
+# The goal of this script is to automate the testing of react-native, react-native-code-push and 
+# their internal compatibility. Run **./run.sh** to create an app and test it.
```
silhouettes: I think this would be better described as a script that can generate a sample app for a specific version of React Native and CodePush. Also, it is no longer named `run.sh`

Examples/generate-app.sh
```diff
@@ -0,0 +1,98 @@
...
+rm -rf testapp_rn
+ 
+# Please make sure you installed react native in your mac machine
+
+echo '************************ Configuration ***********************************';
+
+####################  Code Push Config  #########################W#########################
+
+code_push_cli_version=`code-push --version`;
+echo 'Code Push CLI version: ' + ${code_push_cli_version};
```
silhouettes: Do we need this section? This variable doesn't seem to be used anywhere.

Examples/generate-app.sh
```diff
@@ -0,0 +1,98 @@
...
+npm install --save react-native-code-push@$react_native_code_push_version
+
+echo 'react native link to react native code push'
+react-native link react-native-code-push
+
+rm index.android.js
+rm index.ios.js
+cp ../CodePushDemoApp/*js .
+mkdir images
+cp ../CodePushDemoApp/images/* images
```
silhouettes: Can we also automate the steps for setting up debug mode testing of CodePush?

Examples/generate-app.sh
buptkang: I accept all changes, just have a look at it again.

silhouettes:
Hey Bo, there are still a few comments above that haven't been addressed, could you maybe double check again?

buptkang:
@silhouettes Fixed the comment in the script.

Examples/CodePushDemoApp/demo.js
```diff
@@ -156,4 +150,13 @@ const styles = StyleSheet.create({
...
 
-AppRegistry.registerComponent("CodePushDemoApp", () => CodePushDemoApp);
+/**
+ * Configured with a MANUAL check frequency for easy testing. For production apps, it is recommended to configure a
+ * different check frequency, such as ON_APP_START, for a 'hands-off' approach where CodePush.sync() does not
+ * need to be explicitly called. All options of CodePush.sync() are also available in this decorator.
+ */
+let codePushOptions = { checkFrequency: CodePush.CheckFrequency.MANUAL };
+
+CodePushDemoApp = new CodePush(codePushOptions)(CodePushDemoApp);
```
silhouettes: There is no `new` here

Examples/generate-app.sh
```diff
@@ -0,0 +1,73 @@
...
+
+sed -i.bak '162s/AppRegistry.registerComponent("CodePushDemoApp", () => CodePushDemoApp);/AppRegistry.registerComponent("testapp_rn", () => CodePushDemoApp);/' demo.js
+rm -f demo.js.bak
+
+perl -i -p0e 's/#ifdef DEBUG.*?#endif/jsCodeLocation = [CodePush bundleURL];/s' ios/testapp_rn/AppDelegate.m
```
silhouettes: Why use `perl` here if we are using `sed` above and below?

Examples/generate-app.sh
```diff
@@ -0,0 +1,73 @@
...
+react-native link react-native-code-push
+
+rm index.android.js
+rm index.ios.js
+cp ../CodePushDemoApp/*js .
+mkdir images
+cp ../CodePushDemoApp/images/* images
+
+sed -i.bak '162s/AppRegistry.registerComponent("CodePushDemoApp", () => CodePushDemoApp);/AppRegistry.registerComponent("testapp_rn", () => CodePushDemoApp);/' demo.js
+rm -f demo.js.bak
```
silhouettes: If we're removing the `.bak` files immediately, we probably shouldn't make them to begin with

Examples/generate-app.sh
buptkang: Talked and there is a good reason for it

buptkang: Talked and there is a good reason for it

silhouettes:
@buptkang - No rush at all, but I just realized that this script doesn't allow you to install `react-native-code-push` from a local directory, which is important for testing. No need to do it if you have other work, but if you ever have spare time it might be something worth adding.

sergey-akhalkov:
Thank you @buptkang and @silhouettes, it's very helpful script!

>allow you to install react-native-code-push from a local directory

Please let me know if I could help you with this improvement.

silhouettes:
If you feel that it will make you more productive, feel free to go ahead!

PR #710, opened 2017-02-20
Update Android SDK to 23.0.3

### LINKED ISSUE #709: Update Android SDK version to 23.0.3
The app Gradle file currently references Android SDK 23.0.1 while it should use the latest 23.0.X which is 23.0.3

tlvenn:
Fixes #709

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -2,7 +2,7 @@ apply plugin: "com.android.library"
 
 android {
     compileSdkVersion 23
-    buildToolsVersion "23.0.1"
+    buildToolsVersion "23.0.3"
 
     defaultConfig {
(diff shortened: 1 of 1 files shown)


tlvenn:
Just to give some perspective on how this can affect end users, there are multiple CI which will only ship the latest revision for a given Android SDK major version. For example, Bitrise is including 23.0.3 but not 23.0.2 or 23.0.1

Right now because of this, our builds are failing unless we manually pull the deprecated 23.0.1. By upgrading code push to the latest revision (23.0.3), it fix this issue all together.

tlvenn:
@buptkang any chance to merge that soon ? Thanks !

PR #712, opened 2017-02-21
Correctly outputs CodePushHash.json to custom folder

Igor1201:
Remove `assetsDir` variable, as `generateBundledResourcesHash.js` should output `CodePushHash.json` to `jsBundleDir${targetName}` folder, and not the hardcoded `$buildDir/intermediates/assets/${targetPath}` one.

### DIFF
--- a/android/codepush.gradle
+++ b/android/codepush.gradle
@@ -36,9 +36,8 @@ gradle.projectsEvaluated {
                     "${buildTypeName}"
 
             def jsBundleDirConfigName = "jsBundleDir${targetName}"
-            def assetsDir = "$buildDir/intermediates/assets/${targetPath}"
             def jsBundleDir = elvisFile(config."$jsBundleDirConfigName") ?:
-                    file(assetsDir)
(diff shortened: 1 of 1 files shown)


silhouettes:
@Igor1201: Apologies if my understanding is weak here - we can override where it is written to, but won't it still be [read from the same location](https://github.com/Microsoft/react-native-code-push/blob/3dce59f36e0f780f1773342e20c56fe870ba019a/android/app/src/main/java/com/microsoft/codepush/react/CodePushUpdateUtils.java#L110)?

Igor1201:
@silhouettes yeah, once the app is built, it will be read from the same location (the assets folder inside the app), but my change was to set the folder in which the `CodePushHash.json` is generated before the app itself is built.

In my knowledge, we should fallback both paths to their default locations (`build/intermediates/assets` and `build/intermediates/resources`) but they should be overridable by the react config vars (`resourcesDirDebug`, `resourcesDirRelease`, `jsBundleDirDebug` and `jsBundleDirRelease`), exactly the way we're handling `resourcesDir` right now.

PR #725, opened 2017-02-28
Update Android SDK to version 25

### LINKED ISSUE #724: CodePush doesn't build with latest android SDK, version 25
The build fails with following error massage:
```
The SDK Build Tools revision (23.0.1) is too low for project ':react-native-code-push'. Minimum required is 25.0.0
```

* react-native-code-push version: `1.17.0`
* react-native version: `0.41.2`
* Android version: `25`
* Does this reproduce on a debug build or release build? `both`
* Does this reproduce on a simulator, or only on a physical device? `both`
* Repro steps and/or stack trace: `just trying to compile`

rayvafa:
The app Gradle file currently references Android SDK 23 which is two version old. The latest SDK version is 25.

This fixes #724.

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -1,8 +1,8 @@
 apply plugin: "com.android.library"
 
 android {
-    compileSdkVersion 23
-    buildToolsVersion "23.0.3"
+    compileSdkVersion 25
+    buildToolsVersion "25.0.2"
 
     defaultConfig {
         minSdkVersion 16


MattyK14:
For future reference, react native currently only supports API 16 - 23

sergey-akhalkov:
@MattyK14, thank you for the reference, I've also created PR https://github.com/Microsoft/react-native-code-push/pull/747 that fixes this inconsistency.

ryankask:
I found this gist on the RN Facebook group that should help users that are trying to use newer versions of the SDK: https://gist.github.com/AndrewJack/10dc22708582b009fb74f07904f3bfb1

Basically, you can dynamically change `compileSdkVersion` and `buildToolsVersion` for subprojects.

PR #751, opened 2017-03-10
Update README.md

timscott:
When I fail to supply a CODEPUSH_KEY variable for debug, the build fails. This pull fixes the example code. I also added a comment letting developers know to expect errors from CodePush in debug mode. (This behavior is surprising, but I believe is by design.)


timscott:
Sure.

Issue #577, opened 2016-10-21
[Android] CodePushHash.json is not JSON


insaindesign:
`CodePushHash.json` is just storing a string, and not JSON, so the file type should **NOT** be JSON. This breaks flow type checking, and generally just seems confusing.

The file is read as a string from https://github.com/Microsoft/react-native-code-push/blob/3dce59f36e0f780f1773342e20c56fe870ba019a/android/app/src/main/java/com/microsoft/codepush/react/CodePushUpdateUtils.java#L110 and written to as a crypto hex string in https://github.com/Microsoft/react-native-code-push/blob/bb04535fbef4cb580da4c29ae628452cc0774fc8/scripts/generateBundledResourcesHash.js#L74. At no point is it encoded or needing to be decoded as JSON.

panarasi:
@insaindesign I think this is a good point. However, if we change the name now, it would break existing apps. Is there a way you could ignore it in the flow type checking ?

hawkrives:
@panarasi In my project, I have flow configured to ignore all .json files, which it does – it will error on other JSON files if I don't – but it still errors on the `CodePushHash.json`.

We discovered that adding `.*/CodePushHash.json`, with the leading `.*/`, will force flow to ignore the file and not error.

panarasi:
@sergey-akhalkov Can we investigate this ? We would need a backward compatible change.

sergey-akhalkov:
Closing this due to #763 has been merged, please feel free to reopen if needed.

PR #763, opened 2017-03-22
flow type checking fix: use "CodePushHash" file name instead of "Code…

sergey-akhalkov:
…PushHash.json"

`CodePushHash.json` file name breaks flow type checking.
To fix the issue we need to delete `CodePushHash.json` file and
use `CodePushHash` file name instead to store the hash value.
Relates to https://github.com/Microsoft/react-native-code-push/issues/577


silhouettes:
I'll merge this for you as I'd like to include this in today's release.

Issue #366, opened 2016-06-07
Idea: codePush.InstallMode.ON_NEXT_SUSPEND


nevir:
Another mode, similar to `codePush.InstallMode.ON_NEXT_RESUME`, except that it applies the update in the background (after some delay).  

This would be a potential speed up when resuming the app (e.g. when brought back from suspension) - I'm not sure if the added complexity is worth it (for `sync()`), though 

---

On the iOS side, this would require some extra work via a background task (and some API exposed for it), I think.  Similar setup for (newer) Android, too, I think.

nevir:
(or, it might be worth waiting for react native to provide an abstraction for background tasks)

lostintangent:
@nevir That's an awesome idea :) We'll definitely have to do an investigation into what this would look like. In lieu on RN improving their background task story, we also have quite a few users embedding RN into existing native apps, and have asked to be able to do the install in the background so as to mitigate the restart flash.

nevir:
> and have asked to be able to do the install in the background so as to mitigate the restart flash.

The way we're mitigating that is to expose a hook to show/hide our splash screen (which is handy for masking any view screwups on app launch, too).  We hide the splash screen once everything has spun up (around the same time we call `notifyAppReady`).  Not sure _that_ is easily to abstract, though

lostintangent:
@nevir I threw together prototype of this feature to test it out, which in addition to adding the new install mode, also supports the existing `minimumBackgroundDuration`, that way you can do either of the following:

```javascript
// Restart the app as soon as the app is put into the background,
// instead of waiting for it to be resumed to the foreground
codePush.sync({ installMode: codePush.InstallMode.ON_NEXT_SUSPEND });

// Restart the app _while_ it is in the background, but only after it has
// been in the background for 2 minutes, so that user context isn't
// lost unless the app suspension is long enough to not matter
codePush.sync({ 
    installMode: codePush.InstallMode.ON_NEXT_SUSPEND,
    minimumBackgroundDuration: 2 * 60
});
```

Does that match the use cases you had in mind? It definitely is a pretty awesome experience.

nevir:
Yeah!  That's perfect!  We'd want to use the `minimumBackgroundDuration` approach

(also, just to make sure: I'm assuming this works via `localPackage.install()` as well)

lostintangent:
@nevir Yep this would work with LocalPackage.install as well!

tomauty:
[...]

PR #770, opened 2017-03-28
installModes: implement new ON_NEXT_SUSPEND install mode

sergey-akhalkov:
Restart the app _while_ it is in the background, but only after it has been in the background for `minimumBackgroundDuration` seconds (0 by default), so that user context isn't lost unless the app suspension is long enough to not matter.

Relates to https://github.com/Microsoft/react-native-code-push/issues/366


silhouettes:
@BretJohnson - could you review this PR?

silhouettes:
@BretJohnson just a friendly ping!

sergey-akhalkov:
@BretJohnson, no problem, thanks!

PR #800, opened 2017-04-18
Create tvOS target

cmcewen:
This creates a tvOS target - I haven't tested but at the very least it prevents JS errors when running on tvOS that come from not having the native module included in the project.


max-mironov:
@cmcewen - thanks for contribution. I've tried to apply your PR and then build it in xCode with blank RN app but I've got the following errors:
image
Could you please have a look at this and guide me if I'm missing something?

sergey-akhalkov:
Hi @cmcewen, thanks for the contribution! Currently, we are experiencing some issues with this, so could you please have a chance to look at the exception @max-mironov have sent above?
Could you please also confirm this PR #824 has the same improvements as yours?

max-mironov:
@cmcewen we believe that both - your PR and #824 serves the same goal but the latter also has support for podspecs, so if you don't mind we'll close this one to avoid duplication. Please let us know if you have any concerns or you have any suggestions.

PR #803, opened 2017-04-20
Fix update always available bug for newly installed android apps

jiaminglu:
This pull request fixed:

1. For newly installed android apps, update check always return new bundle, even if the asset bundle is already the latest, since `packageHash` is not passed to `/updateCheck`.
2. A typo in `codepush.gradle`

### DIFF
--- a/CodePush.js
+++ b/CodePush.js
@@ -46,7 +46,7 @@ async function checkForUpdate(deploymentKey = null) {
     queryPackage = localPackage;
   } else {
     queryPackage = { appVersion: config.appVersion };
-    if (Platform.OS === "ios" && config.packageHash) {
+    if (config.packageHash) {
       queryPackage.packageHash = config.packageHash;
     }
   }


sergey-akhalkov:
Hello @jiaminglu, thank you for the contribution. I've just tested your PR and first point (`For newly installed android apps...`) works like a charm, but second one crashes my test app in release mode. I agree that the `def resourcesDirConfigName = "jsBundleDir${targetName}"` line is looks like a typo, but as of now it crashes the app when we change it. 
So I suggest to revert second point for now, because it needs to be investigated more deeply. Please let me know what do you think.

jiaminglu:
That commit is removed.

The default configuration works well and the problem with `resourcesDirConfigName` won't cause any problem if we stick to the default configuration.

The problem with this line is that, if we choose to use custom configurations, it might cause the generated resources to be placed in wrong place, which resulted in wrong hash calculated.

sergey-akhalkov:
@jiaminglu, got it, thanks for the clarification, I'll copy you explanation and create new issue in react-native-code-push repo, just to make sure we won't forgot about it.

max-mironov:
@jiaminglu thanks for contribution, your PR looks good, merging your PR to Master.

PR #813, opened 2017-04-27
WPF:WINDOWS Added ability to update application installed in Program Files folder

abodalevsky:
1) Desktop application can be installed at any location, the most popular are: 
 - _User local data folder_, in this case the bundle can be stored in the same location and be  unique for the user. 
 - _Program Files folder_, in this case the application is unique for the system and will be shared  among all users of the system, the bundle should be unique for the system as well. 

Usually, user has no write access to Program Files folder or at least admin privileges have to been requested.  In this case the bundle will be stored in ProgramData folder as it is recommended.

2) Small refactoring to share more code between UWP and WPF projects.

3) Windows examples were updated to `react-native-windows 0.43.0-rc.0`

### DIFF
--- a/windows/CodePush.Shared/CodePushConstants.cs
+++ b/windows/CodePush.Shared/CodePushConstants.cs
@@ -27,7 +27,7 @@
         internal const string UnzippedFolderName = "unzipped";
 #if WINDOWS_UWP
         internal const string AssetsBundlePrefix = "ms-appx:///ReactAssets/";
-        internal const string FileBundlePrefix = "ms-appdata:///local";
+        internal const string FileBundlePrefix = "ms-appdata:///local/";
 #else
         internal const string AssetsBundlePrefix = "ReactAssets/";
 #endif
--- a/windows/CodePush.Shared/CodePushUtils.cs
+++ b/windows/CodePush.Shared/CodePushUtils.cs
@@ -38,8 +38,7 @@ namespace CodePush.ReactNative
 #if WINDOWS_UWP
             return Package.Current.Id.Version.Major + "." + Package.Current.Id.Version.Minor + "." + Package.Current.Id.Version.Build;
 #else
-            var version = FileVersionInfo.GetVersionInfo(Environment.GetCommandLineArgs()[0]);
-            return $"{version.FileMajorPart}.{version.FileMinorPart}.{version.FileBuildPart}";
+            return applicationInfo.Version;
 #endif
         }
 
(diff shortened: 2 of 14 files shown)


sergey-akhalkov:
Hi @abodalevsky, thank you so much for your contribution, LGMT.

windows/CodePush.Shared/CodePushUtils.cs
```diff
@@ -61,13 +60,10 @@ internal static string GetAssetsBundlePrefix()
...
-        internal static string GetFileBundlePrefix()
+        internal static string ExtractSubFolder(string fullPath)
         {
-#if WINDOWS_UWP
-            return CodePushConstants.FileBundlePrefix;
-#else
-            return GetAppFolder();
-#endif
+            var codePushSubPathArray = fullPath.Split(Path.DirectorySeparatorChar);
+            return String.Join("/", codePushSubPathArray.SkipWhile((value, index) => codePushSubPathArray.Length - index > 4).ToArray());
```
sergey-akhalkov: @abodalevsky, could you please clarify what means number 4?

windows/CodePush.Shared/CodePushConstants.cs
```diff
@@ -27,7 +27,7 @@ internal class CodePushConstants
         internal const string UnzippedFolderName = "unzipped";
 #if WINDOWS_UWP
         internal const string AssetsBundlePrefix = "ms-appx:///ReactAssets/";
-        internal const string FileBundlePrefix = "ms-appdata:///local";
+        internal const string FileBundlePrefix = "ms-appdata:///local/";
```
sergey-akhalkov: @abodalevsky, do I get it right if `FileBundlePrefix` const is used in `GetJavaScriptBundleFileAsync` method only, so there are no possible issues with this change?

windows/CodePush.Shared/CodePushUtils.cs
abodalevsky: 4 means last 4 segments of the path where JS bundle is located. The idea of the change was to unify the calculation of js bundle file location for UWP and WPF. While they have different prefix but last 4 segments are the same for both implementations.
For example for
 `F:\react-native-code-push\Examples\CodePushDemoApp\windows\CodePushDemoApp.Wpf\bin\x64\DebugBundle\CodePush\cc1dd1ab\CodePush\index.windows.bundle`
the part that will be added to specific prefix will be:
`CodePush/cc1dd1ab/CodePush/index.windows.bundle`

windows/CodePush.Shared/CodePushConstants.cs
abodalevsky: It should be fine I have checked it before commit

Issue #816, opened 2017-04-28
CodePush update crashes silently on iOS


peacechen:
### Description

code-push updates are causing our app to crash silently.  This happens only on a release build so I'm not able to capture any debug information.  If I set `CheckFrequency` from `ON_APP_RESUME` to `MANUAL` to stop it from automatically downloading updates, the app does not crash.

In debug mode the app does not crash.  Code Push ON_APP_RESUME and explicit `sync()` updates work in debug mode only.

Everything has been sync'd up between the base app release update and the CodePush updates, so no native components have changed between the deployed app and the CodePush update.

CodePush was working on a previous version of our app, if that makes any difference.

### Reproduction

Build a release version and run on either a real device or simulator.  Using the exact same build, change one line in the JS and release it with code-push:
```
code-push release-react MyAppName ios --plistFile "ios/MyProject/Info.plist" -m --description "test update"
```

### Additional Information

* react-native-code-push version: 1.16.1-beta (have also tried 1.15.1-beta and 1.17.3-beta)
* react-native version: 0.34.1
* iOS/Android/Windows version: iOS 10.2.1
* Does this reproduce on a debug build or release build?  release build only
* Does this reproduce on a simulator, or only on a physical device?  real device and simulator

Amurmurmur:
I think I have a similar issue as you. I was able to catch exception in a release/staging build on ios.
Though a crash only happens on the ios platform.
 #815

peacechen:
Thanks for the reference.  Can't tell whether it's the same issue since our app uses CodePush 1.15-1.17, while yours uses 2.0.2 and yours didn't have this issue on 1.17.  Unless this issue is more fundamental to the CodePush platform and is triggering a bug on multiple versions.

My CodePush CLI decided to expire my login token moments ago, which it hasn't since I first started using it months ago.  Is Microsoft resetting its servers?

axemclion:
@pniko This looks like a serious issue. Could we quickly investigate to see what the issue here is ? Once @sergey-akhalkov is back to work, he could continue the investigation.

pniko:
Sure thing. Setting up a test right now to see if I can reproduce the issue before @sergey-akhalkov becomes available again.

didierbrun:
Hello,
It seems that the mutableUpdatePackage dictionary in CodePushPackage.m contains a "download" key that is a Function, thus it causes a crash during JSON serialization on this line :   
[...]

PR #817, opened 2017-05-01
Fix update crash by removing download key in mutableUpdatePackage.

peacechen:
`download` key is a function and isn't serializable by NSJSONSerialization.
See issue #816.


max-mironov:
@peacechen - thanks for contribution. Since currently the issue is fixed on [JS side](https://github.com/Microsoft/react-native-code-push/pull/809) we believe we can postpone to add this check on native side for now. We will incorporate this fix later when we will integrate ["Support for Brownfield apps"](https://github.com/Microsoft/code-push/issues/423) feature. Currently close this PR. Thanks again for looking at this.

PR #820, opened 2017-05-02
verify code-push service with RN 0.44

sergey-akhalkov:
code push service runs well with RN 0.44


maxim-pop:
I assume package.json version bump and release not needed here.

Issue #841, opened 2017-05-17
[Android] App restart loop when using InstallMode.ON_NEXT_RESUME


AndrewJack:
### Description

When `InstallMode.ON_NEXT_RESUME` is used on android the app will go into a restart loop and crash the app.

### Reproduction

1. Initialise with the following code push options
   ```
     {
          checkFrequency: CheckFrequency.ON_APP_RESUME,
          installMode: InstallMode.ON_NEXT_RESUME
      }
   ```
2. Run the app - wait for it to download the update
3. Press the home button
4. Resume the app from the multi task view
Observe: App will go black and into a loop of `onResume` -> `loadBundle`

I think the cause is around here - https://github.com/Microsoft/react-native-code-push/blob/master/android/app/src/main/java/com/microsoft/codepush/react/CodePushNativeModule.java#L470

### Additional Information

* react-native-code-push version: 2.0.3-beta
* react-native version: 0.44
* iOS/Android/Windows version: Android 7.1
* Does this reproduce on a debug build or release build? Release build
* Does this reproduce on a simulator, or only on a physical device? Reproduced on a device

AndrewJack:
Trying to reproduce on the Example

sergey-akhalkov:
Hi @AndrewJack, thanks for reaching us. 

>Trying to reproduce on the Example

Got it, this should be very useful for us, please let me know if any updates.

AndrewJack:
@sergey-akhalkov I've figured it out.

- Enable proguard
- perform the same steps as above

Then this happens:
- `loadBundle` is called from `onResume`
- It then fails on this line (because proguard + reflection) https://github.com/Microsoft/react-native-code-push/blob/master/android/app/src/main/java/com/microsoft/codepush/react/CodePushNativeModule.java#L160
- This then is caught and `loadBundleLegacy ` is called - https://github.com/Microsoft/react-native-code-push/blob/master/android/app/src/main/java/com/microsoft/codepush/react/CodePushNativeModule.java#L84
- This then re-creates the activity
- Which calls `onHostResume` again - starting the infinite loop - https://github.com/Microsoft/react-native-code-push/blob/master/android/app/src/main/java/com/microsoft/codepush/react/CodePushNativeModule.java#L473

Do we still need this legacy support in v2 of code push?

AndrewJack:
I've fixed my code by adding `-keep class com.facebook.react.** { *; }` for now

AndrewJack:
[...]

PR #845, opened 2017-05-18
Remove support for legacy React Native versions

AndrewJack:
Fixes: #841 

v2 of code push only supports RN 0.43 and later. This means we no longer need to use reflection for backward compatibility.

### DIFF
--- a/android/app/src/main/java/com/microsoft/codepush/react/CodePushNativeModule.java
+++ b/android/app/src/main/java/com/microsoft/codepush/react/CodePushNativeModule.java
@@ -1,14 +1,12 @@
 package com.microsoft.codepush.react;
 
 import android.app.Activity;
-import android.content.Context;
 import android.os.AsyncTask;
 import android.os.Handler;
 import android.os.Looper;
 import android.provider.Settings;
-import android.view.Choreographer;
 
-import com.facebook.react.ReactActivity;
+import com.facebook.react.ReactApplication;
 import com.facebook.react.ReactInstanceManager;
 import com.facebook.react.bridge.Arguments;
 import com.facebook.react.bridge.LifecycleEventListener;
@@ -18,8 +16,8 @@ import com.facebook.react.bridge.ReactContextBaseJavaModule;
 import com.facebook.react.bridge.ReactMethod;
 import com.facebook.react.bridge.ReadableMap;
 import com.facebook.react.bridge.WritableMap;
-import com.facebook.react.modules.core.DeviceEventManagerModule;
 import com.facebook.react.modules.core.ChoreographerCompat;
+import com.facebook.react.modules.core.DeviceEventManagerModule;
 import com.facebook.react.modules.core.ReactChoreographer;
 
 import org.json.JSONArray;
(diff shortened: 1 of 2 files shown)


android/app/src/main/java/com/microsoft/codepush/react/CodePushNativeModule.java
```diff
@@ -88,6 +83,11 @@ private void loadBundleLegacy() {
             // no-op to prevent any null pointer exceptions.
             return;
         }
+
+        if (mLifecycleEventListener != null) {
+            getReactApplicationContext().removeLifecycleEventListener(mLifecycleEventListener);
+        }
```
AndrewJack: If restarting we need to remove the removeLifecycleEventListener

AndrewJack: Should this callback be removed after a successful install?

android/app/src/main/java/com/microsoft/codepush/react/CodePushNativeModule.java
```diff
@@ -124,24 +119,20 @@ private void setJSBundle(ReactInstanceManager instanceManager, String latestJSBu
...
-            Field jsBundleField = instanceManager.getClass().getDeclaredField("mJSBundleFile");
-            jsBundleField.setAccessible(true);
-            jsBundleField.set(instanceManager, latestJSBundleFile);
+            CodePushUtils.log("Unable to set JSBundle - CodePush may not support this version of React Native");
+            throw new IllegalAccessException("Could not setJSBundle");
         }
     }
 
     private void loadBundle() {
+        clearLifecycleEventListener();
```
AndrewJack: Any issue with clearing this before we attempt to load the bundle? We don't need the callback after this point.

We must call this before `loadBundleLegacy()` or it will go into an infinite restart loop.

AndrewJack:
@sergey-akhalkov any chance of this being reviewed/ merged soon?

max-mironov:
@AndrewJack - thanks for great contribution. I believe we should test it very careful before merging this changes. 
Also I'vent got a chance yet to dive deeper into it but we have one [PR](https://github.com/Microsoft/react-native-code-push/pull/861) that probably has some overlap with your changes, do you mind to have a quick look at it also to see if it somehow deals with your problem and if you think it ever make sense (or your changes will fix [this item ](https://github.com/Microsoft/react-native-code-push/issues/847)also). We really appreciate your experience here and your insight would be very valuable for us.

AndrewJack:
@max-mironov I've had a look through your changes. I'll rebase my PR once yours has been merged.

AndrewJack:
@max-mironov I've rebased after your changes were merged.

max-mironov:
Hello @AndrewJack. Your changes looks good and reasonable to me! I've double checked all main cases (with different installModes) on emulator and on real device with signed apk (Android7) and haven't found any issues. We really appreciate your efforts and help here - thank you for this!

I'm good with merging this stuff but probably @sergey-akhalkov  could also have a look at this before we incorporate this changes to Master.

Issue #795, opened 2017-04-14
Attempt to invoke virtual method 'char[] java.lang.String.toCharArray()' on a null object reference


xwartz:
### Description

This crash happens on Android,  but we do not have the relevant equipment, only have the Fabric crash log.

### Reproduction

Fabric logs:

```
Fatal Exception: java.lang.RuntimeException: Unable to start activity ComponentInfo{org.consenlabs.token/org.consenlabs.token.MainActivity}: java.lang.NullPointerException: Attempt to invoke virtual method 'char[] java.lang.String.toCharArray()' on a null object reference
       at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:2464)
       at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:2524)
       at android.app.ActivityThread.access$900(ActivityThread.java:154)
       at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1391)
       at android.os.Handler.dispatchMessage(Handler.java:102)
       at android.os.Looper.loop(Looper.java:234)
       at android.app.ActivityThread.main(ActivityThread.java:5526)
       at java.lang.reflect.Method.invoke(Method.java)
       at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:726)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:616)
Caused by java.lang.NullPointerException: Attempt to invoke virtual method 'char[] java.lang.String.toCharArray()' on a null object reference
       at java.io.File.fixSlashes(File.java:183)
       at java.io.File.(File.java:130)
       at com.microsoft.codepush.react.FileUtils.deleteDirectoryAtPath(FileUtils.java:58)
       at com.microsoft.codepush.react.CodePushUpdateManager.rollbackPackage(CodePushUpdateManager.java:292)
       at com.microsoft.codepush.react.CodePush.rollbackPackage(CodePush.java:259)
       at com.microsoft.codepush.react.CodePush.initializeUpdateAfterRestart(CodePush.java:218)
       at com.microsoft.codepush.react.CodePush.(CodePush.java:76)
       at org.consenlabs.token.MainApplication$1.getPackages(MainApplication.java:70)
       at com.facebook.react.ReactNativeHost.createReactInstanceManager(ReactNativeHost.java:74)
       at com.facebook.react.ReactNativeHost.getReactInstanceManager(ReactNativeHost.java:41)
[...]

PR #870, opened 2017-06-01
If current package folder is null don't delete

AndrewJack:
Fixes: #795 

I've seen this in my crashlytics logs - same as the above issue.

**Issue**: `getCurrentPackageFolderPath()` can return null

**Solution**: check if path is null before deleting the directory.


AndrewJack:
@matrosov-nikita I've made those suggested changes

PR #884, opened 2017-06-15
fix import example for #import <React/RCTLog.h>

nerdmed:
Left the instructions for RN < 40 as the docs seem to keep that in mind.


sergey-akhalkov:
Hi @nerdmed, thanks for the contribution! Looks good to me, so merging this.

PR #897, opened 2017-06-21
[WPF] Fix application crash when storage file was deleted while app is being active.

abodalevsky:
If by some reason persistent storage file was deleted while application is running, next time when CodePush tries to access it it crashes with the following exception: 
`System.IO.FileNotFoundException: Could not find file 'C:\Users\USER_NAME\AppData\Local\PATH_TO_APP\CodePush\CodePush'.` 

The fix is, if file was deleted - recreate the storage file. Of course all settings will be lost when file was deleted.


Examples/CodePushDemoApp/windows/CodePushDemoApp/CodePushDemoApp.nuget.targets
```diff
@@ -1,9 +0,0 @@
-﻿<?xml version="1.0" encoding="utf-8" standalone="no"?>
```
abodalevsky: this file should be created by Visual Studio dynamically

sergey-akhalkov:
Hi @abodalevsky, thanks for the contribution, LGTM.

Issue #836, opened 2017-05-11
bundle generation TEMP_FILE_PATH sometimes breaks parallel react-native builds


webkodierer:
### Description

During bundle creation a temp file is used in _TMPDIR_/CodePushResourcesMap.json 

For concurrent executions (for us this is the case on our continuous integration server) this allows race conditions resulting in unexpected build failures and _maybe_ to other unexpected results. 

Sequence: 

First in scripts/recordFilesBeforeBundleCommand.js the file is created and written: 
```
var TEMP_FILE_PATH = path.join(require("os").tmpdir(), "CodePushResourcesMap.json");
// [ ... ]
fs.writeFile(TEMP_FILE_PATH, JSON .... 
```

later in scripts/generateBundledResourcesHash.js the file is read and removed: 
```
var TEMP_FILE_PATH = path.join(require("os").tmpdir(), "CodePushResourcesMap.json");
// [ ... ]
var oldFileToModifiedTimeMap = require(TEMP_FILE_PATH);
// [ ... ]
fs.unlinkSync(TEMP_FILE_PATH);
```

For us it happens quite often that multiple jobs are running in parallel, breaking each other. 
This is the sample output from the affected jenkins job:

```
[ ... ]
bundle: Done copying assets
:app:generateBundledResourcesHashLocalReleasemodule.js:471
    throw err;
    ^

Error: Cannot find module '/tmp/CodePushResourcesMap.json'
    at Function.Module._resolveFilename (module.js:469:15)
    at Function.Module._load (module.js:417:25)
    at Module.require (module.js:497:17)
    at require (internal/module.js:20:19)
    at Object.<anonymous> (/srv/jenkins_home/workspace/my-feature-branch-S7T3KVECBPSJLFN6HQNB4IPKGNWPRYTU3HWBIQIZ6YVEAAKS4E6Q/node_modules/react-native-code-push/scripts/generateBundledResourcesHash.js:32:32)
    at Module._compile (module.js:570:32)
    at Object.Module._extensions..js (module.js:579:10)
    at Module.load (module.js:487:32)
    at tryModuleLoad (module.js:446:12)
    at Function.Module._load (module.js:438:3)
 FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':app:generateBundledResourcesHashLocalRelease'.
> Process 'command 'node'' finished with non-zero exit value 1
```
### Reproduction

[...]

PR #902, opened 2017-06-23
Fix dependency issue

mikelambert:
Fix this crash I see when building for android:
```
:app:generateBundledResourcesHashDebug
module.js:471
    throw err;
    ^

Error: Cannot find module '/var/folders/94/wngs1jw91_n2_jjjrfljtqrc0000gn/T/CodePushResourcesMap-1b95774b.json'
```

Looks like these two tasks (`recordFileBeforeBundleCommand` and `generateBundledResourcesHash`) run in parallel, which create problems because one depends on a file generated by the other. Not sure why others don't have this issue, though, but I suspect this is exactly the issue reported in https://github.com/Microsoft/react-native-code-push/issues/836

### DIFF
--- a/android/codepush.gradle
+++ b/android/codepush.gradle
@@ -80,6 +80,9 @@ gradle.projectsEvaluated {
                 commandLine (*nodeExecutableAndArgs, "${nodeModulesPath}/react-native-code-push/scripts/generateBundledResourcesHash.js", resourcesDir, "$jsBundleDir/$bundleAssetName", jsBundleDir, resourcesMapTempFileName)
             }
 
+            // We need to generate and record the resources map, but we use it to generate the bundle hash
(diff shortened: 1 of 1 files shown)


sergey-akhalkov:
Hi @mikelambert, thanks for the contribution, LGTM!

mikelambert:
That changelist 857 was supposed to help the original reported bug 836.

However, AFAIK the bug was a race condition where the "reader" ran before the "writer" did, resulting in an error trying to open the file.

Passing in the same filename to both scripts (as 857 does) is smart and probably a better design, but it doesn't change anything to prevent the writer from running before the reader (which can still happen with that change).

Thus this change, which enforces an explicit ordering dependency between them.

sergey-akhalkov:
@mikelambert, got it, thanks for the clarification!

PR #920, opened 2017-07-14
Handle TargetBinaryVersionMismatch #2

disharmonized:
Added an optional callback function that can be used to notify user if there are any binary updates.


CodePush.js
```diff
@@ -237,7 +240,7 @@ const sync = (() => {
   let syncInProgress = false;
   const setSyncCompleted = () => { syncInProgress = false; };
 
-  return (options = {}, syncStatusChangeCallback, downloadProgressCallback) => {
+  return (options = {}, syncStatusChangeCallback, downloadProgressCallback, handleBinaryVersionMismatch) => {
```
max-mironov: If this 4th parameter is callback may be it make sense to add *callback* postfix to this to avoid confusion (on first sight looks like boolean variable for me, not a function)

CodePush.js
```diff
@@ -76,6 +76,9 @@ async function checkForUpdate(deploymentKey = null) {
       (!localPackage || localPackage._isDebugOnly) && config.packageHash === update.packageHash) {
     if (update && update.updateAppVersion) {
       log("An update is available but it is not targeting the binary version of your app.");
+      if(handleBinaryVersionMismatch && typeof handleBinaryVersionMismatch === "function") {
```
max-mironov: Probably space is missing between if and brackets

disharmonized: got it

CodePush.js
disharmonized: fixed

max-mironov:
LGTM. @sergey-akhalkov - if you could also have a look at this and don't have any concerns I'm ok with merging this.

Issue #921, opened 2017-07-14
[Android] Rollback + Binary update causes app to crash


AndrewJack:
### Description

`initializeUpdateAfterRestart` is called (in the constructor of `CodePush`)  before `getJSBundleFileInternal` & `clearUpdates`. This can be confirmed by running the example app.

This causes the app to get into a unexpected state and crash when rolling back because the expected `codepush.json` status file has been deleted.

If a binary update has occurred the `CodePush/` directory and settings are cleared. If the app has a  pending update which has to be rolled back the app will crash because the necessary file have been deleted.

I will post a few stack traces captured by crashlytics, which are side effects of this core issue.

~The use of headlessTask may be part of this issue.~ HeadlessTask only caused the rollbacks due to the absense of `notifyAppReady`

### Reproduction

1. Run the example app with `InstallMode.ON_NEXT_RESTART`
2. Push a code push update that causes a rollback or don't call `codePush.notifyAppReady()` or `sync`
3. Let the code push install
4. Quit the app and remove from memory
5. Open app
6. Quit the app and remove from memory
7. Install a new binary version 
8. Run the app

**Actual**:
Observe crash
**Expected**:
Should load binary version without crashing

### Additional Information

* react-native-code-push version: 2.1.0
* react-native version: 0.44.2
* iOS/Android/Windows version: Android (any)
* Does this reproduce on a debug build or release build? release
* Does this reproduce on a simulator, or only on a physical device? either

AndrewJack:
```
Fatal Exception: java.lang.RuntimeException: Unable to start service com.example.react.headlessTask.CodePushTaskService@ef12e5f with Intent { cmp=com.example/.react.headlessTask.CodePushTaskService }: com.microsoft.codepush.react.g: Error updating current package info
       at android.app.ActivityThread.handleServiceArgs(ActivityThread.java:3314)
       at android.app.ActivityThread.-wrap21(ActivityThread.java)
       at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1565)
       at android.os.Handler.dispatchMessage(Handler.java:102)
       at android.os.Looper.loop(Looper.java:154)
       at android.app.ActivityThread.main(ActivityThread.java:6077)
       at java.lang.reflect.Method.invoke(Method.java)
       at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:865)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:755)
[...]

PR #924, opened 2017-07-17
Prevent Rollback crash when binary was updated.

AndrewJack:
Fixes: #921 

Senario:
When a code push update failed (rollback on next start), but the binary version was updated before the next restart.

This causes a the initialisation code to start a rollback, and clear the previous update at the same time causing IO crashes due to the unexpected state (deleted `/CodePush` directory).

Easier to reproduce with `InstallMode.ON_NEXT_RESTART`

### DIFF
--- a/android/app/src/main/java/com/microsoft/codepush/react/CodePush.java
+++ b/android/app/src/main/java/com/microsoft/codepush/react/CodePush.java
@@ -143,7 +143,6 @@ public class CodePush implements ReactPackage {
     public String getJSBundleFileInternal(String assetsBundleFileName) {
         this.mAssetsBundleFileName = assetsBundleFileName;
         String binaryJsBundleUrl = CodePushConstants.ASSETS_BUNDLE_PREFIX + assetsBundleFileName;
-        long binaryResourcesModifiedTime = this.getBinaryResourcesModifiedTime();
 
         try {
             String packageFilePath = mUpdateManager.getCurrentPackageBundlePath(this.mAssetsBundleFileName);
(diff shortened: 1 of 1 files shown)


max-mironov:
@AndrewJack Great stuff and thanks for taking care of this. Looks good to me. 
cc @sergey-akhalkov to also have a look at this before merging

android/app/src/main/java/com/microsoft/codepush/react/CodePush.java
```diff
@@ -232,6 +228,24 @@ boolean isRunningBinaryVersion() {
         return sIsRunningBinaryVersion;
     }
 
+    private boolean isPackageBundleLatest(JSONObject packageMetadata) {
+        Long binaryModifiedDateDuringPackageInstall = null;
+        String binaryModifiedDateDuringPackageInstallString = packageMetadata.optString(CodePushConstants.BINARY_MODIFIED_TIME_KEY, null);
+        if (binaryModifiedDateDuringPackageInstallString != null) {
+            binaryModifiedDateDuringPackageInstall = Long.parseLong(binaryModifiedDateDuringPackageInstallString);
+        }
```
AndrewJack: Ah just spotted this - The `NumberFormatException` should be caught

android/app/src/main/java/com/microsoft/codepush/react/CodePush.java
```diff
@@ -193,6 +183,12 @@ void initializeUpdateAfterRestart() {
         // the app was just freshly updated.
         mDidUpdate = false;
 
+        JSONObject packageMetadata = this.mUpdateManager.getCurrentPackage();
+        if (!isPackageBundleLatest(packageMetadata) && hasBinaryVersionChanged(packageMetadata)) {
+            CodePushUtils.log("Skipping initializeUpdateAfterRestart(), binary version is newer");
+            return;
+        }
```
AndrewJack: Should I move this inside `if (pendingUpdate != null) {` to avoid reading the file `getCurrentPackage` file every time.

android/app/src/main/java/com/microsoft/codepush/react/CodePush.java
max-mironov: Aye, good note, and it can be moved away from `getJSBundleFileInternal` method probably

android/app/src/main/java/com/microsoft/codepush/react/CodePush.java
max-mironov: Agreed!

sergey-akhalkov:
Hi @AndrewJack, thank you for the contribution, looks good to me.

Issue #927, opened 2017-07-18
showing an update dialog in iOS


kelset:
In the [`Plugin Usage`](https://github.com/Microsoft/react-native-code-push#plugin-usage) section of the README, the very last paragraph says:

> NOTE: While Apple's developer agreement fully allows performing over-the-air updates of JavaScript and assets (which is what enables CodePush!), it is against their policy for an app to display an update prompt. Because of this, we recommend that App Store-distributed apps don't enable the updateDialog option when calling sync, whereas Google Play and internally distributed apps (e.g. Enterprise, Fabric, HockeyApp) can choose to enable/customize it.

I've been going through the Apple Developer Program License Agreement, which can be downloaded from [this page](https://developer.apple.com/terms/), but I couldn't find any rule around **not** showing/presenting the user with an Alert / a UI element to tell him/her to update the app to the latest version.

Moreover, the only instance of "not allowance" I could find was an [Ionic app back in January](https://forum.ionicframework.com/t/ionic-deploy-ios-apple-app-store-rejection/77971) and lately apps like Ryanair and Monzo have adopted a similar behaviour (showing a full page "alert" to tell you to update).

Can it be that maybe now this "interaction" is not disallowed anymore?

Even the "mighty" paragraph 3.3.2 has been updated to be more "open" to OTA installations:

> Interpreted code may be downloaded to an Application but only so long as such code: (a) does not change the primary purpose of the Application by providing features or functionality that are inconsistent with the intended and advertised purpose of the Application as submitted to the App Store, (b) does not create a store or storefront for other code or applications, and (c) does not bypass signing, sandbox, or other security features of the OS.

(version 20170605)

Do you have any more "precise" information around this issue? Or should the README be updated to not disincentivizes the update dialog?

max-mironov:
Hello @kelset, thank you for your question.

So while there is actually no strict rule and 100% guarantee that the app will be banned this way users should be aware of the following statement (taken from [AppStore Review Guidance](https://developer.apple.com/app-store/review/guidelines/)):

> Apps must not force users to rate the app, review the app, download other apps, or other similar actions in order to access functionality, content, or use of the app.

That's the thing that should be taken into account while using the update dialog with iOS.

kelset:
Hey @max-mironov, thanks for pointing out that extract, didn't read that document.
I guess that, then, there is probably a fine line there of showing something without "forcing" and still allowing the user to keep using the app in the standard way.

[...]

PR #936, opened 2017-07-22
Clarify store rulings

kelset:
Related to issue #927, my take on expanding the section that mentions Apple's ruling over OTA updates & update prompts.

I think it is worth linking to the actual docs by Apple to remove any doubts from the reader, and avoid confusion or buzzing around compliancy.


pniko:
Thanks for the feedback @kelset!

PR #958, opened 2017-08-08
implement script to generate CodePushified React Native apps

sergey-akhalkov:
The script serves to generate CodePushified React Native app to reproduce issues or for testing purposes.

**The script fully automatically:**
1. creates CodePush apps for Android and iOS
2. generates plain react-native project
3. installs React Native Module for CodePush
4. inserts Staging deployment keys both for Android and iOS

**Requirements:**
1. npm i -g react-native-cli
2. npm i -g code-push-cli
3. code-push register

**Usage:** `node create-app.js <appName> <reactNativeVersion> <reactNativeCodePushVersion>`

Parameters (all optional):
```
    <appName> - CodePushDemoAppTest
    <reactNativeVersion> - react-native@latest
    <reactNativeCodePushVersion> - react-native-code-push@latest
```


Examples/create-app.js
```diff
@@ -0,0 +1,128 @@
...
+        else {
+            console.log(err);
+        }
+    });
+
+function setupAssets() {
+    fs.unlinkSync('./index.ios.js');
+    fs.unlinkSync('./index.android.js');
+
+    fs.createReadStream('../CodePushDemoApp/demo.js').pipe(fs.createWriteStream('demo.js'));
```
ruslan-bikkinin: may be you should use writeFileSync, readFileSync here?

Examples/create-app.js
```diff
@@ -0,0 +1,128 @@
...
+        fs.writeFile('demo.js', result, 'utf8', function (err) {
+            if (err) return console.error(err);
+            setupAccess();
+        });
+    });
+}
+
+var setupAccess = function () {
+    process.chdir('../');
+    execSync('chown -R `whoami` ' + appName);
```
ruslan-bikkinin: There is no `chown` commands on windows, it should be check for platform here

Examples/create-app.js
sergey-akhalkov: @ruslan-bikkinin, are there any reasons to use writeFileSync and readFileSync instead of createReadStream and createWriteStream?

Examples/create-app.js
sergey-akhalkov: fixed

Examples/create-app.js
ruslan-bikkinin: @sergey-akhalkov yeah, streams are working asynchronously and should be processed respectively. Usage of writeFileSync/readFileSync are the simplest alternative to copy files in this particular case.

sergey-akhalkov: fixed, thanks

buptkang:
@sergey-akhalkov after merging this PR, you can delete the `Examples/generate-app.sh` code since that script achieves the same goal as this PR.

sergey-akhalkov:
Hi @buptkang, thanks for the note, will do. The main difference between `generate-app.sh` and `create-app.js` is that `create-app.js` also creates 2 CodePush apps using `code-push app add <appName> <platform> react-native` command both for Android and iOS. Then the script inserts Staging deployment keys while linking. 
Also the `create-app.js` allows to specify not package version only, but [source code location](https://medium.com/@joshyhargreaves/react-native-init-from-react-native-master-branch-or-any-untagged-commit-dd048d2a4db7) too both for react-native and react-native-code-push.

buptkang:
Totally agree on it. That is reason we should remove that `generate-app.sh` code and use `create-app.js` instead. We should also update the document so that users can use this script to re-generate issues that they raise up.

PR #966, opened 2017-08-11
Implement code signing for client android SDK

ruslan-bikkinin:
Basic support for bundle's signature verification for android applications has been implemented. 

By default, feature is disabled, so no need for clients not interested in code signing to worry about changing code-push sdk initialization code after update.

To use this feature, developers need to use brand new [CodePushBuilder](https://github.com/Microsoft/react-native-code-push/pull/966/files#diff-8e6ec6fad0f8525783fa4ae3c0d6a2a5) or advanced [constructor](https://github.com/ruslan-bikkinin/react-native-code-push/blob/53a933b6ab910e5c8b501dc62c0f886c84b31e35/android/app/src/main/java/com/microsoft/codepush/react/CodePush.java#L89-L109) with parameter `publicKeyResourceDescriptor` specified. Public key must be assisted in `string.xml`, so it can be available via `R.string` reference.

Initialization examples:

* via `CodePushBuilder`
```java
    @Override
    protected List<ReactPackage> getPackages() {
              new MainReactPackage(),
              new CodePushBuilder("deployment-key-here",getApplicationContext())
                      .setPublicKeyResourceDescriptor(R.string.publicKey /* string item in strings.xml */)
                      .build()
      );
    }
```

* via constructor
```java
    @Override
    protected List<ReactPackage> getPackages() {
              new MainReactPackage(),
              new CodePush(
                      "deployment-key-here",
                      getApplicationContext(),
                      BuildConfig.DEBUG,
                      "https://codepush.azurewebsites.net/"
                      R.string.publicKey /* string item in strings.xml */)
      );
    }
```


ruslan-bikkinin:
Description updated with more detailed information about feature usage.

geof90:
Awesome to finally see progress in this!

android/app/src/main/java/com/microsoft/codepush/react/CodePush.java
```diff
@@ -72,11 +81,31 @@ public CodePush(String deploymentKey, Context context, boolean isDebugMode) {
...
     }
 
+    public CodePush(String deploymentKey, Context context, boolean isDebugMode, @NonNull String serverUrl, Integer publicKeyResourceDescriptor) {
+        this(deploymentKey, context, isDebugMode, serverUrl);
+
+        if (publicKeyResourceDescriptor != null) {
+            try {
+                mPublicKey = mContext.getString(publicKeyResourceDescriptor);
+            } catch (Resources.NotFoundException e) {
+                throw new CodePushUnknownException(
```
max-mironov: I see that we have several places where we throw an `UnknownException` but why don't we add a more meaningful name to the exception class e.g. something like `CodePushInvalidPublicKeyException`? 
The key point here from my side here - is this a really Unknown exception or do we know what kind of  exception it is?

android/app/src/main/java/com/microsoft/codepush/react/CodePushUpdateManager.java
```diff
@@ -244,7 +245,16 @@ public void downloadPackage(JSONObject updatePackage, String expectedBundleFileN
                 }
 
                 if (isDiffUpdate) {
-                    CodePushUpdateUtils.verifyHashForDiffUpdate(newUpdateFolderPath, newUpdateHash);
+                    CodePushUtils.log("Applying diff update.");
```
max-mironov: Don't see if we ever need this `if block` here due to we can move this `log` statement (and this is the only thing that this block is doing now) to Line 255 right after new `if (isDiffUpdate)` block begins.

ruslan-bikkinin: @max-mironov no, we can't, because logging should be executed anyway regardless signature verification enabled or not.

android/app/src/main/java/com/microsoft/codepush/react/CodePush.java
ruslan-bikkinin: @max-mironov good idea, thanks, i'll try to do it

android/app/src/main/java/com/microsoft/codepush/react/CodePush.java
```diff
@@ -36,18 +39,24 @@
 
     // Config properties.
     private String mDeploymentKey;
-    private String mServerUrl = "https://codepush.azurewebsites.net/";
+    private static String mServerUrl = "https://codepush.azurewebsites.net/";
 
     private Context mContext;
     private final boolean mIsDebugMode;
 
+    private String mPublicKey;
```
ruslan-bikkinin: @max-mironov it is not so critical, but, yeah, i can do it

android/app/src/main/java/com/microsoft/codepush/react/CodePushUpdateManager.java
max-mironov: Gotcha! May be we can also state what kind of update is applying currently with
`isDiffUpdate ? CodePushUtils.log("Applying diff update.") : CodePushUtils.log("Applying full update.")` but that is completely up to you.

ruslan-bikkinin: @max-mironov, good proposition, i think i'll do it your way ☝️

ruslan-bikkinin: @max-mironov changes added, thx!

PR #974, opened 2017-08-24
Implement code signing for client ios SDK

ruslan-bikkinin:
Basic support for bundle's signature verification for ios applications has been implemented. 

By default, feature is disabled, so no need for clients not interested in code signing to worry about changing code-push sdk initialization code after update.

To use this feature, developers need to add `PublicKey` string row in`Info.plist` with content of public key they want to use to verify bundles signature.


ios/CodePush/CodePushConfig.m
```diff
@@ -34,7 +35,8 @@ - (instancetype)init
     NSString *buildVersion = [infoDictionary objectForKey:(NSString *)kCFBundleVersionKey];
     NSString *deploymentKey = [infoDictionary objectForKey:@"CodePushDeploymentKey"];
     NSString *serverURL = [infoDictionary objectForKey:@"CodePushServerURL"];
-
+    NSString *publicKey = [infoDictionary objectForKey:@"PublicKey"];
```
max-mironov: As the value is taken from plist file that can be shared between different libraries does it make sense to name it `CodePushPublicKey` to avoid confusion about the reference of this property? (the same as for deploymentKey/ServerUrl)

ios/CodePush/CodePushPackage.m
```diff
@@ -39,8 +39,22 @@ + (void)downloadAndReplaceCurrentBundle:(NSString *)remoteBundleUrl
     }
 }
 
++ (void) handleFailedDataIntegrityCheck:(void (^)(NSError *err))failCallback
+                                   error:(NSError **)error
+{
+    CPLog(@"The update contents failed the data integrity check.");
+    if (!error || !*error) {
```
max-mironov: Not sure if `!*error` is needed here, I believe `if (!error)` should be enough.

max-mironov: Also don't understand why do we need at all `err` variable, seems that it is not used anywhere at all

ios/CodePush/CodePushPackage.m
```diff
@@ -231,18 +245,38 @@ + (void)downloadPackage:(NSDictionary *)updatePackage
...
                                                         }
                                                         
-                                                        if (isDiffUpdate && ![CodePushUpdateUtils verifyHashForDiffUpdate:newUpdateFolderPath
-                                                                                                             expectedHash:newUpdateHash
-                                                                                                                    error:&error]) {
-                                                            if (error) {
-                                                                failCallback(error);
+                                                        if(isDiffUpdate){
+                                                            CPLog(@"Applying diff update.");
+                                                        }else{
```
max-mironov: Just style issue to use spaces between `}else{` or probably use [Ternary ](https://en.wikipedia.org/wiki/Ternary_operation)operator here

ios/CodePush/CodePushPackage.m
```diff
@@ -231,18 +245,38 @@ + (void)downloadPackage:(NSDictionary *)updatePackage
...
+                                                            CPLog(@"Applying full update.");
+                                                        }
+                                                        
+                                                        BOOL isSignatureVerificationEnabled = (publicKey != nil);
+                                                        if (isSignatureVerificationEnabled) {
+                                                            BOOL isSignatureValid = [CodePushUpdateUtils verifySignatureFor:newUpdateFolderPath
+                                                                                      withPublicKey:publicKey
+                                                                                              error:&error];
+                                                            if(!isSignatureValid){
```
max-mironov: Same style issue, please use spaces (and please check all the places below)

ios/CodePush/CodePushUpdateUtils.m
```diff
@@ -256,7 +280,88 @@ + (BOOL)verifyHashForDiffUpdate:(NSString *)finalUpdateFolder
...
     }
     
+    CPLog(@"Expected hash: %@, actual hash: %@", expectedHash, updateContentsManifestHash);
+    
     return [updateContentsManifestHash isEqualToString:expectedHash];
 }
 
-@end
\ No newline at end of file
++ (NSString *)cleanPublicKey:(NSString *)publicKeyString
```
max-mironov: From my point of view `Clean` is not a good name here, probably `preparePublicKeyDecoding` or something else could fit better

ios/CodePush/CodePushUpdateUtils.m
```diff
@@ -256,7 +280,88 @@ + (BOOL)verifyHashForDiffUpdate:(NSString *)finalUpdateFolder
...
+    if ([[NSFileManager defaultManager] fileExistsAtPath:signatureFilePath]) {
+        return [NSString stringWithContentsOfFile:signatureFilePath encoding:NSUTF8StringEncoding error:error];
+    }else{
+        *error = [CodePushErrorUtils errorWithMessage:[NSString stringWithFormat: @"Cannot find signature at %@", signatureFilePath]];
+        return nil;
+    }
+    return nil;
+}
+
++ (NSDictionary *) verifyJWT:(NSString *) signature
```
max-mironov: I can see that other `verify` methods here return `BOOL`  value (that make sense!) but this one is not. Maybe `parseJWT` should better reflect the purpose of it.

ios/CodePush/CodePushConfig.m
ruslan-bikkinin: @max-mironov makes sense, I'll do it, tnx!

ios/CodePush/CodePushUpdateUtils.m
```diff
@@ -256,7 +280,88 @@ + (BOOL)verifyHashForDiffUpdate:(NSString *)finalUpdateFolder
...
+                     error:(NSError **)error
+{
+    NSLog(@"Verifying signature for folder path: %@", folderPath);
+    
+    NSString *publicKey = [self cleanPublicKey: publicKeyString];
+    
+    NSString *signature = [self getSignatureFor: folderPath
+                                          error: error];
+    if(signature == nil) {
+        if(error && *error){
```
max-mironov: Again please verify if we ever need this `error && *error`?

ios/CodePush/CodePushUpdateUtils.m
```diff
@@ -256,7 +280,88 @@ + (BOOL)verifyHashForDiffUpdate:(NSString *)finalUpdateFolder
...
+        return false;
+    }
+    
+    NSDictionary *envelopedPayload = [self verifyJWT:signature withPublicKey:publicKey error:error];
+    if(envelopedPayload == nil){
+        CPLog(@"The update could not be verified because it was not signed by a trusted party. %@", *error);
+        return false;
+    }
+    
+    if(![envelopedPayload objectForKey:@"contentHash"]){
```
max-mironov: Probably could move `@contentHash` to constants

ios/CodePush/CodePushUpdateUtils.m
```diff
@@ -256,7 +280,88 @@ + (BOOL)verifyHashForDiffUpdate:(NSString *)finalUpdateFolder
...
+}
+
++ (NSString *)getSignatureFor:(NSString *)folderPath
+                        error:(NSError **)error
+{
+    NSString *signatureFilePath = [NSString stringWithFormat:@"%@/%@/%@", folderPath, ManifestFolderPrefix, BundleJWTFile];
+    if ([[NSFileManager defaultManager] fileExistsAtPath:signatureFilePath]) {
+        return [NSString stringWithContentsOfFile:signatureFilePath encoding:NSUTF8StringEncoding error:error];
+    }else{
+        *error = [CodePushErrorUtils errorWithMessage:[NSString stringWithFormat: @"Cannot find signature at %@", signatureFilePath]];
```
max-mironov: Do we need to check for null here before assigning variable?

ios/CodePush/CodePushPackage.m
ruslan-bikkinin: > Also don't understand why do we need at all err variable, seems that it is not used anywhere at all

this is a part of callback function signature

ruslan-bikkinin: >Not sure if !*error is needed here, I believe if (!error) should be enough.

Okay, i will refactor it to simplify conditions

ios/CodePush/CodePushPackage.m
ruslan-bikkinin: @max-mironov got it!

ios/CodePush/CodePushPackage.m
ruslan-bikkinin: @max-mironov got it!

ios/CodePush/CodePushUpdateUtils.m
ruslan-bikkinin: @max-mironov I'll fix it, thx!

ios/CodePush/CodePushUpdateUtils.m
ruslan-bikkinin: @max-mironov I'll tried to make code on android and ios looks the same as possible, so that's why I call this function that way. If it is crucial, I can change it on both platforms, what do you think?

ios/CodePush/CodePushUpdateUtils.m
ruslan-bikkinin: @max-mironov not so important for now, i think. May be later, if payload structure will become more complicated.

ios/CodePush/CodePushUpdateUtils.m
ruslan-bikkinin: @max-mironov No, we don't, because we just send Error to method caller in order to it be able to process error. This is method's caller responsibility to provide initialized pointer for NSError*.

ios/CodePush/CodePushUpdateUtils.m
ruslan-bikkinin: @max-mironov Okay, thx!

ios/CodePush/CodePushUpdateUtils.m
max-mironov: yep, looks like this can be modified for both SDK's

ios/CodePush/CodePushUpdateUtils.m
max-mironov: Agreed!

ios/CodePush/CodePushUpdateUtils.m
max-mironov: Well, that should be in most cases, but if someone will not do it we'll get an exception here. Although we do not expect that someone else should call it who knows how it would be changed in future? I'd prefer to stick to the arguments provided e.g. in [this answer](https://stackoverflow.com/a/16969465/8434706). But that's only my opinion, no pressure from my side :)

ios/CodePush/CodePushPackage.m
```diff
@@ -230,19 +244,36 @@ + (void)downloadPackage:(NSDictionary *)updatePackage
...
-                                                                                                                    error:&error]) {
-                                                            if (error) {
-                                                                failCallback(error);
+                                                        BOOL isSignatureVerificationEnabled = (publicKey != nil);
+                                                        if (isSignatureVerificationEnabled) {
+                                                            BOOL isSignatureValid = [CodePushUpdateUtils verifySignatureFor:newUpdateFolderPath
+                                                                                      withPublicKey:publicKey
+                                                                                              error:&error];
+                                                            if (!isSignatureValid) {
+                                                                [self handleFailedDataIntegrityCheck:failCallback error:&error];
```
sergey-akhalkov: I would prefer to avoid of using `handleFailedDataIntegrityCheck` method at all and implement error handling like this:
```objective-c
    if (!error) {
        error = [CodePushErrorUtils errorWithMessage:@"Code Signing integrity check error..."];
    }

    CPLog(error);

    failCallback(*error);
    return;
``` 
This will allow us to have different error messages for code signing integrity check and for general integrity check determine in future where the issue occurs.

ios/CodePush/CodePushPackage.m
```diff
@@ -230,19 +244,36 @@ + (void)downloadPackage:(NSDictionary *)updatePackage
...
+                                                            } else {
+                                                                CPLog(@"The update contents succueded the data integrity check.");
+                                                            }
+                                                        } else {
+                                                            if (isDiffUpdate) {
+                                                                if (![CodePushUpdateUtils verifyFolderHash:newUpdateFolderPath
+                                                                                             expectedHash:newUpdateHash
+                                                                                                    error:&error]) {
+                                                                    
+                                                                    [self handleFailedDataIntegrityCheck:failCallback error:&error];
```
sergey-akhalkov: The same as above is here:
```objective-c
    if (!error) {
        error = [CodePushErrorUtils errorWithMessage:@"General integrity check error..."];
    }

    CPLog(error);

    failCallback(*error);
    return;
```

ios/CodePush/CodePushPackage.m
```diff
@@ -39,8 +39,22 @@ + (void)downloadAndReplaceCurrentBundle:(NSString *)remoteBundleUrl
...
+{
+    CPLog(@"The update contents failed the data integrity check.");
+    if (!error) {
+        failCallback([CodePushErrorUtils errorWithMessage:@"The update contents failed the data integrity check."]);
+        return;
+    }
+    
+    failCallback(*error);
+    return;
+}
```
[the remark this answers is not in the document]
sergey-akhalkov: Please also double check that `error:(NSError **)error` does not cause any issues, I mean here `if (!error) {` and here `error = [CodePushErrorUtils...`

ios/CodePush/CodePushPackage.m
```diff
@@ -230,19 +244,36 @@ + (void)downloadPackage:(NSDictionary *)updatePackage
                                                                 return;
                                                             }
                                                         }
+
+                                                        if (isDiffUpdate) {
+                                                            CPLog(@"Applying diff update.");
+                                                        } else {
+                                                            CPLog(@"Applying full update.");
+                                                        }
```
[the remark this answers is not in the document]
ruslan-bikkinin: @sergey-akhalkov ternary operator works only for variable assignment. The only way to refactor this code using ternary operator is:
```objective c
NSString * message = (isDiffUpdate) ? @"Applying diff update." : @"Applying full update.";
CPLog(message);
```

but i think elegance of this approach  is quiet arguable.

ios/CodePush/CodePush.h
```diff
@@ -176,10 +178,22 @@ failCallback:(void (^)(NSError *err))failCallback;
...
 + (NSString *)modifiedDateStringOfFileAtURL:(NSURL *)fileURL;
 
-+ (BOOL)verifyHashForDiffUpdate:(NSString *)finalUpdateFolder
++ (BOOL)isHashIgnoredFor:(NSString *) relativePath;
+
++ (BOOL)verifyFolderHash:(NSString *)finalUpdateFolder
                    expectedHash:(NSString *)expectedHash
                           error:(NSError **)error;
 
++ (NSString *)preparePublicKeyForDecoding:(NSString *)publicKeyString;
```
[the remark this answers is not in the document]
ruslan-bikkinin: @sergey-akhalkov thanks for suggestion, I'll do it!

ios/CodePush/CodePushPackage.m
```diff
@@ -230,19 +244,36 @@ + (void)downloadPackage:(NSDictionary *)updatePackage
...
+                                                        BOOL isSignatureVerificationEnabled = (publicKey != nil);
+                                                        if (isSignatureVerificationEnabled) {
+                                                            BOOL isSignatureValid = [CodePushUpdateUtils verifySignatureFor:newUpdateFolderPath
+                                                                                      withPublicKey:publicKey
+                                                                                              error:&error];
+                                                            if (!isSignatureValid) {
+                                                                [self handleFailedDataIntegrityCheck:failCallback error:&error];
                                                                 return;
+                                                            } else {
+                                                                CPLog(@"The update contents succueded the data integrity check.");
```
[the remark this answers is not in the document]
ruslan-bikkinin: @sergey-akhalkov, I'll fix it.

ios/CodePush/CodePushUpdateUtils.m
```diff
@@ -256,7 +280,87 @@ + (BOOL)verifyHashForDiffUpdate:(NSString *)finalUpdateFolder
...
+                     error:(NSError **)error
+{
+    NSLog(@"Verifying signature for folder path: %@", folderPath);
+    
+    NSString *publicKey = [self preparePublicKeyForDecoding: publicKeyString];
+    
+    NSError *signatureVerificationError;
+    NSString *signature = [self getSignatureFor: folderPath
+                                          error: &signatureVerificationError];
+    if(signatureVerificationError) {
```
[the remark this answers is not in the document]
ruslan-bikkinin: @sergey-akhalkov I'll fix it

ios/CodePush/CodePushPackage.m
ruslan-bikkinin: @sergey-akhalkov nice proposition, I'll do it!

ios/CodePush/CodePushPackage.m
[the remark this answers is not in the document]
ruslan-bikkinin: @sergey-akhalkov thx, I'll do it.

PR #980, opened 2017-08-29
Restore compatibility with RN v0.46

cooperka:
As [discussed in #895](https://github.com/Microsoft/react-native-code-push/pull/895#issuecomment-325700522), this change is fully forward- and backwards-compatible. It can be released as a patch (e.g. `v5.0.1` or `v4.1.1`, whatever you prefer).

This would be really helpful for people who continue to use RN v0.46 but want to receive updates from react-native-code-push.


pniko:
Thanks @cooperka!

PR #989, opened 2017-09-05
Readme file cleaning

buptkang:
This PR cleans the contents, resolve typos, rephrase certain sentences, remove unreachable links.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,11 +33,11 @@ This plugin provides client-side integration for the [CodePush service](http://c
 
 ## How does it work?
 
-A React Native app is composed of JavaScript files and any accompanying [images](https://facebook.github.io/react-native/docs/images.html#content), which are bundled together by the [packager](https://github.com/facebook/react-native/tree/master/packager) and distributed as part of a platform-specific binary (i.e. an `.ipa` or `.apk` file). Once the app is released, updating either the JavaScript code (e.g. making bug fixes, adding new features) or image assets, requires you to recompile and redistribute the entire binary, which of course, includes any review time associated with the store(s) you are publishing to.
+A React Native app is composed of JavaScript files and any accompanying [images](https://facebook.github.io/react-native/docs/images.html#content), which are bundled together by the packager and distributed as part of a platform-specific binary (i.e. an `.ipa` or `.apk` file). Once the app is released, updating either the JavaScript code (e.g. making bug fixes, adding new features) or image assets, requires you to recompile and redistribute the entire binary, which of course, includes any review time associated with the store(s) you are publishing to.
 
-The CodePush plugin helps get product improvements in front of your end users instantly, by keeping your JavaScript and images synchronized with updates you release to the CodePush server. This way, your app gets the benefits of an offline mobile experience, as well as the "web-like" agility of side-loading updates as soon as they are available. It's a win-win!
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -33,11 +33,11 @@ This plugin provides client-side integration for the [CodePush service](http://c
...
 
-The CodePush plugin helps get product improvements in front of your end users instantly, by keeping your JavaScript and images synchronized with updates you release to the CodePush server. This way, your app gets the benefits of an offline mobile experience, as well as the "web-like" agility of side-loading updates as soon as they are available. It's a win-win!
+The CodePush plugin helps product improvements in front of your end users instantly, by keeping your JavaScript and images synchronized with updating you release to the CodePush server. This way, your app gets benefits of an offline mobile experience, as well as the "web-like" agility of side-loading updates as soon as they are available. It's a win-win!
```
Zakeelm: would change this back to original wording

README.md
```diff
@@ -47,7 +47,7 @@ In order to ensure that your end users always have a functioning version of your
 - Android (4.1+)
 - Windows (UWP)
 
-We try our best to maintain backwards compatability of our plugin with previous versions of React Native, but due to the nature of the platform, and the existence of breaking changes between releases, it is possible that you need to use a specific version of the CodePush plugin in order to support the exact version of React Native you are using. The following table outlines which CodePush plugin versions officially support the respective React Native versions:
+We maintain backwards compatability of our plugin with previous versions of React Native. Due to the nature of the platform, and the existence of breaking changes between releases, it is possible that you need to use a specific version of the CodePush plugin in order to support the exact version of React Native you are using. The following table outlines which CodePush plugin versions officially support the respective React Native versions:
```
Zakeelm: I personally think the original flows a little better. Feels like more of a cohesive thought than with the two difference sentences.

README.md
```diff
@@ -111,12 +112,13 @@ Then continue with installing the native module
 
 ## Plugin Usage
 
-With the CodePush plugin downloaded and linked, and your app asking CodePush where to get the right JS bundle from, the only thing left is to add the necessary code to your app to control the following policies:
+After link the CodePush plugin, specify your app where CodePush can get the right JS bundle, you need to add necessary code in your app to tackle the below two issues:
```
Zakeelm: This sentence is a little confusing, I would change "After link" to "After linking" and do a few more structural changes to this sentence.

README.md
```diff
@@ -232,9 +234,9 @@ code-push release-react MyApp-Android android --targetBinaryVersion "~1.1.0"
 
 ```
 
-The CodePush client supports differential updates, so even though you are releasing your JS bundle and assets on every update, your end users will only actually download the files they need. The service handles this automatically so that you can focus on creating awesome apps and we can worry about optimizing end user downloads.
+The CodePush client supports differential updates, so even though you are releasing your JS bundle and assets on every update, you will only actually download the files you need. The service handles this optimizing downloads so that you can focus on creating awesome apps.
```
Zakeelm: Would keep as original. "your end users" and "they" are better words to describe the developer's users. "You" sounds more you're talking directly to the developer, which I don't think is the goal here

pniko:
Not quite sure this PR represents an improvement on our existing docs. Closing this for now.

Issue #999, opened 2017-09-14
[5.1.0-beta] `react-native link` gets stuck


fungilation:
### Steps to Reproduce
1. `react-native unlink react-native-code-push`
2. `react-native link react-native-code-push`
3. Doesn't return to cli, stuck on last line: "codepush.gradle" is already linked in the build definition

### Expected Behavior
Exit cleanly on `link`

### Actual Behavior
See 3 above

### Reproducible Demo
Output copied from terminal, with [STRINGS] replaced:
```
$ react-native unlink react-native-code-push
Scanning 611 folders for symlinks in [PROJECT]/node_modules (7ms)
rnpm-install info Unlinking react-native-code-push android dependency 
rnpm-install info Android module react-native-code-push has been successfully unlinked 
rnpm-install info Unlinking react-native-code-push ios dependency 
rnpm-install info iOS module react-native-code-push has been successfully unlinked 
$ react-native link react-native-code-push
Scanning 611 folders for symlinks in [PROJECT]/node_modules (8ms)
? What is your CodePush deployment key for Android (hit <ENTER> to ignore) [KEY]
rnpm-install info Linking react-native-code-push android dependency 
rnpm-install info Android module react-native-code-push has been successfully linked 
rnpm-install info Linking react-native-code-push ios dependency 
rnpm-install info iOS module react-native-code-push has been successfully linked 
"CodePush.h" header already imported.
"jsCodeLocation" already pointing to "[CodePush bundleURL]".
Could not determine correct xcode proj path to retrieve related plist file, there are multiple xcodeproj under the solution.
? What is your CodePush deployment key for iOS (hit <ENTER> to ignore) "getJSBundleFile" is already overridden
"codepush.gradle" is already linked in the build definition
```
... then stuck here, not returning to cli. Have to force quit with CTRL-C

### Environment

* react-native-code-push version: 5.1.0-beta
* react-native version: 0.48.3
* iOS/Android/Windows version: iOS 10.3.3
* Does this reproduce on a debug build or release build? Neither (cli)
* Does this reproduce on a simulator, or only on a physical device? Neither

ruslan-bikkinin:
@fungilation thanks for reaching us! Yeah, it seems to be a bug in postlinking scripts, i will be working on fix tomorrow. If you need additional assistance from our side with linking your application with react-native-code-push feel free to ask.

fungilation:
[...]

PR #1003, opened 2017-09-15
Fix incorrect link behaviour introduced with promise addition in ios postlink script

ruslan-bikkinin:
Due to promise usage in ios postlink script all postlink scripts should be promisified. This PR provide fix for this.

Also its related to #999

### DIFF
--- a/scripts/postlink/run.js
+++ b/scripts/postlink/run.js
@@ -1,2 +1,10 @@
-require("./ios/postlink");
-require("./android/postlink");
+var postlinks = [
+    require("./ios/postlink"),
+    require("./android/postlink")
+];
+
+//run them sequentially
+var result = postlinks.reduce((p, fn) => p.then(fn), Promise.resolve());
+result.catch((err) => {
+    console.error(err.message);
+}); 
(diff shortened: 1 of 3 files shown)


scripts/postlink/run.js
```diff
@@ -1,2 +1,10 @@
-require("./ios/postlink");
-require("./android/postlink");
\ No newline at end of file
+var postlinks = [
+    require("./ios/postlink"),
+    require("./android/postlink")
+];
+
+//run them sequentially
+var result = postlinks.reduce((p, fn) => p.then(fn), Promise.resolve());
```
sergey-akhalkov: Do we really need result variable here?
Would it probably better to do the following:
```js
postlinks.reduce((p, fn) => p.then(fn), Promise.resolve())
    .catch((err) => {
        console.error(err.message);
    });
```

ruslan-bikkinin: Not at all, I'll simplify it as you suggested, @sergey-akhalkov, thx!

fungilation:
Confirmed fix: https://github.com/Microsoft/react-native-code-push/issues/999#issuecomment-330289676

PR #1007, opened 2017-09-18
Fix dex index overflow exception

sergey-akhalkov:


android/app/src/main/java/com/microsoft/codepush/react/CodePushUpdateUtils.java
```diff
@@ -176,11 +180,13 @@ public static void verifyFolderHash(String folderPath, String expectedHash) {
...
         try {
-            final JWTVerifier verifier = new JWTVerifier(publicKey);
-            final Map<String, Object> claims = verifier.verify(jwt);
-            CodePushUtils.log("JWT verification succeeded:\n" + claims.toString());
-            return claims;
-        } catch (Exception e) {
+            SignedJWT signedJWT = SignedJWT.parse(jwt);
+            JWSVerifier verifier = new RSASSAVerifier((RSAPublicKey)publicKey);
+            if (signedJWT.verify(verifier)) {
+                return signedJWT.getJWTClaimsSet().getClaims();
```
max-mironov: Does it make sense to keep logging here to help devs to debug going further?

sergey-akhalkov: fixed, thanks!

android/app/src/main/java/com/microsoft/codepush/react/CodePushUpdateUtils.java
```diff
@@ -176,11 +180,14 @@ public static void verifyFolderHash(String folderPath, String expectedHash) {
...
-        } catch (Exception e) {
+            SignedJWT signedJWT = SignedJWT.parse(jwt);
+            JWSVerifier verifier = new RSASSAVerifier((RSAPublicKey)publicKey);
+            if (signedJWT.verify(verifier)) {
+                CodePushUtils.log("JWT verification succeeded:\n" + claims.toString());
+                return signedJWT.getJWTClaimsSet().getClaims();
+            }
+            return null;
+        } catch (Exception ex) {
             return null;
```
ruslan-bikkinin: I think we need to add exception logging here to be able determine cause of problem in future:
```java
CodePushUtils.log(e.getMessage());
CodePushUtils.log(e.getStackTrace().toString());
```

PR #1018, opened 2017-09-26
Fix proguard error in Android build

AndrewJack:
Couldn't build with v5.1.1 due to the introduction of the com.nimbusds.jose library which causes a proguard error.


ruslan-bikkinin:
Thanks @AndrewJack for quick fix!

PR #1035, opened 2017-10-06
Added requiresMainQueueSetup to prevent warning and future side-effects

djalmaaraujo:
New versions of react will require requiresMainQueueSetup method. Here's the warning:

```
Module CodePush requires main queue setup since it overrides `constantsToExport` but doesn't implement `requiresMainQueueSetup. In a future release React Native will default to initializing all native modules on a background thread unless explicitly opted-out of.
```

screenshot 2017-10-06 19 08 25

### DIFF
--- a/ios/CodePush/CodePush.m
+++ b/ios/CodePush/CodePush.m
@@ -302,6 +302,11 @@ static NSString *bundleResourceSubdirectory = nil;
             };
 };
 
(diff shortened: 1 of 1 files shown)


max-mironov:
@djalmaaraujo Thank you for contribution, looks good, merging it!

Issue #882, opened 2017-06-13
The codepush HOC doesn't copy static data


sergey-akhalkov:
Details in point 3 from https://github.com/Microsoft/react-native-code-push/issues/875#issuecomment-306776606

pniko:
Prioritized for investigation later on. Please take a look @sergey-akhalkov when you have some bandwidth.

max-mironov:
Closing the item since fix for it https://github.com/Microsoft/react-native-code-push/pull/1037 has been merged.

PR #1037, opened 2017-10-09
hoist non-react static methods in decorator

deecewan:
Simple fix for #882

I think that, given the decorator doesn't actually use any of these statics, this could probably be done without the library. However, the library is a tried and tested method, and is [the recommended way](https://reactjs.org/docs/higher-order-components.html#static-methods-must-be-copied-over) of doing this.


max-mironov:
Thanks for contribution @deecewan, really appreciate it! Looks good and merging it to Master.

PR #1047, opened 2017-10-13
Pass resolved install mode to status change callback

RoboTeddy:
If a mandatory upgrade is being downloaded (and an app restart is thus imminent), it can be important to notify the user that this is the case. Strictly speaking, this is possible given react-native-code-push as-is, but it would require maintaining your own version of `syncInternal`, which has some fairly complex internal logic.

Here's an example higher-order component that uses this extra sync information to display an upgrade progress bar, but only if it will cause the app to restart: https://gist.github.com/RoboTeddy/64ffa4078d2f8178a7cb1555a547bc53

I haven't included docs or anything in this PR -- wanted first to check whether the maintainers are open to this feature.

### DIFF
--- a/CodePush.js
+++ b/CodePush.js
@@ -263,7 +263,7 @@ const sync = (() => {
 
     if (syncInProgress) {
       typeof syncStatusCallbackWithTryCatch === "function"
-        ? syncStatusCallbackWithTryCatch(CodePush.SyncStatus.SYNC_IN_PROGRESS)
+        ? syncStatusCallbackWithTryCatch(CodePush.SyncStatus.SYNC_IN_PROGRESS, null)
         : log("Sync already in progress.");
       return Promise.resolve(CodePush.SyncStatus.SYNC_IN_PROGRESS);
     }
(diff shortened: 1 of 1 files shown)


ruslan-bikkinin:
Hi @RoboTeddy, to resolve the installation mode you can use the following approach:
1. call `checkForUpdate`
2. look if update is mandatory or not
3. apply your custom logic

Please let us know if you have any questions.

RoboTeddy:
@ruslan-bikkinin Am I right in thinking that in order to take that approach without having `checkForUpdate` be called twice (once by me and once by `sync`), it becomes necessary handle/maintain all the cases personally (such as failed installs and pending local updates)?

Thanks for responding earlier & I love code-push!

ruslan-bikkinin:
hi @RoboTeddy 
> once by me and once by syncInternal

No need to call sync method, you can use `remotePackage.download` and `remotePackage.install` after calling `checkForUpdate`.

ruslan-bikkinin:
@RoboTeddy Here is the main steps of update:

1. [Call `checkForUpdate`](https://github.com/Microsoft/react-native-code-push/blob/master/CodePush.js#L346)
2. [Download update](https://github.com/Microsoft/react-native-code-push/blob/master/CodePush.js#L350)
3. [Install update](https://github.com/Microsoft/react-native-code-push/blob/master/CodePush.js#L356)

PR #1052, opened 2017-10-22
fix(Documentation): correct url in AppDelegate.m for RN 0.49 and up

itchingpixels:


max-mironov:
@itchingpixels - thanks for pointing, appreciate it!

Issue #970, opened 2017-08-18
Android: getJSBundleFile() Method does not override method from it's super class. 


joshuapinter:
Hi There, I'm getting this error but I have it in the correct place and the correct versions: 

React Native: 0.46.4
React Native Code Push: 4.1.0-beta

```java
public class ReactNative extends Activity implements DefaultHardwareBackBtnHandler, ReactApplication {

    private static final int OVERLAY_PERMISSION_REQ_CODE = 1;
    private ReactRootView mReactRootView;
    private ReactInstanceManager mReactInstanceManager;

    @Override // <-- Method does not override method from it's super class.
    public String getJSBundleFile() {
        return CodePush.getJSBundleFile();
    }

    @Override
    protected void onCreate(Bundle savedInstanceState) {
        // ...
    }
```

I think this is the reason why my app sees that there is an update, downloads the update, restarts the app but nothing actually changes. I imagine it's because it's not getting the updated JS bundle.

Help please! :)

max-mironov:
Hi @joshuapinter,

So I'm seeing that you are using not the recommended way of integration CodePush within your app (well I mean here that not the way we suppose to use it in our docs:)  ).
That probably could be an issue. Can you try to move away `getJSBundleFile` method from `ReactNative ` class to `MainApplication.java` class. Also can you please take a look at the example with similar integration logic that promise to work well: https://github.com/youngcube/AwesomeProject/blob/master/android/app/src/main/java/com/awesomeproject/MainActivity.java

If it doesn't help could you please try to repoduce your issue within this sample so we could also try to build it on our side and see if it works?

Please let us know if it helps.

joshuapinter:
HI @max-mironov. Thanks for getting back to me so quickly!

This is an existing Android application that we've integrated React Native into, so there is no MainApplication.java class. But we do have an application file that I can try moving it to. Let me try that...

joshuapinter:
Still the same error with trying to Override the getJSBundleFile() method. Where does this method live that we're trying to override it? I'll try loading up your example app and see what I can find out there.

max-mironov:
@joshuapinter - hi again.

So regarding your questions: 
[...]

PR #1098, opened 2017-12-05
[docs] Add information about how to configure CodePush for applications using ReactInstanceManager.builder() to initialize RN

ruslan-bikkinin:
RN beginners who is coming from [here](https://facebook.github.io/react-native/docs/integration-with-existing-apps.html) couldn't find information how to configure CodePush for existing android apps. This PR intends to fix this.
Also this PR relates to https://github.com/Microsoft/react-native-code-push/issues/970#issuecomment-326995052

### DIFF
--- a/docs/setup-android.md
+++ b/docs/setup-android.md
@@ -66,6 +66,8 @@ After installing the plugin and syncing your Android Studio project with Gradle,
 
 **For React Native >= v0.29**
 
+If you are using React Native for the whole application please do the following:
+
 Update the `MainApplication.java` file to use CodePush via the following changes:
 
 ```java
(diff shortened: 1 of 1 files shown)


docs/setup-android.md
```diff
@@ -107,6 +109,40 @@ protected String getJSMainModuleName() {
 }
 ```
 
+If you are integrating React Native into existing application please do the following:
+
+Update `MyReactActivity.java` (it could be named differently in you app) file to use CodePush via the following changes:
```
max-mironov: nit: mistype, `differently in your app`

docs/setup-android.md
```diff
@@ -66,6 +66,8 @@ After installing the plugin and syncing your Android Studio project with Gradle,
 
 **For React Native >= v0.29**
 
+If you are using React Native for the whole application please do the following:
```
max-mironov: Not sure if the word `whole application` fits good here. Probably we could rephrase like `If you are integrating Code Push into React Native application please do the following steps:`

docs/setup-android.md
```diff
@@ -107,6 +109,40 @@ protected String getJSMainModuleName() {
 }
 ```
 
+If you are integrating React Native into existing application please do the following:
```
max-mironov: I'd prefer to say here `If you are integrating React Native into existing native application please do the following steps:`

PR #1105, opened 2017-12-12
Changed protected->public for getJSMainModuleName

Titozzz:
Fixes errors when implementing on android because getJSMainModuleName is now public


Titozzz:
Nevermind, this is true with https://github.com/wix/react-native-navigation but not original react-native

Issue #1073, opened 2017-11-09
Why android don't have `setDeploymentKey` method?


hwangjr:
I just found there is a method `setDeploymentKey` in javescript and ios method, but i can't found it in android. can you support it next version ?

>- __(void)setDeploymentKey:(NSString \*)deploymentKey__ - Sets the deployment key that the app should use when querying for updates. This is a dynamic alternative to setting the deployment key in your `Info.plist` and/or specifying a deployment key in JS when calling `checkForUpdate` or `sync`.

hwangjr:
i want to change the deployment key fetched from server, so maybe i just change the `mDeploymentKey` value in [react-native-code-push/CodePush.java at master · Microsoft/react-native-code-push](https://github.com/Microsoft/react-native-code-push/blob/master/android/app/src/main/java/com/microsoft/codepush/react/CodePush.java)

ruslan-bikkinin:
@hwangjr thanks for submitting the issue. 

> i want to change the deployment key fetched from server, so maybe i just change the mDeploymentKey value in react-native-code-push/CodePush.java at master · Microsoft/react-native-code-push

You can not do that because `mDeploymentKey` has private access. You can change deployment key on client side using `deploymentKey` option. Please refer to documentation [section](https://github.com/Microsoft/react-native-code-push/blob/master/docs/api-js.md#codepushoptions) for more information.

In general I think, yes, android SDK lacks of this method and it should be added for consistency purpose. But now we are highly backlogged and can't start working on it right now. We could pick it up when we'll got free time. So if you want this feature be shipped as soon as possible you could make contribution, we will appreciate your help.

hwangjr:
i just recreate the CodePush while i found the code push key has changed.

PR #1115, opened 2017-12-22
Allow setting the deploymentKey on Android

christophermark:
This adds Android support for setting the deployment key. A similar method aleady exists for iOS:
 https://github.com/Microsoft/react-native-code-push/blob/751bf02836f6af652cbd2d9f5a499444a5964a00/ios/CodePush/CodePush.m#L207-L210

Addresses #1073


ruslan-bikkinin:
Hi @charlieschwabacher and thanks for your contribution. Please update [docs](https://github.com/Microsoft/react-native-code-push/blob/master/docs/api-android.md) too and I'll merge it after that.

PR #1116, opened 2017-12-24
Adding check for null ref before accessing array index

RobertPaul01:
When running `react-native link`, I ran into a null reference error. To fix this error, I had to correct this script locally. I think this patch should be included to prevent other people from having this issue.


RobertPaul01:
Yeah I think adding logging would help here. I added a log in my PR. If there's anything additional you want me to log here, I can add it to my PR.

max-mironov:
Thanks @RobertPaul01 for quick turnaround, we are good with merging this!

PR #1174, opened 2018-02-01
[Brownfield] Add tool for running coverage

ruslan-bikkinin:
This PR adds scripts for running coverage. To use it run in terminal:
* `sh runCoverage.sh` - for Unix machines
* `runCoverage` - for Windows machines


sergey-akhalkov:
Ah I got it, we need to add new app because of the gradle and buildTools versions, so LGTM then, but still need to remove some `BUCK` and etc. files that are not really needed.

PR #1182, opened 2018-02-05
[Brownfield] Tests refactoring

annakocheshkova:
Created a sharable test utils module. Renamed tests.


android/app/src/main/java/com/microsoft/codepush/common/utils/FileUtils.java
```diff
@@ -189,6 +189,7 @@ public static String readFileToString(String filePath) throws IOException {
             Exception e = finalizeResources(
                     Arrays.asList(reader, fileInputStream),
                     "Error closing IO resources when reading file.");
+
```
sergey-akhalkov: empty line - please add a comment after that or remove it.

android/app/src/main/java/com/microsoft/codepush/common/utils/PlatformUtils.java
```diff
@@ -2,14 +2,15 @@
 
 /**
  * Interface describing the methods that should be implemented in platform-specific instances of utils.
+ * It can be implemented via platform-specific singleton.
```
sergey-akhalkov: Don't sure if we need this comment, is it some kind of todo?

annakocheshkova: @sergey-akhalkov This is just a note on interface use case

android/app/src/main/java/com/microsoft/codepush/common/utils/FileUtils.java
annakocheshkova: Removed.

PR #1183, opened 2018-02-05
[Brownfield] Restart manager

annakocheshkova:
Created restart manager and 100% test coverage.


android/app/src/main/java/com/microsoft/codepush/common/managers/CodePushRestartManager.java
```diff
@@ -0,0 +1,100 @@
...
+     */
+    public CodePushRestartManager(CodePushRestartListener codePushRestartListener) {
+        mRestartListener = codePushRestartListener;
+    }
+
+    /**
+     * Allows the manager to perform restarts and performs them if there are pending.
+     */
+    public void allowRestarts() {
+        AppCenterLog.logAssert(LOG_TAG, "Re-allowing restarts");
```
ruslan-bikkinin: Why do you use `logAssert` instead of `info` or just `log`?

annakocheshkova: @ruslan-bikkinin Because we are using it already in other project, look at the UpdateManager, for instance

sergey-akhalkov: I believe we should not use `logAssert ` method, please see https://github.com/Microsoft/AppCenter-SDK-Android/search?utf8=%E2%9C%93&q=AppCenterLog.logAssert&type=

You should use `info`, `debug` and etc. Please also investigated AppCenter-SDK-Android repo for the same situations.

android/app/src/main/java/com/microsoft/codepush/common/managers/CodePushRestartManager.java
```diff
@@ -0,0 +1,100 @@
...
+    public boolean restartApp(boolean onlyIfUpdateIsPending) {
+        if (mRestartInProgress) {
+            AppCenterLog.logAssert(LOG_TAG, "Restart request queued until the current restart is completed");
+            mRestartQueue.add(onlyIfUpdateIsPending);
+        } else if (!mAllowed) {
+            AppCenterLog.logAssert(LOG_TAG, "Restart request queued until restarts are re-allowed");
+            mRestartQueue.add(onlyIfUpdateIsPending);
+        } else {
+            mRestartInProgress = true;
+            if (mRestartListener.onRestartReady(onlyIfUpdateIsPending)) {
```
sergey-akhalkov: We should notify React Native instance that we are requesting app restart. Another words React Native client should be subscribed on App Restart events. I would prefer to change `onRestartReady` method name to `onRestart` to avoid of confusing. Please let me know if I've missed something.

annakocheshkova: @sergey-akhalkov This should happen in s callback. Didn't quite get why we should rename the method itself

PR #1184, opened 2018-02-05
[Brownfield] Utils refactoring

annakocheshkova:
* Rewrote all utils to match sigleton pattern;
* Rewrote project to implement dependency injection principles;
* Rewrote tests accordingly.


Mommabearoffive:
Ok

PR #1186, opened 2018-02-06
[Brownfield] Cover CodePushUtils and StringUtils with tests, minor fixes and improvements

ruslan-bikkinin:
This PR includes the following changes:
* Covered `CodePushUtils` and `StringUtils` with tests
* Re-ordered `convertStringToObject` and `convertObjectToJsonString` methods for consistency
* Remove redundant `android/app/src/test/java/com/microsoft/codepush/common/managers/UpdateManagerUnitTests.java`
* Minor improvements


android/app/src/androidTest/java/com/microsoft/codepush/common/utils/UtilsAndroidTest.java
```diff
@@ -0,0 +1,183 @@
...
+        FileUtils fileUtils = FileUtils.getInstance();
+        mUtils = CodePushUtils.getInstance(fileUtils);
+    }
+
+    /**
+     * Tests getting json object from correct json file.
+     */
+    @Test
+    public void testGetJsonObjectFromCorrectJsonFile() throws Exception {
+        String inputJsonString = "{\"key\":\"value\"}";
```
annakocheshkova: I think this and below should be final constants, as done in other files.

android/app/src/androidTest/java/com/microsoft/codepush/common/utils/UtilsAndroidTest.java
```diff
@@ -0,0 +1,183 @@
...
+        File jsonFile = getRealNamedFileWithContent("json.json", inputJsonString);
+        JSONObject result = mUtils.getJsonObjectFromFile(jsonFile.getAbsolutePath());
+        Assert.assertEquals(result.toString(), inputJsonString);
+    }
+
+    /**
+     * Tests getting json object from malformed json file.
```
annakocheshkova: I think it's better make javadocs as specific as possible, as in other test files. For example `{@link CodePushUtils#getJsonObjectFromFile()} should throw a {@link CodePushMalformedDataException} if a {@link JsonException} is thrown during converting.`
Please fix it there and below.

android/app/src/test/java/com/microsoft/codepush/common/utils/StringUtilsUnitTest.java
```diff
@@ -0,0 +1,24 @@
+package com.microsoft.codepush.common.utils;
+
+import org.junit.Test;
+
+import static com.microsoft.codepush.common.utils.StringUtils.isNullOrEmpty;
```
MatkovIvan: There is default API for that. See [TextUtils.isEmpty](https://developer.android.com/reference/android/text/TextUtils.html#isEmpty(java.lang.CharSequence))

android/app/src/sharedTest/java/com/microsoft/codepush/common/testutils/CommonFileTestUtils.java
```diff
@@ -30,13 +31,39 @@ public static File getRealTestFolder() {
...
+    /**
+     * Creates a real (not mocked) file for testing.
+     *
+     * @param fileName name for the file.
+     * @param content  content of file.
+     * @return real test file.
+     */
+    public static File getRealNamedFileWithContent(String fileName, String content) throws IOException {
+        File file = getRealNamedFile(fileName);
+        FileWriter fw = new FileWriter(file);
```
annakocheshkova: Please get rid of the variable names like `fw` - better name it `fileWriter`.

android/app/src/test/java/com/microsoft/codepush/common/utils/FinalizeResourcesTest.java
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    @Parameters(name = "{index}: finalizeResources({0}, null) = {1}")
+    public static Collection<Object[]> data() {
+
+        IOException exception1 = new IOException();
+        IOException exception2 = new IOException();
+
+        return Arrays.asList(new Object[][]{
+                {Arrays.asList(createGoodResource()), null},
```
annakocheshkova: Can't we just use `mock(InputStream.class)` instead of the `createGoodResource`?

android/app/src/test/java/com/microsoft/codepush/common/utils/StringUtilsUnitTest.java
ruslan-bikkinin: @MatkovIvan Thanks for pointing on that!

android/app/src/androidTest/java/com/microsoft/codepush/common/utils/UtilsAndroidTest.java
ruslan-bikkinin: @annakocheshkova My opinion here is that test should be as obvious as possible and this class doesn't contain complex test case/mocks so no need to overload test class.

android/app/src/androidTest/java/com/microsoft/codepush/common/utils/UtilsAndroidTest.java
ruslan-bikkinin: @annakocheshkova agree with you, i'll fix that.

android/app/src/sharedTest/java/com/microsoft/codepush/common/testutils/CommonFileTestUtils.java
ruslan-bikkinin: @annakocheshkova got it.

android/app/src/test/java/com/microsoft/codepush/common/utils/FinalizeResourcesTest.java
ruslan-bikkinin: @annakocheshkova yes we can, but it just worsen understanding test scenario.

annakocheshkova: @ruslan-bikkinin My opinion is use mock where no real object needed. Instead it turns out that you are creating more complex logic than needed.

ruslan-bikkinin: @annakocheshkova Well I believe we need the third one to decide. @sergey-akhalkov what do you think on this?

sergey-akhalkov: @ruslan-bikkinin, I would prefer to use `mock(InputStream.class)` instead of `createGoodResource `

PR #1188, opened 2018-02-07
[Brownfield] Acquisition manager

annakocheshkova:
Implemented `AcquisitionManager`.
* Several AsyncTasks has been separated, rewritten and became **`Jobs`**. 
* A `BaseJob` class has been introduced to contain some common logic.

>Job represents a background task. This is a common naming for tasks that, for example, perform a query to server.

* Several data contracts have been introduced, also, to work with new async tasks.
* Several exceptions have been introduced for work with new functions.


android/app/src/main/java/com/microsoft/codepush/common/connection/QueryUpdateJob.java
```diff
@@ -0,0 +1,89 @@
...
+        Scanner scanner = null;
+        HttpURLConnection connection;
+        try {
+            connection = createConnection(mRequestUrl);
+        } catch (IOException e) {
+
+            /* We can't throw custom errors from this function, so any error will be passed to the result. */
+            return CodePushUpdateResponse.createFailed(new CodePushQueryUpdateException(e));
+        }
+        try {
```
ruslan-bikkinin: Do we really need to begin `try {` from this line? May be we should move it to 72?

android/app/src/main/java/com/microsoft/codepush/common/connection/ReportStatusJob.java
```diff
@@ -0,0 +1,111 @@
...
+     * @param json       report as json string.
+     * @param reportType type of the report as listed in {@link CodePushReportStatusException.ReportType}.
+     */
+    public void setParameters(String requestUrl, String json, CodePushReportStatusException.ReportType reportType) {
+        mRequestUrl = requestUrl;
+        mJson = json;
+        mReportType = reportType;
+    }
+
+    @Override protected CodePushReportStatusResult doInBackground(Void... voids) {
```
ruslan-bikkinin: Please move `@Override` to the line above

android/app/src/main/java/com/microsoft/codepush/common/connection/QueryUpdateJob.java
annakocheshkova: @ruslan-bikkinin Nope. The 62 line is already throwing an exception.

ruslan-bikkinin: @annakocheshkova ok.

android/app/src/main/java/com/microsoft/codepush/common/connection/BaseJob.java
```diff
@@ -0,0 +1,42 @@
...
+import java.io.IOException;
+import java.net.HttpURLConnection;
+import java.net.URL;
+
+/**
+ * This class represent basic code push job.
+ *
+ * @param <T> type of the returned parameter.
+ */
+public abstract class BaseJob<T> extends AsyncTask<Void, Void, T> {
```
sergey-akhalkov: `BaseJob` name looks too generic, what about `BaseHtttJob` or so?

annakocheshkova: Can be

android/app/src/main/java/com/microsoft/codepush/common/datacontracts/CodePushReportStatusResult.java
```diff
@@ -0,0 +1,88 @@
...
+
+    /**
+     * An exception that has occurred when sending the report, <code>null</code> if the report has been sent successfully.
+     */
+    private CodePushReportStatusException codePushReportStatusException;
+
+    /**
+     * The result string from server.
+     */
+    private String result;
```
MatkovIvan: `mResult`

android/app/src/main/java/com/microsoft/codepush/common/connection/QueryUpdateJob.java
```diff
@@ -0,0 +1,90 @@
...
+import java.util.Scanner;
+
+import javax.net.ssl.HttpsURLConnection;
+
+import static com.microsoft.codepush.common.CodePush.LOG_TAG;
+
+/**
+ * Performs sending status reports to server.
+ */
+public class QueryUpdateJob extends BaseHttpJob<CodePushUpdateResponse> {
```
sergey-akhalkov: I would prefer to use the same terms as we've been using before, instead of `QueryUpdateJob` we can use `CheckForUpdateJob`, `RepostStatusDownloadJob`, `RepostStatusDeployJob`, it looks consistent and more clear in terms of existing ecosystem.

android/app/src/main/java/com/microsoft/codepush/common/managers/CodePushAcquisitionManager.java
```diff
@@ -0,0 +1,198 @@
...
+     */
+    public void reportStatusDeploy(CodePushDeploymentStatusReport deploymentStatusReport) throws CodePushReportStatusException {
+
+        /* TODO: Consider moving the following logic to some other place or removing it at all if useless. */
+        deploymentStatusReport.setClientUniqueId(mClientUniqueId);
+        deploymentStatusReport.setDeploymentKey(mDeploymentKey);
+        deploymentStatusReport.setAppVersion(deploymentStatusReport.getLocalPackage() != null ? deploymentStatusReport.getLocalPackage().getAppVersion() : mAppVersion);
+        deploymentStatusReport.setAppVersion(deploymentStatusReport.getLocalPackage() != null ? deploymentStatusReport.getLocalPackage().getLabel() : null);
+        final String requestUrl = mServerUrl + REPORT_DEPLOYMENT_STATUS_ENDPOINT;
+        if (deploymentStatusReport.getLocalPackage() != null) {
```
sergey-akhalkov: Do we need this condition? Further we are manipulating with `deploymentStatusReport.getStatus()` only.

android/app/src/main/java/com/microsoft/codepush/common/connection/QueryUpdateJob.java
annakocheshkova: @sergey-akhalkov We have only one `ReportStatusJob` instead of 2.

android/app/src/main/java/com/microsoft/codepush/common/datacontracts/CodePushReportStatusResult.java
annakocheshkova: @MatkovIvan In data contracts classes we do not use this convention as it is in AppCenter

PR #1194, opened 2018-02-09
[Brownfield] Exceptions

annakocheshkova:
Replaced some usages of `AppCenterLog.error()` with real exceptions.


android/app/src/androidTest/java/com/microsoft/codepush/common/utils/FileAndroidTests.java
```diff
@@ -234,4 +236,41 @@ public void tearDown() throws Exception {
...
+        doReturn(true).when(testFile).isDirectory();
+        mFileUtils.deleteFileOrFolderSilently(testFile);
+    }
+
+    /**
+     * {@link FileUtils#deleteFileOrFolderSilently(File)} should throw a {@link CodePushFileException}
+     * if <code>delete</code> returned <code>null</code>.
+     */
+    @Test(expected = CodePushFileException.class)
+    public void tests2() throws Exception {
```
ruslan-bikkinin: Please use more detailed name.

android/app/src/androidTest/java/com/microsoft/codepush/common/utils/FileAndroidTests.java
```diff
@@ -234,4 +236,41 @@ public void tearDown() throws Exception {
         File testFolder = getTestingDirectory();
         testFolder.delete();
     }
+
+    /**
+     * {@link FileUtils#deleteFileOrFolderSilently(File)} should throw a {@link CodePushFileException}
+     * if <code>listFiles</code> returned <code>null</code>.
+     */
+    @Test(expected = CodePushFileException.class)
+    public void tests() throws Exception {
```
ruslan-bikkinin: Please use more detailed name.

android/app/src/androidTest/java/com/microsoft/codepush/common/utils/FileAndroidTests.java
```diff
@@ -234,4 +236,41 @@ public void tearDown() throws Exception {
...
+        doReturn(false).when(testFile).delete();
+        mFileUtils.deleteFileOrFolderSilently(testFile);
+    }
+
+    /**
+     * {@link FileUtils#deleteFileOrFolderSilently(File)} should throw a {@link CodePushFileException}
+     * if <code>delete</code> on child file returned <code>null</code>.
+     */
+    @Test(expected = CodePushFileException.class)
+    public void tests3() throws Exception {
```
ruslan-bikkinin: Please use more detailed name.

android/app/src/main/java/com/microsoft/codepush/common/exceptions/CodePushFileException.java
```diff
@@ -0,0 +1,28 @@
+package com.microsoft.codepush.common.exceptions;
+
+/**
+ * Exception class for throwing exceptions occurred when working with files.
+ */
+public class CodePushFileException extends Exception {
```
ruslan-bikkinin: What is the reason for adding FileException if we already have IOException?

annakocheshkova: @ruslan-bikkinin Ibelieve its no good to throw an `IOException`, we should throw our own exceptions only.

annakocheshkova: I think I would change that in other places, too

android/app/src/main/java/com/microsoft/codepush/common/managers/CodePushUpdateManager.java
```diff
@@ -395,10 +395,10 @@ public void unzipPackage(File downloadFile) throws CodePushUnzipException {
         String unzippedFolderPath = getUnzippedFolderPath();
         try {
             mFileUtils.unzipFile(downloadFile, new File(unzippedFolderPath));
+            mFileUtils.deleteFileOrFolderSilently(downloadFile);
```
ruslan-bikkinin: What is the reason for moving it here? This method isn't throw no more exception anyway.

annakocheshkova: @ruslan-bikkinin now it throws an `IOException`

android/app/src/main/java/com/microsoft/codepush/common/utils/FileUtils.java
```diff
@@ -137,27 +135,26 @@ public void deleteDirectoryAtPath(String directoryPath) throws IOException {
      * @param file path to file/folder to be deleted.
      */
     @SuppressWarnings("WeakerAccess")
-    public void deleteFileOrFolderSilently(File file) {
+    public void deleteFileOrFolderSilently(File file) throws IOException {
```
ruslan-bikkinin: This method should't throw any errors as said in his description , so please remove this change and place TODO comment here in order to don't forget to return here later to add `trackException` instead.

PR #1204, opened 2018-02-15
[Brownfield] Crashes support

annakocheshkova:
Added `AppCenter Crashes` support w\o breaking compatibility. Introduced several new constructors with `Application` instead of `Context` and additional `appSecret` parameter allowing to configure `AppCenter` and track exceptions in `CodePushLogUtils`.


android/app/src/main/java/com/microsoft/codepush/common/core/CodePushBaseCore.java
```diff
@@ -183,7 +183,9 @@
      * @param publicKeyProvider     instance of {@link CodePushPublicKeyProvider}.
      * @param appEntryPointProvider instance of {@link CodePushAppEntryPointProvider}.
      * @param platformUtils         instance of {@link CodePushPlatformUtils}.
-     * @throws CodePushInitializeException if error occurred during the initialization.
+     * @param application           application instance.
```
ruslan-bikkinin: Please add note that this parameter is needed for Crashes only

android/app/src/main/java/com/microsoft/codepush/common/core/CodePushBaseCore.java
```diff
@@ -183,7 +183,9 @@
...
+     * @param appSecret             the value of app secret from AppCenter portal to configure {@link Crashes} sdk.
```
ruslan-bikkinin: Please add note that if its value equals null, exceptions will not be tracked.

Issue #1123, opened 2017-12-28
Local SSZipArchive Build Error


LiuYibin:
You import the  source code of lib "SSZipArchive"  into your node package manually, but maybe you modify the code or the lib's version you are using is an version with build error.
when I import SSZipArchive in a separated Xcode project, there is no error when building.

LiuYibin:
I import youor lib not via react-native-code-push, and no error any more.

ruslan-bikkinin:
Hi @LiuYibin and thanks for reaching us. Could you please share with us detailed info about build error that was occurred during the building of your app (text output, screenshots)?

LiuYibin:
<img width="1390" alt="wx20171228-171115 2x" src="https://user-images.githubusercontent.com/7300267/34406150-5e628e8e-ebf2-11e7-8e2d-0186e7386f65.png">
<img width="1400" alt="wx20171228-171142 2x" src="https://user-images.githubusercontent.com/7300267/34406151-5e9f2240-ebf2-11e7-9b08-4ec0b85fa246.png">
<img width="1400" alt="wx20171228-171154 2x" src="https://user-images.githubusercontent.com/7300267/34406152-5edc8fa4-ebf2-11e7-829a-e62b0947b823.png">
<img width="1279" alt="wx20171228-171236 2x" src="https://user-images.githubusercontent.com/7300267/34406153-5f1e1460-ebf2-11e7-98cf-bd545339ae6b.png">

as the pictures above, The error will disappear after I delete the code in your CodePush.podspec file and import the SSZipArchive lib dividually.
`s.subspec 'SSZipArchive' do |ss|
    ss.source_files = 'ios/CodePush/SSZipArchive/*.{h,m}', 'ios/CodePush/SSZipArchive/aes/*.{h,c}', 'ios/CodePush/SSZipArchive/minizip/*.{h,c}'
    ss.private_header_files = 'ios/CodePush/SSZipArchive/*.h', 'ios/CodePush/SSZipArchive/aes/*.h', 'ios/CodePush/SSZipArchive/minizip/*.h'
  end`

ruslan-bikkinin:
@LiuYibin so you are trying to add `react-native-code-push` to your app via cocoa pods but your app has it's own `SSZipArchive` lib installed before? Please correct me if I missed something.

LiuYibin:
[...]

Issue #1114, opened 2017-12-21
duplicate React 


kdong007:
Thanks so much for filing an issue or feature request! Please fill out the following (wherever relevant):

### Steps to Reproduce
1. yarn add react-native-code-push 
2. react-native link
3. pod install
4. yarn start

### Expected Behavior
node packager runs without warning/errors

### Actual Behavior
```
Loading dependency graph...jest-haste-map: @providesModule naming collision:
  Duplicate module name: verifyPropTypes
  Paths: /Users/kdong/react-native-app/ios/Pods/React/Libraries/ReactIOS/verifyPropTypes.js collides with /Users/kdong/react-native-app/node_modules/react-native/Libraries/ReactNative/verifyPropTypes.js

This warning is caused by a @providesModule declaration with the same name across two different files.
jest-haste-map: @providesModule naming collision:
  Duplicate module name: PerformanceOverlay
  Paths: /Users/kdong/react-native-app/ios/Pods/React/Libraries/Inspector/PerformanceOverlay.js collides with /Users/kdong/react-native-app/node_modules/react-native/Libraries/Inspector/PerformanceOverlay.js

This warning is caused by a @providesModule declaration with the same name across two different files.
jest-haste-map: @providesModule naming collision:
  Duplicate module name: TransformPropTypes
  Paths: /Users/kdong/react-native-app/ios/Pods/React/Libraries/StyleSheet/TransformPropTypes.js collides with /Users/kdong/react-native-app/node_modules/react-native/Libraries/StyleSheet/TransformPropTypes.js

This warning is caused by a @providesModule declaration with the same name across two different files.
jest-haste-map: @providesModule naming collision:
  Duplicate module name: flattenStyle
  Paths: /Users/kdong/react-native-app/ios/Pods/React/Libraries/StyleSheet/flattenStyle.js collides with /Users/kdong/react-native-app/node_modules/react-native/Libraries/StyleSheet/flattenStyle.js

This warning is caused by a @providesModule declaration with the same name across two different files.
jest-haste-map: @providesModule naming collision:
  Duplicate module name: NetInfo
  Paths: /Users/kdong/react-native-app/ios/Pods/React/Libraries/Network/NetInfo.js collides with /Users/kdong/react-native-app/node_modules/react-native/Libraries/Network/NetInfo.js

This warning is caused by a @providesModule declaration with the same name across two different files.
jest-haste-map: @providesModule naming collision:
  Duplicate module name: bezier
[...]

Issue #241, opened 2016-03-11
duplicate symbol _OBJC_IVAR_$_SSZipArchive._zip in:


chirag04:
I get the above issue when configuring codepush. We are also using https://github.com/plrthink/react-native-zip-archive which also rely on ``SSZipArchive``.

cc @lostintangent

geof90:
Hi @chirag04 (and anyone else who could be hitting this issue),

The full workaround is to
1. Remove SSZipArchive from the CodePush library project
![screen shot 2016-03-11 at 6 22 53 pm](https://cloud.githubusercontent.com/assets/8598682/13720154/5e3cd5fa-e7b6-11e5-8edf-c609b6ddf9bb.png)
2. Update the Header Search Paths of the CodePush library project to point at the version of SSZipArchive that is bundled together with the react-native-zip-archive plugin.
![screen shot 2016-03-11 at 6 24 21 pm](https://cloud.githubusercontent.com/assets/8598682/13720163/83a7d6dc-e7b6-11e5-8286-92b8e4611495.png)

chirag04:
@geof90 well this work but everytime i npm install i would need to delete the folder and set the path again. Can we do one time setup?

geof90:
@chirag04 Sorry for the late reply, I somehow missed this. The best way I see to prevent having the manual deletion step is to make a fork of our project, delete the folder in that fork, and then [specify your fork's github URL as a dependency in your package.json](https://docs.npmjs.com/files/package.json#git-urls-as-dependencies).

cancan101:
[This solution](https://github.com/Microsoft/react-native-code-push/issues/241#issuecomment-195639258) seems less than ideal. Why does this project need to vendor `SSZipArchive`? Is there no way to handle this dependency in a smarter fashion that does not conflict with other npm packages?

lostintangent:
@cancan101 It's definitely non-ideal. Unfortunately, `react-native-zip-archive` also bundles a copy of `SSZipArchive` with it, so we couldn't easily share a common dependency from CocoaPods/etc. with that package, if both it and CodePush were installed.

Which NPM package is creating the conflict for you? Are you also trying to use `react-native-zip-archive`? I'll give some thought to this issue today since I agree it needs to be addressed.

cancan101:
Yes, am trying to use `react-native-zip-archive` and I do see they are vendoring as well. I suppose you are both doing so to solve the problem of dependency management in the case where the user is not using Cocoapods. 

[...]

PR #1210, opened 2018-02-20
[CocoaPods] Resolve dependencies at parent level

rob-keepsafe:
## Issue

Subspecs should not be used as a means to add/remove 3rd party code that CodePush depends upon. It's meant only to add/remove features within a given library itself.

When other projects also rely upon those 3rd party libraries, subspecs break things because they don't allow you to do version resolving and dependency management like CocoaPods does by default.

## Changes

Instead of using subspecs that reference some local files in the repo, we let CocoaPods resolve dependencies based upon versions. Otherwise, you end up with very opaque linking errors trying to add/remove subspecs.

I would recommend using the same approach of using dependency resolving for all other installation methods like Carthage and leave manual installation as an "at your own risk" like it's meant to be.

## Update Instructions 

Updaters will just need to:
 - Update any of their Podfile's to remove references to the subspecs
 - Run `$ bundle exec pod update CodePush`

### DIFF
--- a/CodePush.podspec
+++ b/CodePush.podspec
@@ -3,7 +3,6 @@ require 'json'
 package = JSON.parse(File.read(File.join(__dir__, 'package.json')))
 
 Pod::Spec.new do |s|
-
   s.name           = 'CodePush'
   s.version        = package['version'].gsub(/v|-beta/, '')
   s.summary        = package['description']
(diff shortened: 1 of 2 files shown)


mistikk:
This pr broke these issues again;
#1378
#1236
#1123
#1114
#241

Yes subspecs should not be used as a means to add/remove 3rd party code, but we should produce a new method instead of that.
@rob-keepsafe

PR #1232, opened 2018-03-16
Place the Install button on the right side of the Update Dialog

Hunter-Dolan:
Hello!

We use Code Push's update dialog on several of my projects and have had great results so far. One minor point of annoyance is that the Install button is to the left of the Update Dialog when the "Ignore" option is enabled.

On most alerts, it seems common practice to keep the cancel/ignore button to the left and the "confirm" button to the right (Enable Push Notifications, Location Services, etc). During beta testing we saw several people accidentally ignore the update because of this.

This pull request addresses this issue by swapping the order of the two buttons.


Hunter-Dolan:
🎉

PR #1237, opened 2018-03-23
Change download to use identity Accept-Encoding

datvong-wm:
If the server returns gzip encoding, the size check will fail because getContentLength() could potentially return -1 for gzipped content.

### DIFF
--- a/android/app/src/main/java/com/microsoft/codepush/react/CodePushUpdateManager.java
+++ b/android/app/src/main/java/com/microsoft/codepush/react/CodePushUpdateManager.java
@@ -163,6 +163,7 @@ public class CodePushUpdateManager {
         try {
             URL downloadUrl = new URL(downloadUrlString);
             connection = (HttpURLConnection) (downloadUrl.openConnection());
+            connection.setRequestProperty("Accept-Encoding", "identity");
             bin = new BufferedInputStream(connection.getInputStream());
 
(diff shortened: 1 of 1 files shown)


chaseadamsio:
for a little more context:

When creating a new connection, gzip compression is used by default:

> By default, this implementation of HttpURLConnection requests that servers use gzip compression and it automatically decompresses the data for callers of getInputStream(). ([android HttpURLConnection docs](https://developer.android.com/reference/java/net/HttpURLConnection.html))

If a request is sent with `Accept-Encoding` that wants gzip, when the `downloadPackage` method is called, it will fail every time on the `totalBytes != receivedBytes` because `totalBytes` has a `Content-Length` of -1 ([this happens in `CodePushUpdateManager`](https://github.com/Microsoft/react-native-code-push/blob/d3671783eded9813a9690e86858a3067270776af/android/app/src/main/java/com/microsoft/codepush/react/CodePushUpdateManager.java#L198)). 

This PR should increase robustness of the SDK.

datvong-wm:
Note the Performance section of https://developer.android.com/reference/java/net/HttpURLConnection.html

`By default, this implementation of HttpURLConnection requests that servers use gzip compression and it automatically decompresses the data for callers of getInputStream(). The Content-Encoding and Content-Length response headers are cleared in this case.`

PR #1261, opened 2018-04-20
Fix catching IOException

alexandergoncharov:
Remove `throw` because we shouldn't break app in case if we can't close stream.


android/app/src/main/java/com/microsoft/codepush/react/FileUtils.java
```diff
@@ -47,7 +47,7 @@ public static void copyDirectoryContents(String sourceDirectoryPath, String dest
                         if (fromBufferedStream != null) fromBufferedStream.close();
                         if (destStream != null) destStream.close();
                     } catch (IOException e) {
-                        throw new CodePushUnknownException("Error closing IO resources.", e);
+                        CodePushUtils.log("Error closing IO resources. " + e);
```
iTOYS: Please use `e.getMessage()`

alexandergoncharov:
Closed at now because we finding the best solution for this issue.

PR #1262, opened 2018-04-26
Update postlink.js

tmaly1980:
Fixing bug where CocoaPods files (ios/Pods/Pods.xcodeproj) causes error "Could not determine correct xcode proj path to retrieve related plist file, there are multiple xcodeproj under the solution"


iTOYS:
Hi @tmaly1980 
Thank you for contributing!

alexandergoncharov:
LGTM!

@tmaly1980 Thanks for contributing!

Issue #1244, opened 2018-04-04
Can you change deployment keys and thus code push package with the same app build?


Catpants:
Thanks so much for filing an issue or feature request! Please fill out the following (wherever relevant):

### Steps to Reproduce
1. Call codePush.sync(deploymentKey: x, codePush.InstallMode.IMMEDIATE) from a button within the app. App does nothing.
2. Call codePush.sync(deploymentKey: y, codePush.InstallMode.IMMEDIATE) from another button within the app. App does nothing.

### Expected Behavior
What you expected to happen?
I expect the app to restart and apply a new code push package based on the different deployment key.
### Actual Behavior
What actually happens?
```
Nothing happens on the ui.
  STACK TRACE AND/OR SCREENSHOTS
logs:
An update is available, but it is being ignored due to having been previously rolled back
```

### Environment

* react-native-code-push version:5.1.3-beta
* react-native version:0.53.0
* iOS/Android/Windows version: tested on ios 11
* Does this reproduce on a debug build or release build? both
* Does this reproduce on a simulator, or only on a physical device? only tested on device

(The more info the faster we will be able to address it!)

iTOYS:
Hi @Catpants 
Please see [example usage](https://github.com/Microsoft/react-native-code-push/blob/master/docs/api-js.md#syncoptions) of SyncOptions.
Shortly, in your case:
```javascript
codePush.sync({ 
   deploymentKey: "KEY", 
   installMode: codePush.InstallMode.IMMEDIATE
});
```

Thanks!

Catpants:
@iTOYS I apologise, my original code example was not accurate. I am indeed calling it like this:
      codePush.sync(
        { installMode: codePush.InstallMode.IMMEDIATE, deploymentKey: Config.CODEPUSH_KEY_IOS }
      )
I still have the issue where code push is rolling back and not syncing on demand with two different keys. I would like to know, is it even possible to do this?

newyankeecodeshop:
I have noticed that calling `sync` to change deployment from Production to Staging will not install the staging build if CodePush determines that the current (production) bundle is up-to-date. We had to write native code which calls `CodePush.clearUpdates()` before calling `sync` so that the requested bundle would be downloaded.

iTOYS:
@Catpants 
I've checked again with this config:
```
let codePushOptions = { checkFrequency: codePush.CheckFrequency.MANUAL };
export default App = codePush(codePushOptions)(App);
```
[...]

PR #1269, opened 2018-05-03
Expose clearUpdates method to RN for purging downloaded packages

marc-whil:
This is useful when changing to a deployment with an older version than
the currently installed package.

In our case, we need our QA team to be able to toggle between Staging and Production deployments, even if it means they're switching to an older release.  This patch gives us the option to clear downloaded updates when we change deployment keys, allowing us to consistently switch to the new deployment key.  This would not be needed if the desired behavior was to always upgrade to the newest versions and never downgrade to an older version, as might be desired in a "beta channel" situation with real users.

Thanks to @newyankeecodeshop for pointing out that this is possible here: https://github.com/Microsoft/react-native-code-push/issues/1244#issuecomment-378970595

### DIFF
--- a/CodePush.js
+++ b/CodePush.js
@@ -542,6 +542,7 @@ if (NativeCodePush) {
     sync,
     disallowRestart: RestartManager.disallow,
     allowRestart: RestartManager.allow,
+    clearUpdates: NativeCodePush.clearUpdates,
     InstallMode: {
       IMMEDIATE: NativeCodePush.codePushInstallModeImmediate, // Restart the app immediately
       ON_NEXT_RESTART: NativeCodePush.codePushInstallModeOnNextRestart, // Don't artificially restart the app. Allow the update to be "picked up" on the next app restart
--- a/android/app/src/main/java/com/microsoft/codepush/react/CodePushNativeModule.java
+++ b/android/app/src/main/java/com/microsoft/codepush/react/CodePushNativeModule.java
@@ -535,4 +535,9 @@ public class CodePushNativeModule extends ReactContextBaseJavaModule {
             }
         }
     }
+
+    @ReactMethod
+    public void clearUpdates() {
+        mCodePush.clearUpdates();
+    }
 }
--- a/docs/api-js.md
+++ b/docs/api-js.md
@@ -28,6 +28,8 @@ When you require `react-native-code-push`, the module object provides the follow
 
 * [sync](#codepushsync): Allows checking for an update, downloading it and installing it, all with a single call. Unless you need custom UI and/or behavior, we recommend most developers to use this method when integrating CodePush into their apps
 
+* [clearUpdates](#clearupdates): Clear all downloaded CodePush updates. This is useful when switching to a different deployment which may have an older release than the current package.
+
 #### codePush
 
 ```javascript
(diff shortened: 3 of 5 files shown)


alexandergoncharov:
Hi @marc-whil 

Thanks for contributing!

half-ogre:
Howdy @marc-whil, I just wanted to let you know this is still in the queue for us to look at. I recently joined the project and am ramping up on it, so things will move a bit slower. This is a pretty busy week (as as last week with the Build conference), so I probably won't be able to look at it until next week. Apologies!

NickToropov:
Hi @marc-whil, could you please describe your flow with usage `clearUpdates` method more detailed? Some code snippets or demos will also be very useful and help us to make decision about your PR.

marc-whil:
@NickToropov sure, here's an action that happens when our QA team wants to switch between staging/production code push bundles:

```
export const updateCodePushKey = key => dispatch => {
  // change code push actively
  dispatch(setCodePushKey(key))

  // clear code push cache so we're guaranteed to download the package from code push
  codePush.clearUpdates()

  // show update screen
  dispatch(codePushMandatoryUpdateBegin())

  // fetch and apply bundle from code push
  codePush
    .checkForUpdate(key) // there will always be an update here because of clearUpdates above
    .then(remotePackage => {
      if (!remotePackage) {
        // bail if there's no new package
        dispatch(codePushMandatoryUpdateEnd())
        console.log('bailed, theres no new package') // eslint-disable-line no-console
        return
      }

      remotePackage
        .download(progress => {
          // download the package with progress updates
          dispatch(codePushDownloadDidProgress(progress))
        })
        .then(localPackage => {
          // install the package
          console.log('installing local package') // eslint-disable-line no-console
          const installMode = codePush.InstallMode.IMMEDIATE
          localPackage.install(installMode)
          console.log('local package installed') // eslint-disable-line no-console
        })
        .catch(() => {
          console.log('there was an error') // eslint-disable-line no-console
          // download failed
          dispatch(codePushMandatoryUpdateEnd())
          // restart app because this update was mandatory
          codePush.restartApp()
        })
    })
    .catch(err => {
      console.log('CodePush error: ', err) // eslint-disable-line no-console
      bugsnag.leaveBreadcrumb('CodePush error', err)
    })
}
```

If we tried to switch keys without `clearUpdates`, our QA team would not be able to switch keys reliably, as Code Push will not take a downgrade when switching keys.  It won't apply something later over something newer.  By clearing the Code Push cache, Code Push doesn't know that we're really doing a downgrade (in the sense that the switch may bring us to older code), so it applies the bundle from the key we want.

christophermark:
Building off what @marc-whil said, when you switch between different environments (staging/production), we can run into issues with metadata when we attempt to switch between environment deployments.

Often we have Stable and Production running the same build after we promote a successful release. They might share a hash but be on different version numbers. If a user were to switch between the stable and production release channels in-app, CodePush refuses to download a bundle with the same hash, even if it is coming from a different deployment (different deployment key).

In our app, we'll have switched channels from Staging to Production, but when we call `CodePush.getUpdateMetadata()`, it will return the metadata from the Staging release, rather than the production release.

The easiest solution to this would be exposing this `clearUpdates` method so we can clear any updates from a non-related release channel as we switch to a different deployment.

NickToropov:
Hi all! We've decided to merge the PR with some additional notes in docs and logging. Thanks for contribution!

PR #1286, opened 2018-05-21
Prevent crashing app by catching errors

alexandergoncharov:
It is hotfix for preventing crash app by catching Exception without next throwing. In the future this functional will be implemented in _Brownfields_.


NickToropov:
LGTM

PR #1295, opened 2018-05-25
Fix typo

MrChristofferson:
necessarely -> necessarily


alexandergoncharov:
Hi @MrChristofferson,
Thanks for contributing!

Good catch!

PR #1325, opened 2018-06-25
remove AcquisitionManager import from package-mixins

krunalsshah:
[Minor] 
Problem : AcquisitionManager is unused in `package-mixins.js`
Solution : Removing AcquisitionManager import from `package-mixins.js`


NickToropov:
Hi @krunalsshah 
Thanks for contributing!

Issue #1272, opened 2018-05-03
Development/debug build crashes on android after downloading update


ujwal-setlur:
Thanks so much for filing an issue or feature request! Please fill out the following (wherever relevant):

### Steps to Reproduce
1. Build android development version of react-native app with code-push enabled.
2. Load on android emulator or phone
3. App loads and says there is an update.
4. After update is downloaded, app crashes.

### Expected Behavior
New bundle should run

### Actual Behavior
New bundle crashes, and code-push reverts to existing bundle.
```
05-03 23:06:08.639  4208  4251 I ReactNativeJS: Downloading package.
05-03 23:06:09.285  4208  4251 I ReactNativeJS: 65100 of 4928365 received.
05-03 23:06:09.325  4208  4251 I ReactNativeJS: 171596 of 4928365 received.
05-03 23:06:09.374  4208  4251 I ReactNativeJS: 343628 of 4928365 received.
05-03 23:06:09.416  4208  4251 I ReactNativeJS: 417356 of 4928365 received.
05-03 23:06:09.460  4208  4251 I ReactNativeJS: 548428 of 4928365 received.
05-03 23:06:09.505  4208  4251 I ReactNativeJS: 663116 of 4928365 received.
05-03 23:06:09.550  4208  4251 I ReactNativeJS: 785996 of 4928365 received.
05-03 23:06:09.596  4208  4251 I ReactNativeJS: 917068 of 4928365 received.
05-03 23:06:09.638  4208  4251 I ReactNativeJS: 1015372 of 4928365 received.
05-03 23:06:09.683  4208  4251 I ReactNativeJS: 1171020 of 4928365 received.
05-03 23:06:09.726  4208  4251 I ReactNativeJS: 1310284 of 4928365 received.
05-03 23:06:09.770  4208  4251 I ReactNativeJS: 1482316 of 4928365 received.
05-03 23:06:09.816  4208  4251 I ReactNativeJS: 1629772 of 4928365 received.
05-03 23:06:09.860  4208  4251 I ReactNativeJS: 1777228 of 4928365 received.
05-03 23:06:09.904  4208  4251 I ReactNativeJS: 1932876 of 4928365 received.
05-03 23:06:09.949  4208  4251 I ReactNativeJS: 2063948 of 4928365 received.
05-03 23:06:09.993  4208  4251 I ReactNativeJS: 2260556 of 4928365 received.
[...]

PR #1332, opened 2018-07-04
Prevent removing ReactNativeDevBundle in LiveReload

alexandergoncharov:
Fix for https://github.com/Microsoft/react-native-code-push/issues/1272
Prevent removing `ReactNativeDevBundle.js` file in case if LiveReload mode is enabled.
In other cases all behaviour is same.

### DIFF
--- a/android/app/src/main/java/com/microsoft/codepush/react/CodePushNativeModule.java
+++ b/android/app/src/main/java/com/microsoft/codepush/react/CodePushNativeModule.java
@@ -118,7 +118,12 @@ public class CodePushNativeModule extends ReactContextBaseJavaModule {
 
     private void loadBundle() {
         clearLifecycleEventListener();
-        mCodePush.clearDebugCacheIfNeeded();
+        try {
+            mCodePush.clearDebugCacheIfNeeded(resolveInstanceManager());
+        } catch(Exception e) {
+            mCodePush.clearDebugCacheIfNeeded(null);
+        }
+
         try {
             // #1) Get the ReactInstanceManager instance, which is what includes the
             //     logic to reload the current React context.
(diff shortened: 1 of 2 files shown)


android/app/src/main/java/com/microsoft/codepush/react/CodePushNativeModule.java
```diff
@@ -118,7 +118,12 @@ private void setJSBundle(ReactInstanceManager instanceManager, String latestJSBu
 
     private void loadBundle() {
         clearLifecycleEventListener();
-        mCodePush.clearDebugCacheIfNeeded();
+        try {
+            mCodePush.clearDebugCacheIfNeeded(resolveInstanceManager());
```
max-mironov: Can we reuse `final ReactInstanceManager instanceManager = resolveInstanceManager();` variable from line 125 (from below) in `clearDebugCacheIfNeeded` to avoid unnecessary using reflection twice and additional try-catch block?

alexandergoncharov: Yeah, we can, but we have some cases when `resolveInstanceManager()` has errors. In this case we should use `clearDebugCacheIfNeeded` with `null` argument. I think that adding this logic into next try-catch block is unnecessary as it make it more difficult for understanding. 
Could you please share your thoughts about it?

max-mironov: No worries, I believe current solution fits well. The only thing that I'd probably add is a comment with e.g. reference to the issue or some explanation of the intent because it could be not evident why do we need this logic at all.

PR #1367, opened 2018-08-03
throw exception if js bundle file name is not in download

### LINKED ISSUE #1366: [Android] a unzipped folder exist in device, if downloaded bundle has missing js bundle file name
Thanks so much for filing an issue or feature request! Please fill out the following (wherever relevant):

### Steps to Reproduce
1. Set Up instruction in Android 
https://github.com/Microsoft/react-native-code-push/blob/master/docs/setup-android.md
It expects to set 
```
@Override
        protected String getJSBundleFile() {
            return CodePush.getJSBundleFile();
        }
```
Reference 
https://github.com/facebook/react-native/blob/master/ReactAndroid/src/main/java/com/facebook/react/ReactNativeHost.java#L131
- In my case `null` resulted in exception highlighted below 
https://github.com/Microsoft/react-native-code-push/blob/master/android/app/src/main/java/com/microsoft/codepush/react/CodePushUpdateManager.java#L241
2. I noticed the directory with `packageHash` still exists in `data/data/<package_id>/files/CodePush/<packagehash>/CodePush/

### Expected Behavior
What you expected to happen?
Exception thrown and no need to create a copy first till preconditions are met

### Actual Behavior
What actually happens?
a unzipped folder exist in device, if downloaded bundle has missing js bundle file name 

### Solution
Throw exception prior to creating a copy of unzipped folder

krunalsshah:
Fixes #1366


android/app/src/main/java/com/microsoft/codepush/react/CodePushUpdateManager.java
```diff
@@ -223,67 +223,67 @@ public void downloadPackage(JSONObject updatePackage, String expectedBundleFileN
...
             }
 
-            FileUtils.copyDirectoryContents(unzippedFolderPath, newUpdateFolderPath);
-            FileUtils.deleteFileAtPathSilently(unzippedFolderPath);
-
             // For zip updates, we need to find the relative path to the jsBundle and save it in the
             // metadata so that we can find and run it easily the next time.
-            String relativeBundlePath = CodePushUpdateUtils.findJSBundleInUpdateContents(newUpdateFolderPath, expectedBundleFileName);
-
+            String relativeBundlePath = CodePushUpdateUtils.findJSBundleInUpdateContents(unzippedFolderPath, expectedBundleFileName);
```
krunalsshah: the only change is to fail first if the pre condition is not met. `relativeBundlePath` can be discovered from `unzippedFolderPath` before copying `unzippedFolderPath` to  `newUpdateFolderPath`

krunalsshah: other change is formatting the code without `else` block that existed earlier.

alexandergoncharov:
Hi @krunalsshah,
Thanks for contributing!

android/app/src/main/java/com/microsoft/codepush/react/CodePushUpdateManager.java
```diff
@@ -223,67 +223,70 @@ public void downloadPackage(JSONObject updatePackage, String expectedBundleFileN
...
                 }
+                throw new CodePushInvalidUpdateException("Update is invalid - A JS bundle file named \"" + expectedBundleFileName + "\" could not be found within the downloaded contents. Please check that you are releasing your CodePush updates using the exact same JS bundle file name that was shipped with your app's binary.");
+            }
 
-                if (isDiffUpdate) {
-                    CodePushUtils.log("Applying diff update.");
-                } else {
-                    CodePushUtils.log("Applying full update.");
-                }
+            FileUtils.copyDirectoryContents(unzippedFolderPath, newUpdateFolderPath);
```
alexandergoncharov: Could you please add removing of unzipped folder like in Master? 
`FileUtils.deleteFileAtPathSilently(unzippedFolderPath);`
From here: https://github.com/Microsoft/react-native-code-push/blob/master/android/app/src/main/java/com/microsoft/codepush/react/CodePushUpdateManager.java#L234

android/app/src/main/java/com/microsoft/codepush/react/CodePushUpdateManager.java
```diff
@@ -223,67 +223,70 @@ public void downloadPackage(JSONObject updatePackage, String expectedBundleFileN
...
+            String relativeBundlePath = CodePushUpdateUtils.findJSBundleInUpdateContents(unzippedFolderPath, expectedBundleFileName);
             if (relativeBundlePath == null) {
-                throw new CodePushInvalidUpdateException("Update is invalid - A JS bundle file named \"" + expectedBundleFileName + "\" could not be found within the downloaded contents. Please check that you are releasing your CodePush updates using the exact same JS bundle file name that was shipped with your app's binary.");
-            } else {
-                if (FileUtils.fileAtPathExists(newUpdateMetadataPath)) {
-                    File metadataFileFromOldUpdate = new File(newUpdateMetadataPath);
-                    metadataFileFromOldUpdate.delete();
+                FileUtils.deleteFileAtPathSilently(unzippedFolderPath);
+                if (FileUtils.fileAtPathExists(newUpdateFolderPath)) {
+                    FileUtils.deleteFileAtPathSilently(newUpdateFolderPath);
```
NickToropov: It seems that this deletion is unnecessary here as the same is done on lines 148-151 and the folder isn't created later till `FileUtils.copyDirectoryContents(unzippedFolderPath, newUpdateFolderPath)` (your 245 line)

NickToropov:
Hi @krunalsshah, please do similar changes for iOS too to keep logic in sync.

android/app/src/main/java/com/microsoft/codepush/react/CodePushUpdateManager.java
krunalsshah: @alexandergoncharov Please notice https://github.com/Microsoft/react-native-code-push/pull/1367/files/1378ed54ab5c73dafcd2a49d17974db26515b357#diff-4aa56c56a1afb536396acce24d2d2015R242 (CodePushInvalidUpdateException  will be thrown & by adding the line as in master would leave the copy of unzipped folder in the memory as the statement will not be executed, so the PR moves the delete operation before the exception is thrown)

android/app/src/main/java/com/microsoft/codepush/react/CodePushUpdateManager.java
krunalsshah: In case of `isDiffUpdate` true , it may have been created. So the conditional to check if `newUpdateFolderPath` exists followed by delete may be necessary to make sure. Please let me know.

https://github.com/Microsoft/react-native-code-push/pull/1367/files/1378ed54ab5c73dafcd2a49d17974db26515b357#diff-4aa56c56a1afb536396acce24d2d2015R227

krunalsshah:
@NickToropov iOS implementation looks fine. I may not be a expert with Obj C, though it seems Android specific problem.

android/app/src/main/java/com/microsoft/codepush/react/CodePushUpdateManager.java
alexandergoncharov: @krunalsshah , yeah, That's great! You made the good thing that you've added removing unzippedFolderPath and newUpdateFolderPath before Error. But in the successful case(without any Error) unzipped folder will not be deleted at all. But it should be deleted after coping context.

Please correct me if I missed something.

krunalsshah: @alexandergoncharov yes you are absolutely correct. I have pushed the commit.

alexandergoncharov: @krunalsshah Thanks!

NickToropov:
@krunalsshah please check ios native [code](https://github.com/Microsoft/react-native-code-push/blob/9ff52368e2dad005ceb8ea7d8dabfd9b08b465ce/ios/CodePush/CodePushPackage.m#L191). Here unzipped content is copied to `newUpdateFolderPath` and below there is a code for checking if `relativeBundlePath` exists. This logic corresponds to existing logic in android native code. So you should update ios logic too for consistency.

krunalsshah:
@NickToropov 

My friend @ghostatron was going to look at the iOS implementation. I am not an expert at iOS :( 

If he doesn't reply. please feel free to make the change.

NickToropov:
@krunalsshah, will wait forward your response! Ping me when you have any updates

NickToropov:
Hey @krunalsshah! Do you have any updates?

krunalsshah:
@NickToropov Sorry for the delayed reply. I was on a vacation. w.r.t iOS change, I'd seek help from the core-team. I am constrained on time (due to catch up after vacation) :) Please let me know your thoughts.

NickToropov:
Hi @krunalsshah! my thoughts are the same - do in iOS part the same changes as you have done in Android part. You said that your friend @ghostatron was going to look at the iOS implementation. Did he have a chance to do it? Please let me know if you and @ghostatron won't do required changes

PR #1373, opened 2018-08-08
Fix Gradle warning: 'compile' dependencies are obsolete

iain-merrick-fanduel:
When using the React code push plugin, Android Studio shows this warning and links to:

https://developer.android.com/studio/build/gradle-plugin-3-0-0-migration?#new_configurations

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -26,6 +26,6 @@ android {
 }
 
 dependencies {
-    compile "com.facebook.react:react-native:+"
-    compile 'com.nimbusds:nimbus-jose-jwt:5.1'
+    implementation "com.facebook.react:react-native:+"
+    implementation 'com.nimbusds:nimbus-jose-jwt:5.1'
 }


alexandergoncharov:
Hi @iain-merrick-fanduel,
Thanks for contributing!

We'll investigate it as soon as possible. It is in queue. Sorry for delay.

Thanks,
Alexander

alexandergoncharov:
Hi @iain-merrick-fanduel ,

Thanks for this suggestions but I have to prevent this PR for now as we trying to keep max sync with react-native `build.gradle` file. For now there is still `compile`: https://github.com/facebook/react-native/blob/master/ReactAndroid/build.gradle#L296

I think we should to consider this pr in the future. But for now I'm going to close this PR.
Please let me know if you have any questions.

Thanks,
Alexander

iain-merrick-fanduel:
That sounds fine, I just wanted to make sure this is updated when/if "compile" is dropped in some future version of the Android Studio plugin.

Could all those 'compile' instances be updated now? Or do we still need to support older versions of Gradle?

alexandergoncharov:
Yeah, we would like to support old Gradle versions too. And yeah, we will update it when it will be really necessary. We trying to support new and also old versions.

alexandergoncharov:
@iain-merrick-fanduel , I'm going to close this PR for now but return to it when it will be necessary to update 'compile'. 
Please feel free to reopen it if you have any questions.

Thanks,
Alexander

tapz:
"It will be removed at the end of 2018."

PR #1386, opened 2018-08-22
 Fixed create-app.js script

yuri-kulikov:
### Examples/CodePushDemoApp/App.js
* Fixed react-native Image issue  
  By replacing ```resizeMode={Image.resizeMode.contain}``` with ```resizeMode={"contain"}```
### Examples/create-app.js
* Added flag ```reactNativeVersionIsLowerThanV049``` to choose **.js** code sources for the created application 
* Fixed code that changes ```node_modules/react-native/script/react-native-xcode.sh```
* Fixed code that changes ```node_modules/react-native/react.gradle```
* Replaced ```execSync(`npm i --save ${reactNativeCodePushVersion}`);```
with ```execSync(`yarn add ${reactNativeCodePushVersion}`);```
to fix few react-native android version dependencies.  
Also added code to check if ```yarn``` is installed.

### DIFF
--- a/Examples/create-app.js
+++ b/Examples/create-app.js
@@ -35,6 +35,15 @@ if (fs.existsSync(appName)) {
 let appNameAndroid = `${appName}-android`;
 let appNameIOS = `${appName}-ios`;
 let reactNativeVersion = args[1] || `react-native@${execSync('npm view react-native version')}`.trim();
+
+let reactNativeVersionBellowV049;
+if (reactNativeVersion) {
+    let reactNativeVersionNumberString = reactNativeVersion.split("@")[1];
+    reactNativeVersionBellowV049 = reactNativeVersionNumberString.split('.')[1] < 49
+} else {
+    reactNativeVersionBellowV049 = false;
+}
+
 let reactNativeCodePushVersion = args[2] || `react-native-code-push@${execSync('npm view react-native-code-push version')}`.trim();
 
 console.log(`App name: ${appName}`);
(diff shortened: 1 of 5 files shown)


Examples/create-app.js
```diff
@@ -89,7 +98,9 @@ function generatePlainReactNativeApp(appName, reactNativeVersion) {
 
 function installCodePush(reactNativeCodePushVersion) {
     console.log(`Installing React Native Module for CodePush...`);
-    execSync(`npm i --save ${reactNativeCodePushVersion}`);
+    // execSync(`npm i --save ${reactNativeCodePushVersion}`);
+    // Fixed some problems with dependencies
+    execSync(`yarn add ${reactNativeCodePushVersion}`);
```
max-mironov: Don't get this, could you please clarify what problems w/ dependencies are when using npm? It seems like I usually use npm for new react-native+cp project and don't see any issues.

Examples/create-app.js
```diff
@@ -152,14 +171,17 @@ function optimizeToTestInDebugMode() {
...
     execSync(`perl -i -p0e 's/#ifdef DEBUG.*?#endif/jsCodeLocation = [CodePush bundleURL];/s' ios/${appName}/AppDelegate.m`);
-    execSync(`sed -ie '17,20d' node_modules/react-native/${rnXcodeShLocationFolder}/react-native-xcode.sh`);
-    execSync(`sed -ie 's/targetName.toLowerCase().contains("release")$/true/' node_modules/react-native/react.gradle`);
+    execSync(`sed -ie 's/targetName.toLowerCase().contains("release")/true/' node_modules/react-native/react.gradle`);
 }
 
 function grantAccess(folderPath) {
     execSync('chown -R `whoami` ' + folderPath);
-    execSync('chmod -R 755 ' + folderPath);
+    // execSync('chmod -R 755 ' + folderPath);
```
max-mironov: Could you please provide the exact error/problem caused by this line? It works on mac previously so just wondering what happens now? 
Also if it turns out that we really do not need this it make sense to remove this like and not to comment it.

Examples/create-app.js
```diff
@@ -35,6 +35,15 @@ if (fs.existsSync(appName)) {
 let appNameAndroid = `${appName}-android`;
 let appNameIOS = `${appName}-ios`;
 let reactNativeVersion = args[1] || `react-native@${execSync('npm view react-native version')}`.trim();
+
+let reactNativeVersionBellowV049;
+if (reactNativeVersion) {
+    let reactNativeVersionNumberString = reactNativeVersion.split("@")[1];
+    reactNativeVersionBellowV049 = reactNativeVersionNumberString.split('.')[1] < 49
+} else {
+    reactNativeVersionBellowV049 = false;
```
max-mironov: I'm no sure if we need this condition at all. 
1) We can assign `reactNativeVersionBellowV049` false by default
2) If we do not have `reactNativeVersion` specified may be it make sense to terminate the script and provide an error? Will the code  below work if we do not specify rn version?

Examples/create-app.js
```diff
@@ -35,6 +35,15 @@ if (fs.existsSync(appName)) {
 let appNameAndroid = `${appName}-android`;
 let appNameIOS = `${appName}-ios`;
 let reactNativeVersion = args[1] || `react-native@${execSync('npm view react-native version')}`.trim();
+
+let reactNativeVersionBellowV049;
+if (reactNativeVersion) {
+    let reactNativeVersionNumberString = reactNativeVersion.split("@")[1];
```
max-mironov: Lets move these two lines below that identifies if rn version is lower than certain rn version to separate function like `function isReactNativeVesionLowerThan(reactNativeCodePushVersion, versionToCheckFor)`

Examples/create-app.js
```diff
@@ -35,6 +35,15 @@ if (fs.existsSync(appName)) {
 let appNameAndroid = `${appName}-android`;
 let appNameIOS = `${appName}-ios`;
 let reactNativeVersion = args[1] || `react-native@${execSync('npm view react-native version')}`.trim();
+
+let reactNativeVersionBellowV049;
```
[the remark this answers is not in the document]
max-mironov: also probably `reactNativeVersionIsLowerThanV049` would be a better naming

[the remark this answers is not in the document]
yuri-kulikov: Yeah, it's a typo and I'll rename it as you suggest.

Examples/create-app.js
yuri-kulikov: When I use ```npm```, I'm getting this (when I try to build the created app on android):
```
FAILURE: Build failed with an exception.
* What went wrong:
Could not list contents of 'path_to_app/node_modules/fsevents/node_modules/.bin/node-pre-gyp'. Couldn't follow symbolic link.
```
But ```yarn``` solve this problem.
I'll create a method then will check if ```yarn``` is installed.

Examples/create-app.js
yuri-kulikov: I guess that it was my system specific issue and I can't reproduce it now, so I think it's better to just return this code.

Examples/create-app.js
yuri-kulikov: 1. Yes, I can do this, it will be ```false``` by default.
1. If ```reactNativeVersion``` isn't specified, then script will just use latest version:
```
reactNativeVersion = args[1] || `react-native@${execSync('npm view react-native version')}`.trim();
```

Examples/create-app.js
yuri-kulikov: Yes, it makes sense, I'll do this.

Examples/create-app.js
```diff
@@ -175,4 +194,14 @@ function copyRecursiveSync(src, dest) {
     } else {
         fs.linkSync(src, dest);
     }
+}
+
+function isReactNativeVesionLowerThan(version) {
+    if (!reactNativeVersion ||
```
max-mironov: nit: probably won't hurt to validate `version` param is not null and is number

PR #1388, opened 2018-08-24
Fixed 'Too many open files in system' problem in 'create-app.js'

yuri-kulikov:
Deleted ```chmod``` command as not necessary and caused ```Too many open files in system```


max-mironov:
Thanks for fixing!

Issue #626, opened 2016-11-30
Text files diff


lisong:
Do you have plan to use `google-diff-match-patch`  calculate Text file diff and merge ?

silhouettes:
Hi @lisong, not currently. We currently do a file-based diff (replace each file that has changed, but keep the files that have not changed). This makes sense to me, because most of the size of an app is likely to come from assets, like images and video. Is there any reason you think we should do a text-based diff?

lisong:
thanks for reply! sometimes you just fix a bug. maybe just a condition and react-native-cli has not split js bundles files, when app contains many business. jsbundles is really very big. it a expensive network. maybe google-diff-match-patch is a choices.

panarasi:
@lisong Thanks for the input. We have added this idea as something we would explore for the next version of Code Push. I am closing this issue here since it has been copied to our backlog.

thecachedbyte:
Any update on this? It will be awesome if we can achieve text-based diff patching. Downloading jsbundle of 1 or 2 mb on 2g or 3g network in country like India is a costly affair.

karthikmit:
I second @thecachedbyte, as it is a very important requirement for lite apps designed for low bandwidth constraints.

zjx:
agree with @thecachedbyte, using google-diff-match-patch will significantly reduce the potential expensive costs may charge to end users.

PR #1393, opened 2018-09-02
android add google diff-match-patch

lisong:
Add patchedFiles in `hotcodepush.json` which is use google diff-match-patch tools. this PR is works in android.

### DIFF
--- a/android/app/src/main/java/com/microsoft/codepush/react/CodePushUpdateUtils.java
+++ b/android/app/src/main/java/com/microsoft/codepush/react/CodePushUpdateUtils.java
@@ -132,8 +132,8 @@ public class CodePushUpdateUtils {
                 File fileToPatch = new File(currentPackageFolderPath, fileNameToPatch);
                 File patchFile = new File(unzippedFolderPath, fileNameToPatch);
                 if (fileToPatch.exists() && patchFile.exists()) {
-                    List<diff_match_patch.Patch> patches = dmp.patch_fromText(FileUtils.readFileToString(patchFile.getAbsolutePath()));
-                    Object[] results = dmp.patch_apply((LinkedList<diff_match_patch.Patch>) patches, FileUtils.readFileToString(fileToPatch.getAbsolutePath()));
+                    List<diff_match_patch.Patch> patches = dmp.patch_fromText(FileUtils.readFileToOriginString(patchFile.getAbsolutePath()));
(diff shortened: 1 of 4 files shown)


android/app/src/main/java/com/microsoft/codepush/react/CodePushUpdateUtils.java
```diff
@@ -286,4 +293,4 @@ public static void verifyUpdateSignature(String folderPath, String packageHash,
 
         CodePushUtils.log("The update contents succeeded the code signing check.");
     }
-}
```
lisong: what this ?

alexandergoncharov:
Hi @lisong ,
Thanks for contributing!

Could you please clarify more detailed about your PR? What exactly issue you solved with this PR and what approach have you used?
It can help make review for your PR faster.

Thanks,
Alexander

android/app/src/main/java/com/microsoft/codepush/react/CodePushUpdateUtils.java
alexandergoncharov: If you mean just the last brace then it is mean that you added new line at the end of file. Possible it is so because your automation code-style generator added this new line.

lisong:
@alexandergoncharov  thanks for reply! it's for https://github.com/Microsoft/react-native-code-push/issues/626 , it's useful features. i wonder if necessary, i will finish the mac and windows platform.

android/app/src/main/java/com/microsoft/codepush/react/CodePushUpdateUtils.java
lisong: thanks, it's right, i found the answer  when open with sublime text2.

lisong:
I see. it is need change server too.

Zakeelm:
Hey there!

@lisong Awesome work on this PR. Text file diff would definitely be a wonderful addition to CodePush. Unfortunately, we’re planning on completely re-writing our SDK and service logic in the near future so this is something we don’t think we can commit to at the moment on our end. On the bright side, this is something we actually plan on implementing along with these updates. While I don’t have an ETA on this, it is currently in our roadmap for the future of CodePush 😁

lisong:
OK

kufufu9:
How is it now? @Zakeelm

prorigorous:
How is it now? @Zakeelm

cani1see:
Any news on text diff now? what about microsoft's roadmap.  @Zakeelm

hnny09:
How is it now? @Zakeelm  
Do you have a plan?

PR #1399, opened 2018-09-09
iOS support text file diff use diff-match-patch

lisong:
#626


alexandergoncharov:
Depends https://github.com/Microsoft/react-native-code-push/pull/1393

Zakeelm:
Going to close this. Please see https://github.com/Microsoft/react-native-code-push/pull/1393#issuecomment-420452169 . This is something we're definitely looking at implementing with our planned changes!

lisong:
OK

PR #1400, opened 2018-09-10
Add warning about configuration variables in Info.plist to docs

tspop:
My project was was using configuration variables in Info.plist (as explained here https://www.appcoda.com/xcconfig-guide/).

I think the CLI tools don't support this. It took me a lot of time to figure out what was wrong and I thought others might appreciate the warning. 

Sorry if the usecase is too narrow.


NickToropov:
Hi @tspop! Could you please provide more details about the issue you've faced? It's possible that you had a different issue as it's an ordinary practice to use them with CodePush and everything should work properly. Please take a look at [the doc](https://github.com/Microsoft/react-native-code-push/blob/master/docs/multi-deployment-testing-ios.md) where it's descrubed how to configure environment for multideployment testing.
Let me know if it was helpful or I missed something?

PR #1403, opened 2018-09-14
add support for android gradle plugin 3.2 and above

ricoharisin:
starting from android gradle plugin 3.2, the location for asset folder is moved to `merged_asset`, causing codepush fail to find `CodePushHash`

### DIFF
--- a/android/codepush.gradle
+++ b/android/codepush.gradle
@@ -68,6 +68,14 @@ gradle.projectsEvaluated {
             jsBundleDir = elvisFile(config."$jsBundleDirConfigName") ?:
                     file("$buildDir/intermediates/assets/${targetPath}")
 
+             // In Case version of 'Android Plugin for Gradle'' is 3.2.0 and higher
+            // 'assets' folder is moved to 'merged_assets' with different structure inside
+            def variantName = variant.name
+
+            if (!jsBundleDir.exists() && file("$buildDir/intermediates/merged_assets/${variantName}/merge${targetName}Assets/out").exists()) {
+                jsBundleDir = file("$buildDir/intermediates/merged_assets/${variantName}/merge${targetName}Assets/out")
+            }
+            
             def resourcesDirConfigName = "resourcesDir${targetName}"
             resourcesDir = elvisFile(config."${resourcesDirConfigName}") ?:
                     file("$buildDir/intermediates/res/merged/${targetPath}")


NickToropov:
Hi @ricoharisin! Could you please clarify some questions? I've just created a basic RN (0.57.0) app with integrated CodePush(5.4.3) plugin, bumped android gradle plugin to `3.2+` and gradle itself to `4.6`,  built release apk and ran in Android 6.0.1. Result: red screen with error `Unable to load script from assets 'index.android.bundle'. Make sure your bundle is packaged correctly pr you're running a packager server`. So it seems that RN doesn't support gradle plugin above `3.1.4`. Do you have any other changes in your configuration that allow you to build and run RN app without that error?

alexandergoncharov:
Hi @ricoharisin ,
Thanks for contributing!

Could you please also provide repro steps for code version without your changes for your issue?
I would like to reproduce and debug it as I can't reproduce it.

NickToropov:
@ricoharisin I think that your issue is connected with skipping `bundleJsAndAssets` task. Definition of this task contains instructions for creating directories that are lately used by `generateBundledResourcesHash${targetName}` task. So to prevent the error you can skip `generateBundledResourcesHash${targetName}` task too. But in this case you should manually generate bundle hash on your `separate` step after manually build react native bundle.

ricoharisin:
@NickTropov i think not because if you use android gradle plugin to below 3.2.x it will generate `assets` directory inside `intermediates`. But if you use 3.2.x, `merged_assets` will be generate  instead

NickToropov:
@ricoharisin please try no to skip `bundleJsAndAssets` task and check if you get the same `Error: ENOENT: no such file or directory, open`

squirly:
To keep things complete, setting the following in `app/build.gradle` fixes everything across all `react-native` gradle functionality:
```
project.ext.react = [
    ...
    jsBundleDirDebug: "${buildDir}/intermediates/merged_assets/debug/mergeDebugAssets/out",
    jsBundleDirRelease: "${buildDir}/intermediates/merged_assets/release/mergeReleaseAssets/out"
]
```

adrianha:
Hi @NickToropov ,
I have same case with @ricoharisin that i generate bundle manually and skip `bundleJsAndAssets` task. i tried @ricoharisin 's solution and it worked.
Any thought on this @NickToropov ?

NickToropov:
Hi @adrianha! I think that if you generate bundle manually then you should generate hash manually too or provide additional support for hash generation script(something like @squirly mentioned). But it completely depends on your flow. Our build process logic is linked to RN one so if you change something we can't guarantee working flow go on correctly.

adrianha:
Actually, our build process still follow RN's and hash process is also working as usual. But have you tried to run codepush on android with gradle 3.2 @NickToropov ?

PR #1412, opened 2018-09-22
remove deprecated gradle compile dependency calls

squirly:
Dynamically link to react-native and statically link to jwt dependency

For details on the deprecation see: https://developer.android.com/studio/build/dependencies?utm_source=android-studio#dependency_configurations

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -26,6 +26,6 @@ android {
 }
 
 dependencies {
-    compile "com.facebook.react:react-native:+"
-    compile 'com.nimbusds:nimbus-jose-jwt:5.1'
+    compileOnly "com.facebook.react:react-native:+"
+    implementation 'com.nimbusds:nimbus-jose-jwt:5.1'
 }
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -26,6 +26,6 @@ android {
 }
 
 dependencies {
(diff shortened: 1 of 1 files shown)


alexandergoncharov:
Hi @squirly ,
Thanks for reporting!

android/app/build.gradle
```diff
@@ -26,6 +26,6 @@ android {
 }
 
 dependencies {
-    compile "com.facebook.react:react-native:+"
-    compile 'com.nimbusds:nimbus-jose-jwt:5.1'
-}
\ No newline at end of file
+    compileOnly "com.facebook.react:react-native:+"
```
alexandergoncharov: Could you please clarify why are you using `compileOnly` here for react-native?

squirly:
@alexandergoncharov I have updated this PR.

squirly:
@msftclas Please check this PR

alexandergoncharov:
Hi @squirly ,

Sorry for this delay.
I investigated it more closer and found that this changes affect users who still using Android Gradle Plugin less than 3.0.0 version. We are trying to support as much users as possible.
But you made good changes and a little later we will return to this PR and finish it. So, your efforts will not be lose.

So, for now I'm going to close this PR but please feel free to reopen it if you have any questions.

Thanks,
Alexander

PR #1419, opened 2018-09-27
[ts] Add missing null return type to `checkForUpdate`

squirly:
See documentation at https://github.com/Microsoft/react-native-code-push/blob/master/CodePush.js#L57


NickToropov:
Hi @squirly! Thanks for contributing.

Issue #1427, opened 2018-10-11
RN upgrade 0.56.1 -- app:generateBundledResourcesHashDebug can't find CodePushHash


peacechen:
### Steps to Reproduce

I upgraded an existing project to RN 0.56.1.  Previously the app built and ran on Android, but fails after the upgrade.
Are there any special steps needed for RN 0.56?  I scanned through the documentation and the setup is correct (same as before).
Perhaps of note is that I had to upgrade to Gradle 4.6 and associated tools (Android SDK 28, etc).

### Expected Behavior
Build & run on Android without errors

### Actual Behavior
Result of `react-native run-android`.  The file `CodePushHash` doesn't exist.

> 
> > Task :app:generateBundledResourcesHashDebug FAILED
> 4f53cda18c2baa0c0354bb5f9a3ecbe5ed12ab4d8e11ba873c2f11161202b945
> fs.js:663
>   return binding.open(pathModule.toNamespacedPath(path),
>                  ^
> 
> Error: ENOENT: no such file or directory, open '/Users/me/myproject/android/app/build/intermediates/assets/debug/CodePushHash'
>     at Object.fs.openSync (fs.js:663:18)
>     at Object.fs.writeFileSync (fs.js:1314:33)
>     at /Users/me/myproject/node_modules/react-native-code-push/scripts/generateBundledResourcesHash.js:81:16
>     at addFileToManifest (/Users/me/myproject/node_modules/react-native-code-push/scripts/generateBundledResourcesHash.js:98:9)
>     at /Users/me/myproject/node_modules/react-native-code-push/scripts/generateBundledResourcesHash.js:72:9
>     at addFileToManifest (/Users/me/myproject/node_modules/react-native-code-push/scripts/generateBundledResourcesHash.js:98:9)
>     at addJsBundleAndMetaToManifest (/Users/me/myproject/node_modules/react-native-code-push/scripts/generateBundledResourcesHash.js:70:5)
>     at Object.<anonymous> (/Users/me/myproject/node_modules/react-native-code-push/scripts/generateBundledResourcesHash.js:66:5)
>     at Module._compile (module.js:660:30)
>     at Object.Module._extensions..js (module.js:671:10)
> 

### Reproducible Demo

Can't repro the complex environment of our app in the example project.

### Environment

* react-native-code-push version: 5.4.2
* react-native version: 0.56.1
* iOS/Android/Windows version: Android 6
* Does this reproduce on a debug build or release build? Debug
[...]

PR #1447, opened 2018-11-07
Fixed the next issue

eladgel:
https://github.com/Microsoft/react-native-code-push/issues/1427


yuri-kulikov:
Hi @eladgel!

Thank you for your work, it's really great, but I'm afraid this will not help users to solve the problem completely.
The main reason behind this issue is that `react-native` < 0.57.2 can't work with `gradle plugin` >= 3.2 properly. So even if we create this folder, it will not help us make the **release version** of the application work.

You can read more about it [here](https://github.com/Microsoft/react-native-code-push/issues/1427#issuecomment-430913598).

Please let me know if you have any questions!

PR #1477, opened 2018-12-18
Replace deprecated 'compile' gradle configuration with 'implementation'

hyochan:
According to [ReactNative@0.57](https://github.com/react-native-community/react-native-releases/blob/master/CHANGELOG.md#057) default gradle version is now `4.4`.

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -26,6 +26,6 @@ android {
 }
 
 dependencies {
-    compile "com.facebook.react:react-native:+"
-    compile 'com.nimbusds:nimbus-jose-jwt:5.1'
+    implementation "com.facebook.react:react-native:+"
+    implementation 'com.nimbusds:nimbus-jose-jwt:5.1'
 }


yuri-kulikov:
Hello @hyochan,
Thanks for contributing!

I'm afraid we still can't merge it. We already had such [PR](https://github.com/Microsoft/react-native-code-push/pull/1412) but we decided to wait (while 'compile' works with the latest gradle) to keep backward compatibility as long as possible. Someone, for example, can still use `react-native 0.56.1` with the old `gradle` and the last `react-native-code-push` version.

hyochan:
@yuri-kulikov Hi. I think it isn't a good idea because as framework updates, the packages need to update. If you are afraid it isn't backward-compatible, then the users have the choice not to upgrade `react-native-code-push` version (This could be posted in the readme). In my case, I always have to change this in the local environment because it isn't `forward-compatible` (Even when I re-install `node_modules`. I hoped this to be done for a long time, and I've just created `PR` because it hasn't been updated for a while.

yuri-kulikov:
@hyochan you're right and we have a plans to make it in the nearest future (probably at the beginning of next year). We'll just reopen and merge it when we are ready.

hyochan:
Thank you.

hammadzz:
@yuri-kulikov this is a bit painful that we are holding this in for a few who aren't updating they should need to update for bug fixes. It is now 2019 can we have this in asap?

PR #1494, opened 2018-12-29
Update README.md

NULLPointerGuy:
Code-push fails in case we have multiple instances of ReactInstance Manager in Android, addressing this in the updated readme.


NULLPointerGuy:
Hi @alexandergoncharov 
Thanks for the reply!
Can it be documented somewhere else?  had a hard time finding the reason for failure.

PR #1536, opened 2019-03-09
log when load bundle fails.

deepueg:
Took some time for us to figure why the new bundle was not getting installed on onResume. Since our application class does not extend `ReactApplication` it was throwing a class cast exception. Solved it by passing in the `ReactHolder` instance. 
Adding a log message to save some time for other devs.


alexandergoncharov:
Hi @deepueg,
Thanks for contributing!

PR #1595, opened 2019-05-25
docs: update android staging config

zhigang1992:
Add initWith release

without this, there is no `installReleaseStaging` taks

### DIFF
--- a/docs/multi-deployment-testing-android.md
+++ b/docs/multi-deployment-testing-android.md
@@ -25,9 +25,7 @@ To set this up, perform the following steps:
 
             releaseStaging {
                 ...
-                // The `initWith` property allows you to copy configurations from other build types,
-                // then configure only the settings you want to change. This one copies the debug build
-                // type, and then changes the manifest placeholder and application ID.
+                // The `initWith` property allows you to copy configurations from release
                 initWith release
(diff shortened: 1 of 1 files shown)


alexandergoncharov:
Hi @zhigang1992,
Thanks for contributing!

Could you please clarify these changes more detailed? I successfully run app with `releaseStaging` variant without `initWith`.

Also, I see that you copied comment for `initWith` from https://developer.android.com/studio/build/build-variants#build-types. There is information that you change the manifest placeholder and application ID in the comment(like in link) but there are not these changes in your code.
Please feel free to ask any questions.

zhigang1992:
sry for the miss leading comment, I only meant to copy the first line of comment.

I think the purpose here is pretty straight forward. Since we only want to modify `CODEPUSH_KEY`, it makes sense to just copy all the configs from release.

I haven't dig too deep on what caused my issue `installReleaseStaging` though, I'm not sure why it was happening on my env.

alexandergoncharov:
Possible yeah, this makes sense but in this case, all works correctly for me without this changes.
Just for now, I'm going to close this PR. Please open a separate issue for your case with some dummy project and we will continue our conversation there.  I would like to reproduce your issue and debug it. In the future, we can reopen this PR =)

Please let me know if you have any questions.

PR #1675, opened 2019-09-04
Redux Saga URL corrected

josevarghese:
Fixed the Doc URL of Redux Saga


alexandergoncharov:
Hi @josevarghese ,
Thanks for contributing!

It is a good catch!

PR #1740, opened 2019-12-10
Move PublicKey and ServerUrl properties to string resources

alexandergoncharov:
* Since it isn't possible to use a custom constructor with `autolinker`, setting `PublicKey` and `ServerUrl` in string resources. These properties are available by putting new item in `strings.xml` with name `CodePushPublicKey` and `CodePushServerUrl` respectively.
__Note: Code Signing Feature is enabled automatically if property `CodePushPublicKey` exists in string resources.__
* Changed prefix for deployment key. Now it is `CodePushDeploymentKey`
* Updated docs.


docs/api-android.md
```diff
@@ -1,5 +1,23 @@
 ### Java API Reference (Android)
 
+### API for React Native 0.60 version and above
+
+Since `autolinking` uses `react-native.config.js` to link plugins, constructors are specified in that file. But you can override custom variables to manage the CodePush plugin by placing these values in string resources.
+
+* __Public Key__ - used for bundle verification in Code Signing Feature. Please refer to [Code Signing](setup-android.md#code-signing-setup) section for more details about Code Signing Feature.
```
pfleidi: We could call it "the Code Signing Feature":

```suggestion
* __Public Key__ - used for bundle verification in the Code Signing Feature. Please refer to the [Code Signing](setup-android.md#code-signing-setup) section for more details about Code Signing Feature.
```

Examples/CodePushDemoApp-pre0.49/android/app/src/main/java/com/codepushdemoapp/MainApplication.java
```diff
@@ -30,7 +30,7 @@ public boolean getUseDeveloperSupport() {
     protected List<ReactPackage> getPackages() {
       return Arrays.<ReactPackage>asList(
           new MainReactPackage(),
-            new CodePush(getResources().getString(R.string.reactNativeCodePush_androidDeploymentKey), getApplicationContext(), BuildConfig.DEBUG)
```
pfleidi: Just out of curiosity: Does anybody have historic context regarding why we used to prefix this string with `reactNativeCodePush_android`?

docs/setup-android.md
```diff
@@ -52,7 +52,7 @@ In order to integrate CodePush into your Android project, please perform the fol
...
+   To let the CodePush runtime know which deployment it should query for updates against, open your app's `strings.xml` file and add a new string named `CodePushDeploymentKey`, whose value is the key of the deployment you want to configure this app against (like the key for the `Staging` deployment for the `FooBar` app). You can retrieve this value by running `code-push deployment ls <appName> -k` in the CodePush CLI (the `-k` flag is necessary since keys aren't displayed by default) and copying the value of the `Deployment Key` column which corresponds to the deployment you want to use (see below). Note that using the deployment's name (like Staging) will not work. That "friendly name" is intended only for authenticated management usage from the CLI, and not for public consumption within your app.
```
pfleidi:
```suggestion
   To let the CodePush runtime know which deployment it should query for updates, open your app's `strings.xml` file and add a new string named `CodePushDeploymentKey`, whose value is the key of the deployment you want to configure this app against (like the key for the `Staging` deployment for the `FooBar` app). You can retrieve this value by running `code-push deployment ls <appName> -k` in the CodePush CLI (the `-k` flag is necessary since keys aren't displayed by default) and copying the value of the `Deployment Key` column which corresponds to the deployment you want to use (see below). Note that using the deployment's name (like Staging) will not work. The "friendly name" is intended only for authenticated management usage from the CLI, and not for public consumption within your app.
```

docs/api-android.md
alexandergoncharov: Good catch! Thanks. Updated.

docs/setup-android.md
alexandergoncharov: Thanks! Updated

Examples/CodePushDemoApp-pre0.49/android/app/src/main/java/com/codepushdemoapp/MainApplication.java
alexandergoncharov: That is good question. I did not find any real reason to keep this prefix. The IOS part has the `CodePush` prefix(`CodePushDeploymentKey`) which looks more consistent without any unnecessary context. And since we are adding new properties, it will be nice to have one prefix for all properties(like `CodePushDeploymentKey`, `CodePushPublicKey`, `CodePushServerUrl`).

alexandergoncharov:
@builtbyproxy, If I understand correctly your `androidKey` is `PublicKey`, then Yes.
Using `PublicKey` and the Code Signing feature is optional. You can just not use it.
And now the Code Signing feature in Android will work as in IOS - enabled if `strings.xml`(or `plist.info` for IOS) has `CodePushPublicKey` property.
So, It will be the same functionality for Android and IOS.
Hope it was helpful.

builtbyproxy:
Very helpful! Thanks a lot @alexandergoncharov

PR #1801, opened 2020-02-26
add overloaded type for CodePush function that accepts a React Component

gwmccull:
per the documentation, the `codePush` function can be used in one of the following ways:
```
codePush(config)(App)
codePush(App)
```
The current types only support the former call signature. I added a type definition that supports the 2nd call signature


gwmccull:
I chose to use `any` for the component type since that was what the current type definition uses and because the React types are not available in this repo

alexandergoncharov:
Hi @gwmccull ,
Thanks for contributing!

Good catch!

alexandergoncharov:
@gwmccull could you please merge the current Master into your branch?

gwmccull:
@alexandergoncharov I merged master

PR #1806, opened 2020-03-05
add appcenter new cli commands

ProMasoud:


alexandergoncharov:
Hi @ProMasoud ,
Thanks for contributing!

All Github documentation still using `code-push`. It will be not a good case if we change just one docs page. We will update all docs in the future but I can't provide any ETA for that.
So, I'm going to close this PR for now. Please feel free to reopen it if you have a question.

PR #1876, opened 2020-06-09
Fix everytime spelling

aaronkrohn:
everytime spelling should be two separate words

### DIFF
--- a/docs/api-js.md
+++ b/docs/api-js.md
@@ -63,7 +63,7 @@ This decorator provides support for letting you customize its behaviour to easil
 2. **Silent sync every time the app resumes**. Same as 1, except we check for updates, or apply an update if one exists every time the app returns to the foreground after being "backgrounded".
 
     ```javascript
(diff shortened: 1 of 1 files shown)


docs/api-js.md
```diff
@@ -60,7 +60,7 @@ This decorator provides support for letting you customize its behaviour to easil
     export default MyApp;
     ```
 
-2. **Silent sync everytime the app resumes**. Same as 1, except we check for updates, or apply an update if one exists every time the app returns to the foreground after being "backgrounded".
+2. **Silent sync every time the app resumes**. Same as 1, except we check for updates, or apply an update if one exists every time the app returns to the foreground after being "backgrounded".
 
     ```javascript
     // Sync for updates everytime the app resumes.
```
aaronkrohn, line 66:
```suggestion
    // Sync for updates every time the app resumes.
```

alexandergoncharov:
Hi @aaronkrohn ,
Thanks for contributing!

It is a good catch!

Issue #1879, opened 2020-06-18
Readme erroneously  states that the Video component isn't supported


district-michael:
The readme is out of date when it states that the <Video> component isn't supported.  It can use require('asset.mp4'), which allows the video to be bundled and distributed OTA

alexandergoncharov:
Hi @district-michael ,
Thanks for reporting!

Is it correct that your suggestion is adding an additional example to docs with videos?

district-michael:
Hi @alexandergoncharov,
Yes that is correct.  At the moment, the table states that the Video component cannot be used with `react-native-code-push`, as seen in the below table:

> The following list represents the set of components (and props) that don't currently support their assets being updated via CodePush, due to their dependency on static images (i.e. using the `{ uri: "foo" }` syntax):
> 
> | Component   | Prop(s)                                                              |
> |-------------|----------------------------------------------------------------------|
> | `SliderIOS` | `maximumTrackImage`, `minimumTrackImage`, `thumbImage`, `trackImage` |
> | `Video`     | `source`                                                             |
> 

I would suggest we need to add something to clarify that it works when using `source={require('./filename.mp4'}` but not `source={{uri: "filename.mp4"}}`

alexandergoncharov:
Oh, I see, cool! Thanks for your suggestion!
We will update the docs a bit later. We will keep you posted.

Krasavinigor:
Hi @district-michael!
I opened PR: https://github.com/microsoft/react-native-code-push/pull/1900 for updating documents. Could you please take a look if you have time?

alexandergoncharov:
Hi @district-michael ,

As https://github.com/microsoft/react-native-code-push/pull/1900 was merged I'm going to close this issue for now.
Please feel free to reopen it if you have any questions.

PR #1900, opened 2020-07-09
Update docs for using video component

Krasavinigor:
**Issue**: Outdated information of supported components https://github.com/microsoft/react-native-code-push/issues/1879

**Solution**: Update documents

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,7 +80,7 @@ We work hard to respond to new RN releases, but they do occasionally break us. W
 
 ### Supported Components
 
-When using the React Native assets system (i.e. using the `require("./foo.png")` syntax), the following list represents the set of core components (and props) that support having their referenced images updated via CodePush:
+When using the React Native assets system (i.e. using the `require("./foo.png")` syntax), the following list represents the set of core components (and props) that support having their referenced images and videos updated via CodePush:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -99,7 +100,11 @@ The following list represents the set of components (and props) that don't curre
 
 As new core components are released, which support referencing assets, we'll update this list to ensure users know what exactly they can expect to update using CodePush.
 
+*Note: CodePush works correctly with Video component only if using `require`. Example:*
```
district-michael, line 103: I'd reword this slightly to:
*Note: CodePush only works with Video components when using `require` in the source prop. For example:*

README.md
```diff
@@ -99,7 +100,11 @@ The following list represents the set of components (and props) that don't curre
...
 
+```javascript
+<Video source={require("foo")} />
```
district-michael, line 106: I'd also change the example here to actually include the file extension, ie `require("./foo.mp4")`

README.md
Krasavinigor: Thank you!

README.md
Krasavinigor: Yes, it will be better!

PR #1913, opened 2020-07-26
Fix Android build failure if `bundleJsAndAssets` is disabled

ivanmoskalev:
Previously, we have only checked for the existence of React Native's gradle task `bundleJsAndAssets` to determine whether we should use `generatedAssetsFolders` and `generatedResFolders`, or we should use folders provided by the user through React Native's gradle config (`project.react`). 

This broke Android builds when the user has disabled bundling in `project.react`, because as of React Native 0.63, the [`bundleJsAndAssets` gradle task is always created](https://github.com/facebook/react-native/blob/4f897336b6ce6c640a65f86b65c6d386ea55b944/react.gradle#L220). Consequently, `variant.hasProperty("bundleJsAndAssets")` is always `true`, and the else branch is never reached.

This commit adds an additional check whether `bundleJsAndAssets` task is enabled, provided it exists of course.

### DIFF
--- a/android/codepush.gradle
+++ b/android/codepush.gradle
@@ -48,7 +48,7 @@ gradle.projectsEvaluated {
         // Make this task run right after the bundle task
         def generateBundledResourcesHash;
 
-        if (variant.hasProperty("bundleJsAndAssets")) {
+        if (variant.hasProperty("bundleJsAndAssets") && variant.bundleJsAndAssets.enabled) {
             def reactBundleTask = variant.bundleJsAndAssets
             jsBundleDir = reactBundleTask.generatedAssetsFolders[0].absolutePath
             resourcesDir = reactBundleTask.generatedResFolders[0].absolutePath


ivanmoskalev:
Regrettably, Android tests failed: https://github.com/microsoft/react-native-code-push/runs/911938853#step:13:2406
I'll investigate tomorrow. I'm testing the patch on my own project and everything seems to work, though, so I guess that I'll have to run the failing test on my machine.

ivanmoskalev:
I actually found that the `else` branch causes a circular dependency:
```
* Exception is:
org.gradle.api.CircularReferenceException: Circular dependency between the following tasks:
:app:bundleQaReleaseJsAndAssets
\--- :app:recordFilesBeforeBundleCommandQaRelease
     \--- :app:mergeQaReleaseResources
          \--- :app:bundleQaReleaseJsAndAssets (*)

```
I'm probably gonna close this PR and file an issue after more extensive research.

PR #1955, opened 2020-10-08
Added installing cocoapods to script for creating apps

alexChernyatiev:
Added check that `CocoaPods` installed. If this is not the case, the script ends with error.

### DIFF
--- a/Examples/create-app.js
+++ b/Examples/create-app.js
@@ -61,14 +61,15 @@ let iosStagingDeploymentKey = null;
 createCodePushApp(appNameAndroid, 'Android');
 createCodePushApp(appNameIOS, 'iOS');
 
-if (process.platform === 'darwin') {
-    try {
-        execCommand(`pod --version`);
-        console.log("Cocoapods has already installed")
-    } catch {
-        console.log("Need to install Cocoapods")
-        execCommand(`sudo gem install cocoapods`);
-    }
+if (!isReactNativeVersionLowerThan(60) && process.platform === "darwin") {
+  try {
+    console.log("Verify that CocoaPods installed");
+    execCommand("pod --version");
+  } catch {
+    console.log("Need to install CocoaPods");
+    execCommand("sudo gem install cocoapods");
+  }
+  console.log("Cocoapods has already installed");
 }
 
 generatePlainReactNativeApp(appName, reactNativeVersion);
--- a/Examples/create-app.js
+++ b/Examples/create-app.js
@@ -61,6 +61,16 @@ let iosStagingDeploymentKey = null;
 createCodePushApp(appNameAndroid, 'Android');
 createCodePushApp(appNameIOS, 'iOS');
 
+if (process.platform === 'darwin') {
+    try {
+        execCommand(`pod --version`);
+        console.log("Cocoapods has already installed")
+    } catch {
+        console.log("Need to install Cocoapods")
+        execCommand(`sudo gem install cocoapods`);
+    }
+}
+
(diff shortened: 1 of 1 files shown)


Examples/create-app.js
```diff
@@ -61,6 +61,17 @@ let iosStagingDeploymentKey = null;
 createCodePushApp(appNameAndroid, 'Android');
 createCodePushApp(appNameIOS, 'iOS');
 
+if (!isReactNativeVersionLowerThan(60) && process.platform === "darwin") {
+  try {
+    console.log("Verify that CocoaPods installed");
+    execCommand("pod --version");
+  } catch {
+    console.log("Need to install CocoaPods");
+    execCommand("sudo gem install cocoapods");
```
snowpardx, line 70: Will it work if node process started from non root?

If not probably we can just ask the user to install and stop further execution for script.

Examples/create-app.js
```diff
@@ -61,6 +61,17 @@ let iosStagingDeploymentKey = null;
 createCodePushApp(appNameAndroid, 'Android');
 createCodePushApp(appNameIOS, 'iOS');
 
+if (!isReactNativeVersionLowerThan(60) && process.platform === "darwin") {
```
snowpardx, line 64: Can we please move this code close to start, so that install necessary dependencies before creating apps both in App Center and disk?

snowpardx (changes requested):

Examples/create-app.js
alexChernyatiev: It works if node process started from non root but process will ask enter password.

Examples/create-app.js
alexChernyatiev: No we can't. I tried to move it in linkCodePush() method after checking that platform is MacOS but script failed in generatePlainReactNativeApp() method.

Examples/create-app.js
Krasavinigor: I'm not sure that we should call installing command. In my opinion, @snowpardx's approach is more correct...

Krasavinigor:
Could you please add some details to the description?

Examples/create-app.js
snowpardx: Sorry, am I missing something? From my understanding it should be possible to:
1. move line 47 to line 43
2. place this newly added check after line 43.

This way we should be able to avoid apps creation when not all the dependencies are set up

Examples/create-app.js
alexChernyatiev: Ok. Done

Examples/create-app.js
alexChernyatiev: I understand. Thank you for review.

alexChernyatiev:
> Could you please add some details to the description?

Examples/create-app.js
```diff
@@ -41,6 +41,17 @@ try {
     process.exit();
 }
 
+if (!isReactNativeVersionLowerThan(60) && process.platform === "darwin") {
+    try {
+        console.log("Verify that CocoaPods installed");
+        execCommand("pod --version");
+        console.log("CocoaPods has installed");
+    } catch {
+        console.error(`You must install 'CocoaPods' to use this script!`);
```
snowpardx, line 50:
```suggestion
        console.error(`'CocoaPods' are required to run the script, you can install it with\n'sudo gem install cocoapods'\ncommand
        `);
```

snowpardx (approved):

Krasavinigor (approved):

alexandergoncharov (approved):

andreidubov (approved):

PR #1957, opened 2020-10-14
Fixed problem with android versions <= 4.4

alexChernyatiev:
**Expected Behavior**
Code push should update the app after the user closes the App and restart manually. Then only the update will show up in the App.

**Actual Behavior**
What actually happens?
Nothing happened, no update. And when i use it on splash screen and using loading it will get stuck on loading


Krasavinigor (approved):

ruslan-bikkinin (changes requested):

Krasavinigor (approved):

andreidubov (approved):

n-kurdiukov-akvelon (approved):

ruslan-bikkinin (approved):

android/app/src/main/java/com/microsoft/codepush/react/TLSSocketFactory.java
```diff
@@ -0,0 +1,72 @@
...
+        return enableTLSOnSocket(delegate.createSocket(address, port, localAddress, localPort));
+    }
+
+    private Socket enableTLSOnSocket(Socket socket) {
+        if (socket != null && (socket instanceof SSLSocket)) {
+            ((SSLSocket) socket).setEnabledProtocols(new String[] { "TLSv1.1", "TLSv1.2" });
+        }
+        return socket;
+    }
+}
```
alexandergoncharov, line 72: Small nit: Could you please add an empty line at the end of the file :)

alexandergoncharov (approved):

android/app/src/main/java/com/microsoft/codepush/react/TLSSocketFactory.java
alexChernyatiev: Thank you for review! Done.

alexandergoncharov (approved):

PR #1971, opened 2020-11-02
Added log in catch after creating app on appcenter

alexChernyatiev:
Added log with message to check that app with the same name not contains in appcenter


Examples/create-app.js
```diff
@@ -90,6 +90,9 @@ function createCodePushApp(name, os) {
         execCommand(`appcenter codepush deployment add -a ${owner}/${name} Staging`);
     } catch (e) {
         console.log(`App "${name}" already exists \n`);
+        const ownerResult = execCommand(`appcenter profile list --output json`);
```
Krasavinigor, line 93: nit: If we don't insert any expression into the string, could you replace it to double quotes?

Krasavinigor (approved):

andreidubov (approved):

PR #2007, opened 2020-12-30
No longer throw when attempting to hash

alexmoroz15:
@vmoroz told me that some users are having issues with the Windows implementation because it currently throws a NOT_IMPL error when it reaches a point when it should hash package contents. To keep their builds from breaking, this PR removes the throwing behavior. A hashing fail message is still logged in debug output.


vmoroz (approved):
:shipit:

PR #2064, opened 2021-04-09
add timeout parameter on checkForUpdate

### LINKED ISSUE #2032: Check when App Center services are down
Thanks so much for filing an issue or feature request! Please fill out the following (wherever relevant):

### Steps to Reproduce
1. When a user starts the app and if the App Center is down - it seems that my app is stuck and can't pass the updating screen.

### Expected Behavior
What did you expect to happen?
I'm expecting that somehow `CodePush` object returns the error so I could see if the App Center services are down and if so - I would expect for my users not to be stuck on the loading screen but instead to be forwarded inside the app with the version of the app which they already have.

### Actual Behavior
What actually happens?
Users are stuck at the beginning of the app.

### Environment

* react-native-code-push version: 6.4.1
* react-native version: 0.61.4
* iOS/Android all versions
* Release build
* This is reproducible on the physical device.

b3coded:
### Summary
Added `checkUpdateTimeout` property (like Axios does). It makes the developer prevent app for waiting too much time for `checkForUpdate` method.
I haven't the screenshot now but used `Flipper` to see my Network requests and the route to check update ended in approximately 64 seconds.

Added `checkUpdateTimeout` in `react-native-code-push.d.ts`

Fixes #2032 

### Breaking change

- `checkForUpdate` second parameter is now `checkUpdateTimeout`, `handleBinaryVersionMismatchCallback` is now the third parameter

### DIFF
--- a/CodePush.js
+++ b/CodePush.js
@@ -8,85 +8,103 @@ import hoistStatics from 'hoist-non-react-statics';
 let NativeCodePush = require("react-native").NativeModules.CodePush;
 const PackageMixins = require("./package-mixins")(NativeCodePush);
 
-async function checkForUpdate(deploymentKey = null, handleBinaryVersionMismatchCallback = null) {
-  /*
-   * Before we ask the server if an update exists, we
-   * need to retrieve three pieces of information from the
-   * native side: deployment key, app version (e.g. 1.0.1)
-   * and the hash of the currently running update (if there is one).
-   * This allows the client to only receive updates which are targetted
-   * for their specific deployment and version and which are actually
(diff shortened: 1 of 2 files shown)


smolleyes:
hi

what about this fix ? i can t see this params availble in 7.0.4 version ? 

thanks

NilsBaumgartner1994:
Push this is a nice feature

harlin22:
Will this feature be coming? I have noticed much longer wait times on calling `checkForUpdate` which really slows down the app on launch.

velimir-jankovic:
We are not planning to introduce any breaking changes.

Issue #2057, opened 2021-03-31
High severity vulnerability in netmask dependency


joebernard:
`react-native-code-push@7.0.0` relies on `netmask` via `pac-resolver`. The version being used is vulnerable to exploit as documented here:

https://github.com/advisories/GHSA-pch5-whg9-qr2r

Please update the package to avoid the vulnerability.

PR #2066, opened 2021-04-12
Update version of code-push to fix superagent-proxy vulnerability

dentuzhik:
Upstream dependency of this package already updated `superagent-proxy` and we just need a patch version bump.
https://github.com/microsoft/code-push/pull/759

Fixes https://github.com/microsoft/react-native-code-push/issues/2057

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -83,30 +83,35 @@
       }
     },
     "@nodelib/fs.scandir": {
-      "version": "2.1.3",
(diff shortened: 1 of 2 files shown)


alexandergoncharov:
Hi @dentuzhik ,
Thanks for contributing!

dentuzhik:
Thanks @alexandergoncharov 
Do you have some idea when we can expect a published package?

alexandergoncharov:
We are updating other dependencies now. It will be published during this week.

PR #2076, opened 2021-04-17
Add deployment key creation via CLI to the docs

christian-hess-94:
When me and my team were following the guide, we didn't have a deployment key setup to our account.

After some search we found a solution for creating a deployment key within the cli, the command:
`appcenter codepush deployment add -a {UserName}/{ProjectNameInAppCenter} DEPLOYMENT_ENV` 

Adding this line to show the usage of this command can help someone that doesn't have any previuous experience with CodePush or similar services

We suggest using the Staging | Production | Development to specify the environtment for which the key will be added, but we are not sure if this has to be enforced or the user is free to name the environment as they will.


velimir-jankovic:
This is already documented in https://github.com/microsoft/appcenter-cli#commands. Closing.

PR #2214, opened 2022-01-31
Bump nimbus-jose-jwt to 9.15.2 in android/app/build.gradle

owjsub:
In testing `nimbus-jose-jwt` on 9.15.2 with react-native-codepush, there were no observed compile time errors or breaking API changes, so updating to 9.15.2 should have low regression risk.


velimir-jankovic:
We will update this and other dependencies. No ETA for now.

PR #2269, opened 2022-05-17
Update multi-deployment-testing-android.md

luanfv:
As of version 0.68.0 of React Native, Hermes is enabled by default and it is necessary to do one more step in creating the releaseStaging for it to work, otherwise the generated APK crashes


velimir-jankovic:
Not relevant anymore.

https://github.com/microsoft/appcenter-cli/pull/1988

PR #2311, opened 2022-07-26
Fix pipeline tests

denysop:


.github/workflows/react-native-code-push-ci.yml
```diff
@@ -27,12 +27,23 @@ jobs:
...
+          java-version: '11'
       - name: Package Installation
-        run: npm install
+        run: npm i
+      - name: Install react-native
+        run: npm install -g react-native
+      - name: Setup Android tests
+        run: npm run build:tests && npm run test:setup:android
+      - name: Remove react-native
+        run: npm remove -g react-native
```
yevheniy-solovey-ms, line 42: Why do we need to remove the react-native?

.github/workflows/react-native-code-push-ci.yml
```diff
@@ -44,6 +55,8 @@ jobs:
         run: npm install
       - name: Install react-native-cli
         run: npm install react-native-cli
+      - name: Install reacti-native
+        run: npm install react-native
```
yevheniy-solovey-ms, line 59: And then install it here again

denysop: It is iOS pipeline. There is a separate task to make if functional so you can ignore this part for now.

.github/workflows/react-native-code-push-ci.yml
denysop: It is a workaround for the next error: ```ReferenceError: SHA-1 for file /metro-runtime/src/polyfills/require.js) is not computed.```

Taken from here: https://github.com/facebook/metro/issues/330#issuecomment-572917876

yevheniy-solovey-ms (approved):

PR #2215, opened 2022-01-31
Remove jcenter

owjsub:
"Based on the current timeline, builds that use JCenter will be able to resolve dependencies until February 1, 2022 without changes. After that date, there are no guarantees that you will be able to build your software if you continue to use JCenter." --https://blog.gradle.org/jcenter-shutdown

https://github.com/microsoft/react-native-code-push/issues/2213

PR #2364, opened 2022-10-31
replace jcenter with mavenCentral

abhiin1947:
replace jCenter with mavenCentral. jCentral is having a pretty long outage and it's deprecated. Move on from jCenter and replace it with mavenCentral


velimir-jankovic:
Duplicate https://github.com/microsoft/react-native-code-push/pull/2215

PR #2369, opened 2022-11-09
Replace deprecated "removeEventListener" syntax

adrienharnay:
This syntax has been deprecated for over a year now. This PR addresses this.

### DIFF
--- a/CodePush.js
+++ b/CodePush.js
@@ -180,7 +180,7 @@ async function notifyApplicationReadyInternal() {
   return statusReport;
 }
 
-async function tryReportStatus(statusReport, resumeListener) {
+async function tryReportStatus(statusReport, resumeListener, appStateListener) {
   const config = await getConfiguration();
   const previousLabelOrAppVersion = statusReport.previousLabelOrAppVersion;
   const previousDeploymentKey = statusReport.previousDeploymentKey || config.deploymentKey;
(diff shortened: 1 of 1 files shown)


CodePush.js
```diff
@@ -209,22 +209,23 @@ async function tryReportStatus(statusReport, resumeListener) {
...
         if (refreshedStatusReport) {
-          tryReportStatus(refreshedStatusReport, resumeListener);
+          tryReportStatus(refreshedStatusReport, resumeListener, localAppStateListener);
         } else {
-          AppState.removeEventListener("change", resumeListener);
+          localAppStateListener && localAppStateListener.remove();
         }
       };
-      AppState.addEventListener("change", resumeListener);
+      localAppStateListener = AppState.addEventListener("change", resumeListener);
```
thewulf7, line 228: Not sure I understand this line. Basically token (aka localAppStateListener) from adding listener, as later on passed to scope of resumeListener right to remove it?

If so Im not sure I understand what line 216 is doing.

adrienharnay: Hello, thank you for your answer. What I wanted to do:
- initialize `localAppStateListener` with the `appStateListener ` parameter l.216
- assign it the return of `addEventListener` l.228
- in the event listener, access `localAppStateListener` which will hold the return from `addEventListener`

I agree that initializing `localAppStateListener` might be pointless. Do you want me to make this change or did you have something else in mind?

jbaudanza:
This function has been removed from React Native 0.70, so this warning is now an error:

```
c.AppState.removeEventListener is not a function. (In 'c.AppState.removeEventListener("change",t)', 'c.AppState.removeEventListener' is undefined)
```

trcoffman (changes requested):
Upon reading this PR, I authored #2385 which I think is a cleaner fix. There's no reason to introduce a 3rd argument, as the only purpose of the argument is to (a) determine whether or not we have already established a resumeListener, and (b) to remove the listener at the appropriate time. When the function recurses, it only needs the subscription, not the closure itself, so rather than introduce another argument, we just change what the argument is. This PR can be closed and replaced by #2385.

Titozzz:
@thewulf7 Gentle nudge, feel free to merge #2385 and close this one if the syntax seems clearer to you

thewulf7:
Merged #2385.Closing this

PR #1943, opened 2020-09-08
fix priority in calculating node_modules location in gradle.

maxkorp:
See #1887 for details, but basically if config explicitly states where to find the node_modules folder, that should take priority over calculations based on a config setting saying where the overall root is.

Issue #1887, opened 2020-06-24
gradle calculation of nodeModulesPath is poorly prioritized


maxkorp:
If you specify a project root AND nodeModulesPath in your app/build.gradle, the nodeModulesPath is ignored, and the assumed root/node_modules is used. Instead, the more specific nodeModulesPath config should be used. I have not checked if an analogous problem exists in iOS

### Steps to Reproduce
1. Set up 2 react native projects in a monorepo at `monorepoRoot/packages/appA` and `monorepoRoot/packages/appB`
2. Fix hoisting so that react-native-code-push is installed at `monorepoRoot/node_modules/react-native-code-push`
3. Set values for `root` and `nodeModulesPath` config.ext.react, e.g. 
```
project.ext.react = [
    entryFile: "index.js",
    enableHermes: false,  // clean and rebuild if changing
    root: "../..",
    cliPath: "../../node_modules/react-native/cli.js"
]
project.ext.nodeModulesPath = "../../../../node_modules"
```

### Expected Behavior
I would expect for the nodeModulesPath used by codepush to be `../../../../node_modules`, which means the cli for bundling appA for example calls out to `monorepoRoot/node_modules/react-native-code-push/scripts/generateBundledResourcesHash.js`

### Actual Behavior
the nodeModulesPath used by codepush ends up as `../../node_modules` which  means the lie for bundling appA calls out to 
`monorepoRoot/packages/appA/node_modules/react-native-code-push/scripts/generateBundledResourcesHash.js`

### Proposed solution
I have not yet tested this deeply, but I unless pathing is an issue somewhere I am not seeing, we can reorder the calculation if statements to allow nodeModulesPath to take priority. 
Existing code (from codepush.gradle): 
```
        def nodeModulesPath;
        if (config.root) {
            nodeModulesPath = Paths.get(config.root, "/node_modules");
        } else if (project.hasProperty('nodeModulesPath')) {
            nodeModulesPath = project.nodeModulesPath
        } else {
            nodeModulesPath = "../../node_modules";
        }

```
Proposed change:
```
        def nodeModulesPath;
        if (project.hasProperty('nodeModulesPath')) {
            nodeModulesPath = project.nodeModulesPath
        } else if (config.root) {
            nodeModulesPath = Paths.get(config.root, "/node_modules");
        } else  else {
            nodeModulesPath = "../../node_modules";
        }
```

### Reproducible Demo
I'll work on getting one set up.

### Environment

[...]

PR #2507, opened 2023-04-25
Prioritize nodeModulesPath over config.root if present.

joarkosberg:
This has been discussed before in issue https://github.com/microsoft/react-native-code-push/issues/1887 and in pull request https://github.com/microsoft/react-native-code-push/pull/1943, but was closed because of inactivity.

Like mentioned in links, it would maybe be more natural to prioritize the direct `nodeModulesPath` property if present, not the reference to projects root.
And with react native 0.71 new android config, I have seen that root is present even when not setting it in project. Which means `nodeModulesPath` is never used, which makes it useless in that case. Therefore taking up this change again, to switch the priority.

### DIFF
--- a/android/codepush.gradle
+++ b/android/codepush.gradle
@@ -33,10 +33,10 @@ gradle.projectsEvaluated {
         }
 
         def nodeModulesPath;
-        if (config.root) {
-            nodeModulesPath = Paths.get(config.root.asFile.get().absolutePath, "/node_modules");
-        } else if (project.hasProperty('nodeModulesPath')) {
+        if (project.hasProperty('nodeModulesPath')) {
             nodeModulesPath = project.nodeModulesPath
+        } else if (config.root) {
+            nodeModulesPath = Paths.get(config.root.asFile.get().absolutePath, "/node_modules");
         } else {
             nodeModulesPath = "../../node_modules";
         }


DaniyarJakupov:
Faced the same issue with the wrong path to node_modules in monorepo setup after upgrading to react-native 0.72.3. Now `root` is always passed and it overrides custom `nodeModulesPath` that we have. For now used `patch-package`, but would be good to merge this PR
cc @DmitriyKirakosyan

mendesr-ppb:
**TL;DR**

We have just upgraded to RN 0.72.3, and our project follows a monorepo setup too. While the PR isn't tested and merged, we will apply the patch as a workaround for now.

--

We were about to reply in a related thread from the Issue Tracker, but we will share our struggles here instead.

We are running `RN 0.72.3` and `RN CodePush 8.0.2` against a monorepo setup also with different workspaces, e.g. web and native.

We noticed that `codepush.gradle` can [read](https://github.com/microsoft/react-native-code-push/blob/master/android/codepush.gradle#L39) a custom `node_modules` path via the `nodeModulesPath` project property.  Inside `android/app/build.gradle`, we leveraged `project.ext.nodeModulesPath` for our convenience to tell CodePush where to find its scripts i.e. the monorepo root `node_modules`.

The approach worked fine while running `RN 0.69.7` and `CodePush 7.1.0`. After the RN upgrade, we stopped seeing the `project.ext.nodeModulesPath` set because the `react` config in `android/app/build.gradle` would always have the `root` property defined with the root of the RN project, by default. Below is the error output:

```
Error: Cannot find module '/Users/$USER/monorepo/apps/native/node_modules/react-native-code-push/scripts/generateBundledResourcesHash.js'
    at Function.Module._resolveFilename (node:internal/modules/cjs/loader:933:15)
    at Function.Module._load (node:internal/modules/cjs/loader:778:27)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:81:12)
    at node:internal/main/run_main_module:17:47 {
  code: 'MODULE_NOT_FOUND',
  requireStack: []
}
```

We were hoping to read `'/Users/$USER/monorepo/node_modules/react-native-code-push/scripts/generateBundledResourcesHash.js'` instead.


_android/app/build.gradle_
```
/**
 * Set the monorepo root path.
 */
def rootDir = file("$projectDir/../../../../")

react {
    //   The root of your project, i.e. where "package.json" lives. Default is '..'
    // root = file("../")  <-- ´root´ will always be defined and required to point to RN's root
    reactNativeDir = file("$rootDir/node_modules/react-native")
    codegenDir = file("$rootDir/node_modules/react-native-codegen")
    cliFile = file("$rootDir/node_modules/react-native/cli.js")
    entryFile = file("$rootDir/apps/native/index.js")
    hermesCommand = "$rootDir/node_modules/react-native/sdks/hermesc/%OS-BIN%/hermesc"
}

/**
 * Code Push must use `nodeModulesPath` to assign the monorepo `node_modules`.
 */
project.ext.nodeModulesPath = "../../../../node_modules"
```

_codepush.gradle_
```
def nodeModulesPath;
if (config.root) {
    nodeModulesPath = Paths.get(config.root.asFile.get().absolutePath, "/node_modules");
} else if (project.hasProperty('nodeModulesPath')) {
    nodeModulesPath = project.nodeModulesPath // <-- The monorepo node modules path doesn't seem to be assigned as config.root will always be defined.
} else {
    nodeModulesPath = "../../node_modules";
}
```

We have attempted to override the `root` path in the `react` config to point to `file("../../../../")` (the monorepo root), but then the build task fails to find the `metro.config` file. We're happy to help if you guys need a hand.

agusvazquez:
Hello! I am having the same issue.

Would it be possible to merge and deploy this?

moreiraj-ppb:
Anyone to merge this?

moreiraj-ppb (approved):

lucen-ms (approved):

MikhailSuendukov:
Could you please pull our master branch into your branch? It fixed problems with tests.

joarkosberg:
> Could you please pull our master branch into your branch? It fixed problems with tests.

Fixed 🙂

PR #2565, opened 2023-08-16
chore: clarify supported versions for react-native 0.65-0.70

huynhtastic:
Clarify that v0.65-v0.70 of react-native is only compatible with ^7.0.0 versions of react-native-code-push


velimir-jankovic:
That is already clear in documentation. 
Also, docs will be updated to fix issue for new RN versions mentioning they're compatible with <8.0.0 version of CodePush.

PR #2577, opened 2023-09-01
Instance jsBundleURLForBundleRoot with fallbackExtension

JulioCVaz:
**Sumary**

In react-native 0.70 was updated file RCTBundleURLProvider and removed `fallbackResource` in `jsBundleURLForBundleRoot` method instance. The incorrect import cause problems during build with react-native-code-push. Now is need to use `fallbackExtension` instead `fallbackResource`.

**Reference**: https://github.com/facebook/react-native/blob/a69bc7f1fd1f914b98beb665fc9e688673742611/CHANGELOG-pre-070.md?plain=1#L472


JulioCVaz:

DmitriyKirakosyan:
@JulioCVaz, please resolve the conflicts. The documentation was partly updated, causing conflicts with your changes. Additionally, please remove `fallbackExtension` as it is not required.

Regarding the DemoApp, I don't think we need to modify the code in it, as it still uses an older version of react-native.

JulioCVaz:
> @JulioCVaz, please resolve the conflicts. The documentation was partly updated, causing conflicts with your changes. Additionally, please remove `fallbackExtension` as it is not required.
> 
> Regarding the DemoApp, I don't think we need to modify the code in it, as it still uses an older version of react-native.

Hey @DmitriyKirakosyan thank you for the update. Done in https://github.com/microsoft/react-native-code-push/pull/2577/commits/29bbdf486ec17826e7acbae1bf538620bff16ba8

JulioCVaz:
> @JulioCVaz Let's remove `fallbackExtension:nil` part, as it is not needed.

Perfect @DmitriyKirakosyan I think so this PR is not more necessary :)

DmitriyKirakosyan:
@JulioCVaz , you’re right. All the code parts with `fallbackResource` are under RN <0.70 blocks, so we can close this PR. Thanks a lot for your time and effort in contributing!

Issue #2567, opened 2023-08-17
App crashes in iOS each time new code push is released 


karthinidc:
The issue happened after the RN version update from 0.59.9 to 0.63.4 Previous version (0.59.9) working fine

### Steps to Reproduce

    1. Release the app 
    2. send some code push
    3. kill the app
    4. open the app again and install the code push
    5. see app crashes(if not happened - reproduce steps 3 and 4)

### Expected Behavior
The app works correctly, and updates content with code push

### Actual Behavior
App crashes each time a new code push is released

### Environment

* react-native-code-push version: 7.0.4
* react-native version: 0.63.4
* iOS version: 16.6
* Reproduce on a  release build
* Reproduce only on a physical device

Used code::::

AppDelegate.m

#import <CodePush/CodePush.h>

...

- (NSURL *)sourceURLForBridge:(RCTBridge *)bridge
{
#if DEBUG
  return [[RCTBundleURLProvider sharedSettings] jsBundleURLForBundleRoot:@"index" fallbackResource:nil];
#else
  return [CodePush bundleURL];
#endif
}

Info.plist

	<key>CodePushDeploymentKey</key>
	<string>OUR_PRODUCTION_KEY</string>

anthlasserre:
Hey @karthinidc 👋🏼 I think this might be coming from the `restartApp()` API. Are you able to reproduce this crash by running a restart through Codepush ? `Codepush.restartApp()`

karthinidc:
Yes @ anthlasserre -- This function crashes the App

anthlasserre:
Fine. So the original issue is there 👇🏼 
https://github.com/microsoft/react-native-code-push/issues/2539

karthinidc:
Any fix for this?

anthlasserre:
> Any fix for this?

@karthinidc Not yet, unfortunately... We don't really know what is causing this crash. Do you have `react-native-reanimated` set for your project? If yes, which version?

karthinidc:
Not using react-native-reanimated package in our project.

anthlasserre:
> Not using react-native-reanimated package in our project.

Ok, thanks. So nothing to do with it.

anthlasserre:
@karthinidc Are you able to reproduce the crash locally on a release build ?

karthinidc:
@anthlasserre Only in the release build

anthlasserre:
> @anthlasserre Only in the release build

[...]

PR #2621, opened 2023-11-21
Codepush HOC should unmount app before restart

savv:
RN Apps currently don't get the opportunity to destruct themselves prior to a restart, leading to crashes and bugs:

https://github.com/microsoft/react-native-code-push/issues/2567#issuecomment-1820827232

This is an attempt to mitigate these crashes, by properly unmounting `<App />`.

I'm looking forward to your feedback


savv:

savv:
ping?

viniciusmoreria:
this is not missing the state in the class construtor?

```
      constructor() {
        super();
        this.state = { inactive: false };
      }
```

velimir-jankovic:
@savv Thank you for this contribution! I'm a bit concerned that we may affect existing apps with this change.
I think it would be a good idea to put this change behind an option (for now), what do you think?

Issue #2654, opened 2024-02-01
CodePushMalformedDataException: Unable to parse contents of files/CodePush/codepush.json, the file may be corrupted


MAKARD:
### Steps to Reproduce
1. Make android app read corrupted codepush.json
2. Run the app
3. Observe the crash

### Expected Behavior
App removes corrupted codepush.json file and triggers codepush update flow again 

### Actual Behavior
App crashes
```
java.lang.RuntimeException: Unable to create application com.myapp.MainApplication: com.microsoft.codepush.react.CodePushMalformedDataException: Unable to parse contents of /data/user/0/com.myapp/files/CodePush/codepush.json, the file may be corrupted.
  at android.app.ActivityThread.handleBindApplication(ActivityThread.java:7608)
  at android.app.ActivityThread.access$2600(ActivityThread.java:260)
  at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2435)
  at android.os.Handler.dispatchMessage(Handler.java:110)
  at android.os.Looper.loop(Looper.java:219)
  at android.app.ActivityThread.main(ActivityThread.java:8668)
  at java.lang.reflect.Method.invoke(Method.java)
  at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:513)
  at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1109)
Caused by: com.microsoft.codepush.react.CodePushMalformedDataException: Unable to parse contents of /data/user/0/com.myapp/files/CodePush/codepush.json, the file may be corrupted.
  at com.microsoft.codepush.react.CodePushUtils.getJsonObjectFromFile(CodePushUtils.java:200)
  at com.microsoft.codepush.react.CodePushUpdateManager.getCurrentPackageInfo(CodePushUpdateManager.java:59)
  at com.microsoft.codepush.react.CodePushUpdateManager.getCurrentPackageHash(CodePushUpdateManager.java:109)
  at com.microsoft.codepush.react.CodePushUpdateManager.getCurrentPackage(CodePushUpdateManager.java:119)
  at com.microsoft.codepush.react.CodePush.initializeUpdateAfterRestart(CodePush.java:298)
  at com.microsoft.codepush.react.CodePush.<init>(CodePush.java:88)
  at com.facebook.react.PackageList.getPackages(PackageList.java:146)
  at com.myapp.MainApplication$1.getPackages(MainApplication.java:46)
[...]

Issue #1149, opened 2018-01-18
com.microsoft.codepush.react.CodePushMalformedDataException Unable to parse contents of [..]/codepush.json


maxkomarychev:
Hi,

this is the crash report I've got from crash-analytics service. I wasn't able to reproduce it myself however whatever happens the codepush should not crash, it should safely fallback to original bundle if error happened during it's initialization.

### Expected Behavior

CodePush should not crash the app. It should safely fallback to original bundle.

### Actual Behavior

App crashes.

```
java.lang.RuntimeException An error occurred while executing doInBackground() 
    AsyncTask.java:318 android.os.AsyncTask$3.done
    FutureTask.java:354 java.util.concurrent.FutureTask.finishCompletion
    FutureTask.java:223 java.util.concurrent.FutureTask.setException
    FutureTask.java:242 java.util.concurrent.FutureTask.run
    ThreadPoolExecutor.java:1133 java.util.concurrent.ThreadPoolExecutor.runWorker
    ThreadPoolExecutor.java:607 java.util.concurrent.ThreadPoolExecutor$Worker.run
    Thread.java:761 java.lang.Thread.run


Caused by: com.microsoft.codepush.react.CodePushMalformedDataException Unable to parse contents of /data/user/0/my.package.name/files/CodePush/codepush.json, the file may be corrupted. 
    CodePushUtils.java:198 com.microsoft.codepush.react.CodePushUtils.getJsonObjectFromFile
    CodePushUpdateManager.java:55 com.microsoft.codepush.react.CodePushUpdateManager.getCurrentPackageInfo
    CodePushUpdateManager.java:105 com.microsoft.codepush.react.CodePushUpdateManager.getCurrentPackageHash
    CodePushUpdateManager.java:115 com.microsoft.codepush.react.CodePushUpdateManager.getCurrentPackage
    CodePushNativeModule.java:295 com.microsoft.codepush.react.CodePushNativeModule$4.doInBackground
    CodePushNativeModule.java:292 com.microsoft.codepush.react.CodePushNativeModule$4.doInBackground
    AsyncTask.java:304 android.os.AsyncTask$2.call
    FutureTask.java:237 java.util.concurrent.FutureTask.run
    ThreadPoolExecutor.java:1133 java.util.concurrent.ThreadPoolExecutor.runWorker
    ThreadPoolExecutor.java:607 java.util.concurrent.ThreadPoolExecutor$Worker.run
[...]

PR #2668, opened 2024-02-21
fix: Clear updates on initializeUpdateAfterRestart in case codepush.json is corrupted

MAKARD:
A fix for https://github.com/microsoft/react-native-code-push/issues/2654, https://github.com/microsoft/react-native-code-push/issues/1149

Fixed in the same manner as for getJSBundleFileInternal
image

### DIFF
--- a/android/app/src/main/java/com/microsoft/codepush/react/CodePush.java
+++ b/android/app/src/main/java/com/microsoft/codepush/react/CodePush.java
@@ -301,7 +301,7 @@ public class CodePush implements ReactPackage {
                 packageMetadata = this.mUpdateManager.getCurrentPackage();
             } catch (CodePushMalformedDataException e) {
                 // We need to recover the app in case 'codepush.json' is corrupted
-                CodePushUtils.log(e.getMessage());
+                CodePushUtils.log(e);
                 clearUpdates();
                 return;
             }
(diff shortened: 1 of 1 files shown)


android/app/src/main/java/com/microsoft/codepush/react/CodePush.java
```diff
@@ -295,7 +295,16 @@ void initializeUpdateAfterRestart() {
...
         JSONObject pendingUpdate = mSettingsManager.getPendingUpdate();
         if (pendingUpdate != null) {
-            JSONObject packageMetadata = this.mUpdateManager.getCurrentPackage();
+            JSONObject packageMetadata = null;
+
+            try {
+                packageMetadata = this.mUpdateManager.getCurrentPackage();
+            } catch (CodePushMalformedDataException e) {
+                // We need to recover the app in case 'codepush.json' is corrupted
+                CodePushUtils.log(e.getMessage());
```
DmitriyKirakosyan, line 304: `CodePushUtils.log(e)` might be better here as it will be logged as an exception.

MAKARD: yep good point, fixed

DmitriyKirakosyan (approved):

lucen-ms (approved):

PR #2671, opened 2024-02-27
Update CodePush.java

billxie1988:
As your readme say the key is CodePushServerUrl
`<string moduleConfig="true" name="CodePushServerUrl">https://yourcodepush.server.com</string> `
otherwise will casue custom server 404 error


billxie1988:
CodePushServerUrl

PR #2318, opened 2022-08-10
Fix Android rollbacks when app started on background

aamagda:
Fixes for this issue https://github.com/microsoft/react-native-code-push/issues/2305

PR #2677, opened 2024-03-17
Codepush Rollback Issue on Android

shri2kpro:
Handling for rollbacks, when the app starts in background.


MikhailSuendukov:
Hello and thanks for your contribution!
Apparently this PR is a continuation of this PR. I looked at your changes, but my [concerns](https://github.com/microsoft/react-native-code-push/pull/2318#issuecomment-1570111325) about the following issue are still relevant.

> I have concerns that in this case the case will not be processed when a rollback needs to be performed due to an unsuccessful update download or invalid jsbundle. Which actually checks by calling notifyAppReady when the application is rendered. What will happen in this case? I think this case needs to be handled too.

What do you think about it?

BamboVN19:
@microsoft-github-policy-service agree [company="{easybossacademy-pega.com.vn/esFasting-Android}"]

DmitriyKirakosyan:
Closing. Please reopen if you are still interested.

PR #2780, opened 2024-11-06
Enabling test for Android w/ new-arch

kmsbernard:
- **init android app of DemoCodePushAppNewArch example**
- **support new architecture on android sdk**
- **exclude examples from compile**
- **setup test with react-native v0.76.1**
- **install code-push plugin from the new-arch branch**


kmsbernard:
mistakenly opening PR on the upstream repository 😅

PR #2784, opened 2024-11-26
Replace deprecated method for reloading bundle on iOS

esbenvb:
The code is currently using a deprecated method to reload the bundle. This PR will switch to the recommended method of using `RCTTriggerReloadCommandListeners`.

When using New Architecture, the iOS app does not restart properly using the deprecated method. (tested in RN 0.76.3, RN CodePush 9.0.0).

It either hangs or crashes. However, the new bundle is installed and loaded properly when the app is killed and restarted. 

Switching to the one in this PR, makes the iOS app restart properly and use the new bundle, when New Architecture is enabled, and also works when it's disabled.

### DIFF
--- a/ios/CodePush/CodePush.m
+++ b/ios/CodePush/CodePush.m
@@ -5,6 +5,7 @@
 #import <React/RCTEventDispatcher.h>
 #import <React/RCTRootView.h>
 #import <React/RCTUtils.h>
+#import <React/RCTReloadCommand.h>
 #else // back compatibility for RN version < 0.40
 #import "RCTAssert.h"
 #import "RCTBridgeModule.h"
@@ -540,7 +541,7 @@ static NSString *const LatestRollbackCountKey = @"count";
             [super.bridge setValue:[CodePush bundleURL] forKey:@"bundleURL"];
         }
 
-        [super.bridge reload];
+        RCTTriggerReloadCommandListeners(@"react-native-code-push: Restart");
     });
 }
 


DordeDimitrijev:
Hello @esbenvb  can you provide more details about these changes. How do these change affect RNCP behavior and why do you need them?

esbenvb:
> Hello @esbenvb can you provide more details about these changes. How do these change affect RNCP behavior and why do you need them?

I've added a description now, thanks for the reminder

DordeDimitrijev:
Thank you for updating the description. Since this contribution involves replacement of deprecated method and not actual implementation of changes for new architecture, could you please update the title and description accordingly? Currently it seems we support NA, but that is not the case.

esbenvb:
> Could you please update the title and description accordingly? Currently it seems we support NA, but that is not the case.

Done. If possible, I'd like to assist with other feedback regarding NA support, which would make much sense now that it's enabled default, and CodePush can live on as a standalone product.

DordeDimitrijev:
@esbenvb  Do you have a similar solution for android in mind, also could you confirm the changes are compatible with older versions of react native?

esbenvb:
> @esbenvb Do you have a similar solution for android in mind, also could you test the changes with older versions of react native?

Unfortunately, no. Tried to look into the Android code as well but I have more experience debugging iOS code, and didn't find a solution for reloading the bundle properly on Android with NA.

DmitriyKirakosyan (approved):

DordeDimitrijev (approved):

DordeDimitrijev:
Tested and verified the changes. Set up demo app with these changes with and without NA to test out reload of bundle.

lucen-ms (approved):

bardliu:
> > @esbenvb Do you have a similar solution for android in mind, also could you test the changes with older versions of react native?
> 
> Unfortunately, no. Tried to look into the Android code as well but I have more experience debugging iOS code, and didn't find a solution for reloading the bundle properly on Android with NA.
> 
> Regarding older versions, the `RCTTriggerReloadCommandListeners` call that I'm using here has been around for quite some time, also in older RN versions that's not supported by the v8.0.0 and v9.0.0 versions of react-native-code-push.

https://github.com/facebook/react-native/blob/main/packages/react-native/ReactAndroid/src/main/java/com/facebook/react/ReactDelegate.java#L250

hi～ maybe can use this away

John251314:
can merge to 8.3.1? My app need to support ios 13, so we can not use 9.0.1.

DordeDimitrijev:
Hello @John251314 as you're probably familiar appcenter is being deprecated and together with it is RNCP so we won't make any new releases. I suggest you create a fork with the changes you need.

